PR #1, opened 2020-10-09
Implement bazel rule for astore upload

ahungrynacho:
How would I test this locally if bazel in the internal repo references enkit in the github repo? How can I tell my local copy of the internal repo to reference a local copy of the enkit repo?

### DIFF
--- a/bazel/astore/defs.bzl
+++ b/bazel/astore/defs.bzl
@@ -62,19 +62,23 @@ to an artifact store.""",
 )
 
 def _astore_download(ctx):
-    outputs = [ctx.actions.declare_file(ctx.attr.download_src.split("/")[-1])]
-    ctx.actions.run(
-        arguments = ["download", ctx.attr.download_src],
-	executable = ctx.executable._astore_client,
-	outputs = outputs,
+    output = ctx.actions.declare_file(ctx.attr.download_src.split("/")[-1])
+    ctx.actions.run_shell(
+        command = "%s download --no-progress -o %s %s" % (ctx.executable._astore_client.path, output.path, ctx.attr.download_src),
+        use_default_shell_env = False,
+        tools = [ctx.executable._astore_client],
+        outputs = [output],
+	#TODO: remove env after build user is created
+	env = {"HOME": "/home/bbhuynh"},
     )
-    return [DefaultInfo(files = depset(outputs))]
+    return [DefaultInfo(files = depset([output]))]
 
 astore_download = rule(
     implementation = _astore_download,
     attrs = {
         "download_src": attr.string(
             doc = "Provided the full path, download a file from astore.",
+            mandatory = True,
         ),
         "_astore_client": attr.label(
             default = Label("//astore/client:astore"),
@@ -83,8 +87,8 @@ astore_download = rule(
             cfg = "host",
         ),
     },
-    doc = """Downloads artifacts from artifact store - astore. \
-		    With this rule, you can easily download \
-		    files from an artifact store.""",
-)
+    doc = """Downloads artifacts from artifact store - astore. 
 
+With this rule, you can easily download 
+files from an artifact store.""",
+)
--- a/bazel/astore/defs.bzl
+++ b/bazel/astore/defs.bzl
@@ -60,3 +60,47 @@ astore_upload = rule(
 With this rule, you can easily upload the output of a build rule
 to an artifact store.""",
 )
+
+def _astore_download(ctx):
+    push = ctx.actions.declare_file("astore_download.sh")
+    inputs = [ctx.executable._astore_client]
+    template = ctx.file._astore_download_file
+    ctx.actions.expand_template(
+        template = template,
+        output = push,
+        substitutions = {
+            "{astore}": ctx.executable._astore_client.short_path,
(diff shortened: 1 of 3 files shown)


ccontavalli:
> How would I test this locally if bazel in the internal repo references enkit in the github repo? How can I tell my local copy of the internal repo to reference a local copy of the enkit repo?

What I usually do is edit the WORKSPACE file to use:
```
local_repository(
  name = "enkit",
  path = "/path/to/my/modified/enkit/repo",
)
```
while commenting out the http_archive rule for enkit.

bazel/astore/defs.bzl
```diff
@@ -60,3 +60,47 @@ astore_upload = rule(
...
+        },
+        is_executable = True,
+    )
+    return [DefaultInfo(executable = push, runfiles = ctx.runfiles(inputs))]
+
+astore_download = rule(
+    implementation = _astore_download,
+    attrs = {
+        "destination": attr.string(
+            doc = "The file will be downloaded from astore to this path.",
```
ccontavalli, line 84: is this mandatory? the -o parameter is always provided in the script

Either:
* add `mandatory = True` or...
* make the command tolerant to not having a destination specified.

bazel/astore/defs.bzl
```diff
@@ -60,3 +60,47 @@ astore_upload = rule(
...
+        ),
+        "file": attr.string(
+            doc = "Provided the full path, download a file from astore.",
```
ccontavalli, line 87: same comment as above, is this mandatory?

bazel/astore/defs.bzl
```diff
@@ -60,3 +60,47 @@ astore_upload = rule(
...
+            allow_single_file = True,
+            executable = True,
+            cfg = "host",
+        ),
+    },
+    executable = True,
+    doc = """Downloads artifacts from artifact store - astore.
+
+With this rule, you can easily download files
+to an artifact store.""",
```
ccontavalli, line 104: from an artifact store

bazel/astore/defs.bzl
```diff
@@ -60,3 +60,47 @@ astore_upload = rule(
 With this rule, you can easily upload the output of a build rule
 to an artifact store.""",
 )
+
+def _astore_download(ctx):
```
ccontavalli, line 64: ok, almost :) so... when you invoke bazel to do something, you can generally do three things: `run` a target, or `build` the target, or `test` the target. Forget about test.

* `run` -> performs an action, does something (for example, deploys a binary to production).
* `build` -> creates something to be used by the rest of the build.

Now... the `_astore_upload` run I created is a `run` kind of rule. The idea is that the user builds a bunch of stuff, tests it out, and then when he's happy with the build, will finally deploy to astore. So the rule was structured so that the user would call `bazel build whatever` and then `bazel run :upload` to send to astore.

So, what the `_astore_upload` rule does is *create something executable* to be *run later*, when the *user asks for it* by separately invoking `bazel run ...`.

`_astore_download` needs to be a `build` rule: it needs to immediately `create` a file to be used by the rest of the build process. `DefaultInfo` should not return an `executable`, but a set of `files` to use.

This is what I recommend you do:
* instead of using `ctx.actions.expand_template`, use `ctx.actions.run` or `ctx.actions.run_shell` (https://docs.bazel.build/versions/master/skylark/lib/actions.html#run). Ditch the astore_download.sh file - it's a simple enough command, that you don't need a separate shell script.
* in `actions.run`, specify your `ctx.executable._astore_client` as a `tool` the rule needs, rather than as an input. So you can ditch the `inputs` variable as well.
* create an `outputs` array, listing the outputs the rules create (use `ctx.action.declare_file` on the `ctx.attr.file` to create a file object). Pass this `outputs` array as the argument `outputs` of `run` or `run_shell`.

And finally, super important: change the DefaultInfo. This rule is no longer creating an executable someone can run, with a set of inputs it has to process. Rather it is just creating a set of output files.

Should be as simple as: `return [DefaultInfo(files = outputs)]`

ccontavalli (changes requested):

bazel/astore/defs.bzl
ahungrynacho: Yes, you need to specify the full path within the astore directory to download the correct file.

ccontavalli:
All the comments I have added were style nits, this looks pretty good? Thank you!

bazel/astore/defs.bzl
ccontavalli: ah! yes. Imagine you have a function that expects a python set, you are passing in a list, that's the error.

To convert a list to a depset, based on the example here:
https://docs.bazel.build/versions/master/skylark/lib/globals.html#depset

you probably just need to do

```
return [DefaultInfo(files = depset(outputs))]

Comment (ccontavalli) [reply bazel/astore/defs.bzl:64]: the type conversion... in the _kernel_module macro I have:
```
    output = ctx.actions.declare_file(rename)
...
    return DefaultInfo(files = depset([output]))
```

which should just work? I have seen the use of list comprehension otherwise, eg, files = despet([ ... for f in ...])

ahungrynacho:
TODO: remove the hard coded home directory once the build user is created and integrated

PR #6, opened 2021-01-05
[WIP] initial draft of adding astore deletion

aaahrens:
Currently:

- adds command with dry run flag
- adds proto stubs

Plan (detailed):

- finish client stubbing in `astore/client/astore/delete.go`
- add additional/remove flags? 
- implement guess sid or id from single string
- add raw delete functionality server side and unit test
- add delete to rpc server and implement e2e testing with client stubs

Future:

- Running garbage collection could just be a chronjobbed github action that reads the metadata from the remote and deletes as necessary. Linux runners are pretty much free.

- Altering the timestamps of files could be a good way to test this

- for e2e testing we can multistage an intermediate docker image and save code coverage across stages.

### DIFF
--- a/astore/client/astore/delete.go
+++ b/astore/client/astore/delete.go
@@ -1,0 +1,10 @@
+package astore
+
+import (
+	"context"
+	"github.com/enfabrica/enkit/astore/rpc/astore"
+)
+
+func (c *Client) Delete(ctx context.Context, request *astore.DeleteRequest) {
+	//TODO
+}
--- a/astore/client/commands/commands.go
+++ b/astore/client/commands/commands.go
@@ -33,7 +33,6 @@ func New(base *client.BaseFlags) *Root {
 	root.AddCommand(NewTag(root).Command)
 	root.AddCommand(NewNote(root).Command)
 	root.AddCommand(NewPublic(root).Command)
-	root.AddCommand(NewDelete(root).Command)
 	return root
 }
 
--- a/astore/client/commands/commands.go
+++ b/astore/client/commands/commands.go
@@ -33,7 +33,7 @@ func New(base *client.BaseFlags) *Root {
 	root.AddCommand(NewTag(root).Command)
 	root.AddCommand(NewNote(root).Command)
 	root.AddCommand(NewPublic(root).Command)
-
+	root.AddCommand(NewDelete(root).Command)
 	return root
 }
 
--- a/astore/client/commands/delete.go
+++ b/astore/client/commands/delete.go
@@ -1,0 +1,38 @@
+package commands
+
+import (
+	"github.com/spf13/cobra"
+)
+
+type DeleteCommand struct {
+	*cobra.Command
+	root *Root
+
+	name string
+}
+
+func (uc *Delete) Run(cmd *cobra.Command, args []string) error {
+	//TODO
+	return nil
+}
+
+type Delete struct {
+	*cobra.Command
+	root *Root
+	isDryRun bool
+}
+
+func NewDelete(root *Root) *Delete {
+	command := &Delete{
+		Command: &cobra.Command{
+			Use:   "delete",
+			Short: "Deletes an artifact from astore",
+
+		},
+		root: root,
+	}
+	command.Flags().BoolVarP(&command.isDryRun, "dry-run", "dr",  false, "will attempt to delete the resource without doing so")
+	command.Command.RunE = command.Run
+	return command
+}
+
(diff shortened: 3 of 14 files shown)


astore/client/astore/delete.go
```diff
@@ -0,0 +1,10 @@
+package astore
+
+import (
+	"context"
+	"github.com/enfabrica/enkit/astore/rpc/astore"
+)
+
+func (c *Client) Delete(ctx context.Context, request *astore.DeleteRequest) {
```
ccontavalli, line 8: Shouldn't the signature be Delete(...) (*astore.DeleteResponse, error) ?

astore/rpc/astore.proto
```diff
@@ -133,13 +133,23 @@ message NoteResponse {
   repeated Artifact artifact = 1;
 }
 
+message DeleteRequest {
+  string id = 1; //SID or UID (will be interpreted to which based on length)
+}
+
+message DeleteResponse {
+  int32 code = 1; //Valid values are like fs errors, either 0 1 or 2;
```
ccontavalli, line 141: Just return an empty message, no need to define code and message.
gRPC will carry both an error string and error code implicitly for each rpc call.

Just return an error from your server handler, and the client side will receive the same error. 
To specify an error code, you can use [status.Errorf here](https://godoc.org/google.golang.org/grpc/status#Errorf)
If you don't specify an error code explicitly, some default is used (APPLICATION... something something? not sure).

astore/client/commands/delete.go
```diff
@@ -0,0 +1,38 @@
...
+
+func (uc *Delete) Run(cmd *cobra.Command, args []string) error {
+	//TODO
+	return nil
+}
+
+type Delete struct {
+	*cobra.Command
+	root *Root
+	isDryRun bool
```
ccontavalli, line 22: Did not indicate in the ticket, and can be in a follow up: but when a sid is specified, this may result in multiple uids removed [1]. The tool should  default to ask for confirmation, there should be a flag to disable it.


[1]: eg, sid identifies an artifact, like the "tools/enkit" binary. uid identifies a specific version, like the enkit we released 2 months ago, so if you remove a sid, the intent is to remove all versions of enkit from the repo, while if you remove the uid, you are removing a "single file".

astore/rpc/astore.proto
```diff
@@ -133,13 +133,23 @@ message NoteResponse {
   repeated Artifact artifact = 1;
 }
 
+message DeleteRequest {
+  string id = 1; //SID or UID (will be interpreted to which based on length)
```
ccontavalli, line 137: nit: make sure you run `go fmt` before submitting. We have no enforcement mechanism (yet, there should be one).

astore/rpc/astore.proto
ccontavalli: Actually, maybe return the list of things that were actually deleted? Up to you.

astore/client/commands/commands.go
```diff
@@ -33,7 +33,7 @@ func New(base *client.BaseFlags) *Root {
 	root.AddCommand(NewTag(root).Command)
 	root.AddCommand(NewNote(root).Command)
 	root.AddCommand(NewPublic(root).Command)
-
+	root.AddCommand(NewDelete(root).Command)
```
ccontavalli, line 36: Up to you, but I'm personally a fan of continuous, small incremental updates on master. So the code gets tested by automated pipelines (well, once we have them...), other folks in the team have visibility into the features, and there are fewer merge conflicts as we go on, with less branching and merging.

If you remove the registration of the sub-command - so effectively the code is not released/disabled - and get the code to compile and `go fmt` clean, I'd be happy to have this merged already.

astore/client/astore/delete.go
ccontavalli: Ah, this is the Client Delete. Thought this was the server one.

astore/rpc/astore.proto
aaahrens: I think that returning a list of ID's is more idiomatic, especially since deleting an SID would result in a bunch of uid's being deleted also.

ccontavalli: Don't feel strongly. Let's start implementing it, and decide later :)? The goal of returning a list here though is to show the user what was deleted. The uids won't be human friendly/helpful.

Having subsequent lookups for each UID would fail, as the artifact has now been removed from the repository? Well, depends on how we implement the removal, the UID could stick around, just not tied to a path. But if we do provide an API to get all the info out of an UID, it's certainly arguable if it should/should not return results that have been deleted.

ccontavalli (approved):

PR #9, opened 2021-01-28
[WIP] Adding lint rule + other integrations

aaahrens:
adds a go lint rule that takes a go_path target and lints it 
`go_lint(
    name = "lint_astore",
    go_library = ["//astore/client:astore_source"],
    output = "//:lint_result.txt"
)`

what's left to do:

- [x]  export lint results (right now is just in stdout)
- [x]  configure types of lint output (eg. github actions, coverage for local ide etc)
- [x] vendor golang-ci-lint in the rule
- [x]  reuse lint cache?

### DIFF
--- a/bazel/linting/scripts/git.sh
+++ b/bazel/linting/scripts/git.sh
@@ -1,9 +1,11 @@
 #!/usr/bin/env bash
+# This script gets the stable diff from the workspace command
 if [ -z "$1" ]; then
   echo "must present status file"
   exit 1
 fi
 line=$(grep STABLE_GIT_MASTER_DIFF "$1" | cut -d " "  -f2-)
+# this is specifically disabled because we want the side effect
 # shellcheck disable=SC2162
 read -a arr <<< "$line"
 for i in "${arr[@]}"
--- a/bazel/workspace_status.sh
+++ b/bazel/workspace_status.sh
@@ -45,4 +45,4 @@ echo STABLE_GIT_MASTER_AUTHOR "$(git show -s --format='%ae' "$GIT_MASTER_SHA")"
 echo STABLE_GIT_CHANGES "$(git status --porcelain |paste -sd "," -)"
 
 # Spits out the files changed between this branch's remote and the default master branch. Space separated.
-echo STABLE_GIT_MASTER_DIFF "$(git --no-pager diff --name-only "$GIT_ORIGIN_BRANCH"..."$(git rev-parse --abbrev-ref HEAD)" | tr '\r\n' ' ')"
+echo STABLE_GIT_MASTER_DIFF "$(git --no-pager diff --name-only "$GIT_ORIGIN_BRANCH"...origin/master | tr '\r\n' ' ')"


bazel/scripts/lint.sh
```diff
@@ -0,0 +1,23 @@
...
+mkdir -p .cache
+
+
+#create output files
+mkdir -p $(dirname ${LINT_OUTPUT})
+touch ${LINT_OUTPUT}
+export LINT_OUTPUT="$PWD/${LINT_OUTPUT}"
+
+cd $(find $PWD -name enfabrica)/enkit
+echo "$(golangci-lint run --path-prefix / --issues-exit-code 0)" >> ${LINT_OUTPUT}
```
ccontavalli, line 22: `golangci-lint` is a go program. Most of them are very easy to bring into a bazel hermetic environment, without having to rely on the tool being installed in the system (eg, bazel will download and build it for you - bazel will even version it for you! fixing most of the reasons why the golangci-lint guys don't want you to build it!).

How?
- (first time, start from a clean branch, I recommend you send a separate PR for this the first time)
- from the top level directory of the repository, run `go get ...whatever...golangci-lint`, it will change the `go.mod` file, versioning the tool in our repository.
- run something like `bazelisk run //:gazelle -- update-repos -from_file=go.mod -to_macro=bazel/go_repositories.bzl%go_repositories` - this will version all the dependencies in bazel. This would normally update the WORKSPACE file, but in the enkit repo, WORKSPACE file just imports bazel/go_repositories.bzl, and runs the go_repositories macro, thus the extra argument.
- this will generate a BUILD.bazel file for the golangci-lint tool.

Once this is done, you can have a tool dependency in the rule, pointing to the built golangci-lint tool, and pass the path as an environment variable, for example.

In the end, what will happen is that when your rule is run, bazel will look for the tool, won't find it, will download and build it (at the specific version *we* tested and coded in the workspace file, not a random version on the guy's laptop), and run it (and cache it).

It looks like a lot of steps, but once you know how to do it and you've done it a couple times, it's like 10-15 mins work.

bazel/scripts/defs.bzl
```diff
@@ -0,0 +1,46 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_context")
```
ccontavalli, line 1: meta: we have a directory "per feature" under the bazel directory. Everything under `bazel/scripts`.

All you have under `bazel/scripts` should go under a `bazel/lint`.

.gitignore
```diff
@@ -9,3 +9,5 @@ node_modules
 credentials*.json
 *.flag*
 *.config*
+.idea/
```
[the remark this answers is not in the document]
ccontavalli: Address this comment? Separate PR?

bazel/linting/scripts/lint_go.sh
```diff
@@ -0,0 +1,41 @@
...
+
+cd $(find $PWD -name enfabrica)/enkit
+echo $PWD
+#golangci-lint run ./...
+ls lib/khttp/protocol
+for i in "${arr[@]}"
+do
+  if [[ $i == *.go ]]; then
+    go_package=$(dirname $i)
+    golangci-lint run $go_package --issues-exit-code 0 2>&1 | sed 's/.*'enkit'//' | tee ${LINT_OUTPUT}
```
ccontavalli, line 36: so, in PR https://github.com/enfabrica/enkit/pull/29/files I see you pull in golangci-lint, so bazel can build it.

But in this PR, I don't see any code to pull in the golangci built by bazel? The line here is just running the golangci on your system, which won't be available on random cloud machines we run this test on?

The typical way to address this is:
1) Have a dependency in the rule to run the linter on the linting tool, eg, something like _linter = Label("@golangci-lint...").
2) Use the path of the bazel built golangci-lint in scripts, rather than the internal one.

bazel/testing/templates/golangci-lint.yaml
```diff
@@ -0,0 +1 @@
+
```
ccontavalli, line 1: I don't see bazel/testing/templates referenced anywhere? do we need those two files? Did I miss something?

.bazelrc
```diff
@@ -6,6 +6,7 @@
 
 # Cache action outputs on disk so they persist across output_base and bazel shutdown (eg. changing branches)
 build --disk_cache=~/.cache/bazel-disk-cache
+build --workspace_status_command "/bin/bash bazel/workspace_status.sh"
```
ccontavalli, line 9: Please break this PR down further. Can you send a single PR adding the workspace status stuff? I'd like it in the `internal` repo as well. Can you take care of that as well?

bazel/linting/scripts/git.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+#github_sha=$GIT_SHA1
```
ccontavalli, line 2: what are all the comments for? remove?

ccontavalli: (what is this script for? does not seem to do anything?)

ccontavalli (commented):
This does not look like it's ready for merging. Let's chat some time next week.

bazel/linting/defs.bzl
```diff
@@ -0,0 +1,190 @@
...
+                "GIT_DATA": inputs[0].path,
+                "GOLANGCI_LINT": ctx.files._golangci_lint[0].path,
+                "TARGET": target,
+            },
+        )
+    return DefaultInfo(
+        files = depset(outfiles + ctx.files.deps),
+    )
+
+go_lint = rule(
```
ccontavalli, line 63: can you add `doc =` attributes to explain the rule, and the attributes?

eg, what is deps? list of .go files directly? When would you use it over `go_libraries`?

bazel/linting/defs.bzl
```diff
@@ -0,0 +1,190 @@
...
+        ),
+        "deps": attr.label_list(),
+    },
+    toolchains = ["@io_bazel_rules_go//go:toolchain"],
+)
+
+def _parse_lint(ctx):
+    """
+
+    """
```
ccontavalli, line 97: ? what is this block in """ and """?

bazel/linting/deps.bzl
```diff
@@ -0,0 +1,13 @@
+load("//bazel/utils:binary.bzl", "download_binary")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+def lint_deps_init():
+    http_archive(
```
ccontavalli, line 5: wrap this in `maybe`? similar to other _init rule (except maybe did not exist back then, so they do it the old manual way).

https://docs.bazel.build/versions/1.1.0/repo/utils.html#maybe

bazel/linting/scripts/git.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
ccontavalli, line 1: short comment to describe what this does?

bazel/linting/scripts/lint_go.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+#Golang setup
+export PATH="${PWD}/${GO_LOCATION}/bin:$PATH"
+VENDOR_DIR="$(find  "$PWD" -name "$TARGET"_source)"
+SOURCE_LINT="$(find "$VENDOR_DIR" -name "$GO_LIBRARY_NAME")"
+export GOPATH=$SOURCE_LINT
+
+# Store golang-ci binary
+GOLANGCI_LINT=$PWD/$GOLANGCI_LINT
```
ccontavalli, line 10: double quotes? is this shellcheck clean?

bazel/linting/scripts/lint_go.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
```
ccontavalli, line 2: a top level comment? what does this script do?

also, which global environment variables does it expect? not immediately obvious. Looks like $TARGET? $GO_LIBRARY_NAME?

I would consider also adding ifs/conditionals erroring out if those variables are not provided.

bazel/linting/scripts/merge.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
ccontavalli, line 1: same comment as per other files. What does this do? top level comment.

bazel/linting/scripts/merge.sh
```diff
@@ -0,0 +1,41 @@
...
+  done < <(cat "$@")
+}
+
+if [[ "$STRATEGY" == "ALL" ]]; then
+  cleanFiles "$@" >> "$OUT"
+  exit 0;
+fi
+if [[ "$STRATEGY" == "git" ]]; then
+  GIT_FILE_PATH="$(find "$PWD" -name "bin")"
+#  combining them doesn't work, have 0 idea why
```
ccontavalli, line 22: not sure what this comment refers to? remove or clarify?

enkit/BUILD.bazel
```diff
@@ -4,7 +4,7 @@ go_library(
     name = "go_default_library",
     srcs = ["main.go"],
     importpath = "github.com/enfabrica/enkit/enkit",
-    visibility = ["//visibility:private"],
+    visibility = ["//visibility:public"],
```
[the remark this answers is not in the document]
ccontavalli: maybe not all unrelated, you need some targets marked public to be exposed to the lint rule. But see the comment about the other //public target, and there are a few unrelated files.

PR #11, opened 2021-02-08
Add script to generate files for kernel test build

stefanoen:
Our BAZEL build rules for kernel modules expect a .tar.gz file built in a
certain way. For distros that file is generated through our kbuild script, but
since distros ship with KUnit disabled, we can't use that for testing images.

generate_custom_archive.sh can generate a .tar.gz accepted by our BAZEL build
rules starting from a compiled linux kernel source tree. Not the best piece of
code, but it gets the job done ;)

Test: generated a .tar.gz, uploaded to astore, and used this new test image to
build a kernel module.


kbuild/utils/generate_custom_archive.sh
```diff
@@ -0,0 +1,88 @@
...
+
+usage()
+{
+        cat <<EOF
+Generate a kbuild compliant .tar.gz starting from a compiled custom linux source tree.
+usage: ${0##*/} OPTS
+    -k KDIR      Directory containing the compiled custom linux source tree
+    -t OUTDIR    Root directory for containing the generated kernel package
+    -v KVER      Kernel version (used to name the package)
+This will generate a KVER-test.tar.gz file inside OUTDIR, built from the pre-compiled kernel in KDIR.
```
ccontavalli, line 15: nit: I would expand the instructions here more. What is your defintion of "built" from the pre-compiled kernel in KDIR?

The script does not invoke any make target, and does not check anything present in the tree. Is `make modules` enough? Can I just unpack the tarball and have a .config file? what is the recommendation?

ccontavalli (approved):

kbuild/utils/generate_custom_archive.sh
stefanoen: Added the expected steps preceding the call to this script.

PR #13, opened 2021-02-09
WIP: Add rule to declare and run kernel unit tests

stefanoen:
Introduce a bash script to run kernel unit tests inside a user-mode linux image
generated with KUnit enabled. To do that the scripts expects three parameters:
1. Path to the user-mode linux image
2. Path to the rootfs image
3. Path to the folder to expose as hostfs to the user-mode linux proces (this
   can be used to expose test kernel modules to the running user-mode linux
   process, so that it will be able to load and run them at runtime).

The new macro rule simplifies the work required to instantiate the underlying
sh_test rule: it defines the kernel module, linux and rootfs image as required
data to run the test, then passes their runtime paths to the bash script we
just introduced.

### DIFF
--- a/bazel/linux/defs.bzl
+++ b/bazel/linux/defs.bzl
@@ -312,27 +312,25 @@ Example:
     },
 )
 
-def kernel_test(name, repo_name, module, kernel_image, rootfs_image):
-    """Convenience wrapper around sh_test, makes it easier to declare and run KUnit user-mode linux tests.
+def kernel_test(name, module, kernel_image, rootfs_image, tap_parser, repo_name="@enkit"):
+    """Convenience wrapper around sh_test, useful for running kernel tests.
 
     Args:
       name: string, name for the underlying sh_test rule.
-      repo_name: label, @name used to import this external repository. Used to
-                 implicitly define the local file run_um_kunit_tests.sh as the test
-	         runner.
       module: label, KUnit kernel module containing the tests.
       kernel_image: label, executable user-mode linux image file.
       rootfs_image: label, rootfs image file to be used by the linux image.
+      tap_parser: label, script to use to parse the test TAP output.
+      repo_name: label, @name used to import this external repository. Used to
+                 implicitly define the local file run_um_kunit_tests.sh as the test
+                 runner. Default = "@enkit".
     """
     srcs = [repo_name + "//bazel/linux:run_um_kunit_tests.sh"]
     data = [
-        module,
         kernel_image,
-        rootfs_image
+        rootfs_image,
+        module,
+	tap_parser
     ]
-    args = [
-        "$(location %s)" % kernel_image,
-        "$(location %s)" % rootfs_image,
-        "$(location %s)" % module
-    ]
+    args = ["$(location %s)" % elem for elem in data]
     return native.sh_test(name=name, srcs=srcs, data=data, args=args)
--- a/bazel/linux/defs.bzl
+++ b/bazel/linux/defs.bzl
@@ -311,3 +311,23 @@ Example:
         ),
     },
 )
+
+def kernel_unit_test(name, module, linux_image, rootfs_image):
+    """Convenience wrapper around sh_test, makes it easier to declare and run KUnit user-mode linux tests.
+
+    Args:
+      name: string, name for the underlying sh_test rule.
+      module: label, target KUnit kernel module containing the tests.
+      linux_image: label, executable user-mode linux image file.
+      rootfs_image: label, rootfs image file to be used by the linux image.
+    """
+    srcs = ["run_um_kunit_tests.sh"]
+    data = [
+        module,
+        linux_image,
+        rootfs_image]
+    args = [
+        "$(location %s)" % linux_image,
+        "$(location %s)" % rootfs_image,
+        "$(location %s)" % module]
+    return native.sh_test(name=name, srcs=srcs, data=data, args=args)
--- a/bazel/linux/defs.bzl
+++ b/bazel/linux/defs.bzl
@@ -312,22 +312,27 @@ Example:
     },
 )
 
-def kernel_unit_test(name, module, linux_image, rootfs_image):
+def kernel_test(name, repo_name, module, kernel_image, rootfs_image):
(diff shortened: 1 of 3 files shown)


bazel/linux/defs.bzl
```diff
@@ -311,3 +311,23 @@ Example:
...
+def kernel_unit_test(name, module, linux_image, rootfs_image):
+    """Convenience wrapper around sh_test, makes it easier to declare and run KUnit user-mode linux tests.
+
+    Args:
+      name: string, name for the underlying sh_test rule.
+      module: label, target KUnit kernel module containing the tests.
+      linux_image: label, executable user-mode linux image file.
+      rootfs_image: label, rootfs image file to be used by the linux image.
+    """
+    srcs = ["run_um_kunit_tests.sh"]
```
stefanoen, line 324: What's the best way to expose/declare this file?

Ideally, we want store this here because it is common to all kernel unit tests. How would this interact with other repositories importing this definition from their WORKSPACE file?

bazel/linux/run_um_kunit_tests.sh
```diff
@@ -0,0 +1,19 @@
...
+set -e
+
+mkdir -p /tmp/uml
+USER=$(whoami)
+chown "${USER}.${USER}" /tmp/uml
+chmod 777 /tmp/uml
+export TMPDIR=/tmp/uml
+
+OUTFILE="${TMPDIR}/output"
+KUNIT_TAP_PARSER=/path/to/kunit_tap_parser
```
stefanoen, line 16: On the same note as the script itself, how should we expose/declare this other script? I'm guessing this will need to be declared as data in the bazel rule, and its path passed as an argument to this script.

bazel/linux/defs.bzl
stefanoen: After a bit of digging it seems that the `export_files` built-in function does exactly that (https://docs.bazel.build/versions/master/be/functions.html#exports_files). I'll test its behavior and then update the PR.

stefanoen: Now the macro correctly references the local run_um_kunit_tests.sh script. This requires a new parameter representing the name assigned to this project when imported through the WORKSPACE file. I don't know if this is common practice/acceptable, I'm open to other suggestions on how to handle this.

ccontavalli: For rules, it just works? See for example the `_validator` or `_expander` use in `bazel/enkit/defs.bzl`, they use the path relative to the enkit WORKSPACE, despite being referenced from the internal WORKSPACE.

For macros, I suspect you are right, as the macro expansion happens in the context of the including WORKSPACE. I've asked for the best way on the bazel slack channel, but no reply yet. A bunch of stackoverflow questions talk about how to retrieve the name of the repository/workspace, but not sure they work / are correct, they talk about ctx which is not accessible from a macro.

I would leave it as is, modulo the comment above.

bazel/linux/defs.bzl
```diff
@@ -311,3 +311,28 @@ Example:
         ),
     },
 )
+
+def kernel_test(name, repo_name, module, kernel_image, rootfs_image):
```
ccontavalli, line 315: Do we have a README indicating how to use enkit from another repo? If we had one, it'd probably say to import it as @enkit. Can you default repo_name to be = to `@enkit`? So it works for those that use the `default` name.

bazel/linux/run_um_kunit_tests.sh
ccontavalli: yes, declare as data in the bazel rule, and then pass the path as env variable, or command line parameter.

I've often seen rules use ctx.write_template to generate a shell script, expanding the paths and parameters inside to avoid / reduce impact of escape hell.

bazel/linux/run_um_kunit_tests.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -e
+
+LINUXIMG=$1
+ROOTFS=$2
+HOSTFS=$(dirname $(realpath $3))
+
+export TMPDIR=$(mktemp -d)
```
ccontavalli, line 9: now that I think about it, bazel runs tests in a temporary directory already?

Maybe you can just use ./? or mktemp --tmpdir="."? don't feel strongly either way - feel free to leave as is. However, it is marginally better to use bazel's tempdir, as ... bazel will own cleaning it up, ensuring it's in the right location on eg a remote worker node, etc.

ccontavalli (approved):
Fix the comment, set the repo_name to a default value, fix the /pat/to/kunit_tap_parser, and I'll be happy to merge. All other comments are optional.

bazel/linux/defs.bzl
```diff
@@ -311,3 +311,28 @@ Example:
...
+    """Convenience wrapper around sh_test, makes it easier to declare and run KUnit user-mode linux tests.
```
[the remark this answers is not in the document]
stefanoen: Fixed.

bazel/linux/run_um_kunit_tests.sh
stefanoen: It would probably work, but I don't think it's necessary better. I'm already planning on a new iteration for the testing environment, involving rules and specific providers. In that case we can directly create and declare the new file from bazel, so this should be solved.

bazel/linux/run_um_kunit_tests.sh
stefanoen: > yes, declare as data in the bazel rule, and then pass the path as env variable, or command line parameter.

It seems that exported env variables don't carry over the bazel execution environment. I will just consider it as another path parameter.

> I've often seen rules use ctx.write_template to generate a shell script, expanding the paths and parameters inside to avoid / reduce impact of escape hell.

Given the simplicity of our script I think we would just be overengineering it (+ without templating we can run the script manually as well).

bazel/linux/defs.bzl
stefanoen: Yes, I agree that @enkit is a sane default.

ccontavalli (approved):

PR #15, opened 2021-02-12
Introduce kunit tap parser

stefanoen:
Add a directory partially mirroring the Linux kernel kunit tools subdir and use that as the default tap parser for the kernel_test bazel rule.

### DIFF
--- a/bazel/linux/defs.bzl
+++ b/bazel/linux/defs.bzl
@@ -312,7 +312,9 @@ Example:
     },
 )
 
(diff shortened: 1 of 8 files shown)


stefanoen:
I've put the directory mirroring the Linux kernel kunit tools subdir in bazel/linux, instead of kbuid. I think this makes more sense since kunit is related to the `kernel_test` macro, and not to the kbuild script. @ccontavalli, what do you think (I don't feel strongly about this)?

ccontavalli (approved):

PR #18, opened 2021-02-12
bazel/linux: Add README.md intoducing the rules and workflow

stefanoen:


bazel/linux/README.md
```diff
@@ -0,0 +1,33 @@
...
+# Main rules made available
+`kernel_version` defines a specific kernel version to compile and test modules against.
+It must point to a .tar.gz with a specific format, most likely built using either [kbuild](https://github.com/enfabrica/enkit/tree/master/kbuild) or [generate_custom_archive.sh](https://github.com/enfabrica/enkit/blob/master/bazel/linux/run_um_kunit_tests.sh).
+It is usually added to a project WORKSPACE file, defining a repository which can then be referenced using the @rule_name syntax.
+
+`kernel_module` defines a kernel module. It requires a *kernel_version* rule label as a kernel parameter.
+It generates a .ko file that can then be loaded in any machine running a kernel compatible with the one specified by *kernel_version*.
+
+`kernel_test` defines a kernel test. It requires a *kernel_module* rule label as a module parameter, representing the kernel module to be tested.
+Additionally, it requires the labels of two files, one representing a root filesystem image, the other an executable user-mode linux kernel image (compatible with the kernel version used to compile the module to be tested).
```
ccontavalli, line 13: How do we build the UML kernel image? How do we create a root file system? Provide an example of both?

We should check in the init / scripts that get the rootfs to work as described below.

(ok to address this in a follow up PR, this is very good for merge - thanks for adding it!)

ccontavalli (approved):

bazel/linux/README.md
stefanoen: > How do we build the UML kernel image?

Should be solved by https://github.com/enfabrica/enkit/pull/22 and https://github.com/enfabrica/enkit/pull/23

> How do we create a root file system?

I have a buildroot patch that includes the required stuff to generate a barebone rootfs image (basically you `git am file.path && make` and obtain the rootfs img). For the time being, we could simply store the patch somewhere and document the manual process (I don't see this as a common operation). Do you have any preference about where to store the patch (if you want to proceed this way)?

PR #19, opened 2021-02-16
Add golangci lint dep + .gitignore on workspace files

aaahrens:
The touched files that are formatted are vestigial from my master fork, they're the result of go fmt


ccontavalli:
So.. this change looks scary, as it brings in a lot of new dependencies, and updates quite a few as well.

I'm happy to merge it in, but can you break it out from the `go fmt` run? So if in 3 weeks from now we have a problem and end up bisecting or trying to figure out what made enkit or astore work in unexpected ways, we can easily roll back this change if necessary?

PR #27, opened 2021-02-25
[WIP] [Feedback] Adds testing harness, emulator, process and grpc support

aaahrens:
This adds the following procedures and features 

- Sets up a standard for linking external processes in tests and killing them
- Adds support for running datastore emulator and the proper procedure for killing it
- Adds support for running hermetic docker container (NOT CHECKED IN) 
- adds hermetic e2e testing with grpc server
- changes test package to be _test instead of inclusive packages to simulate sdk usage
- Begins README docs on testing


lib/khttp/ktest/mocks.go
```diff
@@ -0,0 +1,145 @@
...
+	datastoreBooted := make(chan bool)
+	//TODO concatenate stdout and stderr?
+	//the datastore emulator writes all logs to the error channel for some reason
+	scannerErr := bufio.NewScanner(outputStdErrPipe)
+	//TODO add timeout on starting, maybe 3 seconds?
+	go func() {
+		for scannerErr.Scan() {
+			if strings.Contains(scannerErr.Text(), "Dev App Server is now running") {
+				datastoreBooted <- true
+			}
```
ccontavalli, line 75: nit: would probably be convenient to `log.Printf` the emulator log, in case something goes wrong and we need to debug?

lib/khttp/ktest/mocks.go
```diff
@@ -0,0 +1,145 @@
...
+		}}
+	if err != nil {
+		return nil, killFunc, err
+	}
+	datastoreBooted := make(chan bool)
+	//TODO concatenate stdout and stderr?
+	//the datastore emulator writes all logs to the error channel for some reason
+	scannerErr := bufio.NewScanner(outputStdErrPipe)
+	//TODO add timeout on starting, maybe 3 seconds?
```
ccontavalli, line 70: This should be really easy and important to do thanks to your choice of using a channel below. I would consider addressing it immediately. Something like this around line 78 would work?

```
select {
            case ret := <-datastoreBooted:
                return &EmulatorDescriptor{...}
            case <-time.After(10 * time.Second):
                kill...
                return failed to start error.
            }
```

lib/khttp/ktest/mocks.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+type AStoreDescriptor struct {
+	Connection *grpc.ClientConn
+	Server     *astore.Server
+}
+
+//Run Astoreserver will spin up an emulated datastore along with an instance of the astore grpc server
+func RunAStoreServer() (*AStoreDescriptor, KillAbleProcess, error) {
+
```
ccontavalli, line 95: nit: empty line.

lib/khttp/ktest/mocks.go
```diff
@@ -0,0 +1,145 @@
...
+	killFunctions := KillAbleProcess([]func(){})
```
ccontavalli, line 96: why not just:
```
killFunctions ;= KillableProcess{}
```

lib/khttp/ktest/network.go
```diff
@@ -0,0 +1,45 @@
...
+		killFunc()
+	}
+}
+
+func (k *KillAbleProcess) AddKillable(process KillAbleProcess) {
+	if k == nil {
+		return
+	}
+	newList := append(*k, process...)
+	*k = newList
```
ccontavalli, line 38: just:
```
*k = append(*k, process...)
```

lib/khttp/ktest/network.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+
+func (k *KillAbleProcess) Add(p func()){
+	newList := append(*k, p)
+	*k = newList
```
ccontavalli, line 44: same as above.

lib/khttp/ktest/mocks.go
```diff
@@ -0,0 +1,145 @@
...
+	emulatorDescriptor, emulatorKill, err := RunEmulatedDatastore()
+	killFunctions.AddKillable(emulatorKill)
+	if err != nil {
+		return nil, killFunctions, err
+	}
+	err = os.Setenv(
+		"STORAGE_EMULATOR_HOST",
```
ccontavalli, line 103: who / what needs this environment variable? What is the effect of this variable? Can you add a comment? Like:

```
// Google library github.com/google/whatever/storage looks at a STORAGE_EMULATOR_HOST before connecting.
// When set, it causes the server to ignore ... configuration parameters, and ...
```

(reason I'm asking: certainly don't remember writing code looking at this env variable, I have no idea what the effects are)

lib/khttp/ktest/mocks.go
```diff
@@ -0,0 +1,145 @@
...
+		fmt.Sprintf("localhost:%d", emulatorDescriptor.Addr.Port))
+	if err != nil {
+		return nil, killFunctions, err
+	}
+	buffListener := bufconn.Listen(2048 * 2048)
```
ccontavalli, line 108: cool, I did not know about bufconn. This will come in handy.

lib/khttp/ktest/network.go
```diff
@@ -0,0 +1,45 @@
...
+func (k *KillAbleProcess) KillAll() {
+	fmt.Println("running kill functions")
+	for _, killFunc := range *k {
+		fmt.Println("running kill function")
+		killFunc()
+	}
+}
+
+func (k *KillAbleProcess) AddKillable(process KillAbleProcess) {
+	if k == nil {
```
ccontavalli, line 34: do we need this if? when/why?

Reason: it can cause errors to go unnoticed, and necessary kill functions to not run. Eg, for some reason k is nil, the test is run, killable functions are added... but never run!!

I would remove the if, and let this fail in case we try to add a killable to a nil KillableProcess. That way, there is no risk that we end up with successful code ignoring necessary cleanups.

something/BUILD.bazel
```diff
@@ -0,0 +1,3 @@
+exports_files([
```
ccontavalli, line 1: see previous comment for recommendation on credentials.json. Two main actions for you to do:
1) Please use the glob trick discussed. Having a `BUILD.bazel` file referencing a non-existent file is very likely to cause a build failure. We should have all build targets always succeed for ease of CI/CD pipeline.

2) Move the credentials directory somewhere else. Given that this is only for astore testing, I would put it somewhere under astore. Consider also that other parts of enkit, eg, the `proxy` in theory require credentials for testing, but different ones. So a top level credentials/ directory can end up being confusing, and require a global naming convention.

See other credentials directories scattered in the repo, also recommend adding a README, to indicate what isthe expected content of the directory.

lib/khttp/ktest/network.go
```diff
@@ -0,0 +1,45 @@
+// +build !release
```
ccontavalli, line 1: This file has some very helpful functions for testing, but are not necessarily for testing the khttp package only.
I don't feel very strongly here, as they are at least related to networking, and generally testing a server, so `khttp/ktest` may be ok. But I wouldn't object a top level `lib/testing` or `lib/ktesting`, with generic testing functions that are useful across other packages.

ccontavalli (commented):
Common themes:
1) Even though the go packages are all in one repo, I'm trying to keep package boundaries well defined. So eg, `lib/khttp` should only have functions and helpers related to making http requests. All `astore` stuff should be confined to the astore directory, ... see comments, asking you to shuffle a couple files around.

2) Most else is minor stuff, good PR.

lib/khttp/ktest/credentials.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"errors"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+//Check Credentials will walk runfiles until it reaches credentials.json and returns it
+func CheckCredentials() (string, error) {
```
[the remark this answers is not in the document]
aaahrens: The reason why I took it from the runfiles dir and didn't glob() the file is because you cannot fetch files outside the $CWD and its children (by design). Also it is a data dependency, just an absolute path instead of a glob. 

https://github.com/aaahrens/enkit/blob/astore_delete/astore/server/astore/BUILD.bazel#L35

I wanted to avoid what currently happening which is creating several folders with different credential files. I think it's much simpler to have one top level folder for credentials and fetch on demand.

lib/khttp/ktest/mocks.go
```diff
@@ -0,0 +1,145 @@
...
+
+func RunEmulatedDatastore() (*EmulatedDatastoreDescriptor, KillAbleProcess, error) {
+	emulatorAddr, err := AllocatePort()
+	if err != nil {
+		return nil, nil, err
+	}
+	cmd := exec.Command("gcloud",
+		"beta", "emulators", "datastore", "start",
+		"--no-store-on-disk",
+		fmt.Sprintf("--host-port=127.0.0.1:%d", emulatorAddr.Port),
```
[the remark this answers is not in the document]
aaahrens: In order:

- This works because listeners don't bind to multiple IP addresses. This is listening on ::<port>, while the emulator is binding to 127.0.0.1. The specific ":0" behaviour is in the tcp.Listen() godoc. 
- Listener is closed at the end of the test since it's bound to the go process. 
- In another upstream I have it returning the whole listener with address as an update, but I support I can do it here as well

astore/server/astore/astore_test.go
```diff
@@ -1,16 +1,27 @@
-package astore
+package astore_test
 
 import (
+	"context"
+	"errors"
+	"fmt"
 	"github.com/enfabrica/enkit/astore/client/astore"
+	rpcAstore "github.com/enfabrica/enkit/astore/rpc/astore"
+	astore2 "github.com/enfabrica/enkit/astore/server/astore"
```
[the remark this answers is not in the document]
aaahrens: refactored and resolved, was actually an autocomplete thing that got reverted.

astore/server/astore/astore.go
```diff
@@ -92,7 +92,7 @@ func (s *Server) List(ctx context.Context, req *astore.ListRequest) (*astore.Lis
 	// Two queries are necessary:
 	//   1) To retrieve artifacts.
 	//   2) To retrieve sub-paths.
-
+	fmt.Println("")
```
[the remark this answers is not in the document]
aaahrens: Removed

something/BUILD.bazel
aaahrens: See above comment, it is a data dependency, but an absolute path not globbed.

lib/khttp/ktest/network.go
aaahrens: I agree,  will refactor to knetworking package

lib/khttp/ktest/mocks.go
aaahrens: I usually don't do struct initializes for type alias because I don't want to obfuscate that it's an alias and not a struct.  

...but I'll do it anyway  +1

lib/khttp/ktest/mocks.go
aaahrens: Implemented

lib/khttp/ktest/mocks.go
aaahrens: It now returns the log at the end of the function if it errors out

lib/khttp/ktest/mocks.go
```diff
@@ -0,0 +1,145 @@
...
+
+type MinioDescriptor struct {
+	Port uint16
+}
+
+//RunMinioServer will spin up a minio serer using the local docker daemon
+//it also returns a func to call that will close and destroy the running image
+//the port and network bind are determined by docker and returned
+func RunMinioServer() (MinioDescriptor, func() error, error) {
+	_ = "gcloud beta emulators datastore start --no-store-on-disk"
```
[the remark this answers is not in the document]
aaahrens: removed

lib/khttp/ktest/mocks.go
aaahrens: https://github.com/googleapis/google-cloud-go/blob/86178128cae668f46d014569d5c25844aa379cdb/storage/storage.go#L116

All google go client (well most) take in some version that environment variable which lets it redirect from google's official endpoints. Unfortunately you can only configure it though os.Env and generally not through opts.

lib/khttp/ktest/network.go
aaahrens: We need this because it lets the testing library boil up killable funcs from other processes. and I agree, removed the if clause and just let it panic

lib/khttp/ktest/network.go
```diff
@@ -0,0 +1,41 @@
...
+func (k *KillAbleProcess) KillAll() {
+	fmt.Println("running kill functions")
+	for _, killFunc := range *k {
+		fmt.Println("running kill function")
+		killFunc()
+	}
+}
+
+func (k *KillAbleProcess) AddKillable(process KillAbleProcess) {
+
```
ccontavalli, line 34: nit: newline here.

lib/khttp/ktest/credentials.go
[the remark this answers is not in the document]
ccontavalli: ok, multiple orthogonal topics, let me try to break down the problem.

1) Regardless of agreeing or disagreeing with a global credentials directory vs one per folder being better, I would prefer to treat this as an orthogonal change. In this PR, I would use the same mechanism (separate credentials, glob()) that all other directories/subprojects are using. And then have a separate PR to change the mechanism, ideally for all projects at the same time, once we have something better ready.

Consider that we will need to re-factor this mechanism to allow for some sort of remote fetching of those credentials at build time - as right now credentials only live on my laptop. And if all use the same mechanism, it'll be easier to change.

2) The bazel way to refer to those files would be to create a rule for the top level credentials/ directory (eg, in the credentials directory itself, or parent directory). A `filegroup()` "packing" the `credentials/` for everyone to refer to. And then here in the BUILD file, just have a data dependency on that rule.

3) The credentials for different components are different - both in scope and configuration. As a habit, I think different components should have different credentials, scoped to allow the minimum set of things that the component is supposed to do.

Also, consider that some of the components we have are logically decoupled / separate, the fact that they are in the same repo is an implementation detail :)

lib/khttp/ktest/mocks.go
```diff
@@ -0,0 +1,145 @@
+// +build !release
```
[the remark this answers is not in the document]
ccontavalli: ping

lib/khttp/ktest/mocks.go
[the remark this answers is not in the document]
ccontavalli: ack.

lib/khttp/ktest/mocks.go
ccontavalli: makes sense. Minor nit: would add a // Causes the google-could-go/storage library to use a local emulator rather than the real endpoint. See https://github.com/googleapis/google-cloud-go/blob/86178128cae668f46d014569d5c25844aa379cdb/storage/storage.go#L109.

lib/khttp/ktest/mocks.go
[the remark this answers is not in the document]
aaahrens: Moved to astore_test package, using the fmt_test package naming convention for blackbox/behaviour testing

astore/astore_test/e2e_test.go
```diff
@@ -0,0 +1,77 @@
+package astore_test
```
ccontavalli, line 1: ah ... generally have a preference for keeping the tests in the same directory as the code, right next to the code, so it's more visible and evident to random developer that there is a test, and where the test is.

Can you move up directly in astore?

astore/astore_test/e2e_test.go
```diff
@@ -0,0 +1,77 @@
...
+		Architecture: "dwarvenx99",
+		Path:         "127.0.0.1:9000/hello/work/example.yaml",
+		Note:         "note",
+		Tag:          []string{"something"},
+	})
+	if err != nil {
+		t.Error(err.Error())
+	}
+
+	fmt.Println("finalizzing +%v", resp.Artifact)
```
ccontavalli, line 74: s/finalizzing/finalizing/

astore/astore_test/e2e_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"testing"
+)
+
+// TODO(aaahrens): fix client so that it's signed urls can depend on an interface for actual e2e testing
+func TestServer(t *testing.T) {
+	astoreDescriptor, killFuncs, err := RunAStoreServer()
+	if killFuncs != nil {
+		defer killFuncs.KillAll()
+	}
+	if err != nil {
```
ccontavalli, line 23: `assert.Nil(t, err)`

astore/astore_test/e2e_test.go
```diff
@@ -0,0 +1,77 @@
...
+		t.Fatal(err.Error())
+	}
+	// Running this as test ping feature.
+	client := astore.New(astoreDescriptor.Connection)
+	res, _, err := client.List("/test", astore.ListOptions{})
+	if err != nil {
```
ccontavalli, line 29: same

astore/astore_test/e2e_test.go
```diff
@@ -0,0 +1,77 @@
...
+		t.Error(err.Error())
+	}
+	fmt.Printf("list response is +%v \n", res)
+	b, err := ioutil.ReadFile("./testdata/example.yaml")
+	if err != nil {
```
ccontavalli, line 34: same

astore/astore_test/e2e_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	ctxWithLogger := ccontext.DefaultContext()
+	ctxWithLogger.Logger = logger.DefaultLogger{Printer: log.Printf}
+	ctxWithLogger.Progress = progress.NewDiscard
+
+	uploadOption := astore.UploadOptions{
+		Context: ctxWithLogger,
+	}
+	u, err := client.Upload(uploadFiles, uploadOption)
+	if err != nil {
```
ccontavalli, line 50: to maintain the error message:
`assert.Nil(t, err, "client upload failed with %s", err)`

astore/astore_test/e2e_test.go
```diff
@@ -0,0 +1,77 @@
...
+		t.Error(err.Error())
+		fmt.Println("erroring in client upload")
+	}
+
+	fmt.Printf("upload is +%v \n", u)
+	storeResponse, err := astoreDescriptor.Server.Store(context.Background(), &rpcAstore.StoreRequest{})
+	if err != nil {
```
ccontavalli, line 57: same

astore/astore_test/e2e_test.go
```diff
@@ -0,0 +1,77 @@
...
+		t.Fatal(err)
+	}
+	if storeResponse.GetSid() == "" || storeResponse.GetUrl() == "" {
```
ccontavalli, line 60:
```
assert.Equal(t, "", storeResponse.GetSid())
assert.Equal(t, "", storeResponse.GetURL())
```

astore/astore_test/e2e_test.go
```diff
@@ -0,0 +1,77 @@
...
+		t.Fatal(errors.New("invalid store response"))
+	}
+	resp, err := astoreDescriptor.Server.Commit(context.Background(), &rpcAstore.CommitRequest{
+		Sid:          storeResponse.GetSid(),
+		Architecture: "dwarvenx99",
+		Path:         "127.0.0.1:9000/hello/work/example.yaml",
+		Note:         "note",
+		Tag:          []string{"something"},
+	})
+	if err != nil {
```
ccontavalli, line 70: same

lib/khttp/ktest/mocks.go
```diff
@@ -0,0 +1,145 @@
...
+		"--no-store-on-disk",
+		fmt.Sprintf("--host-port=127.0.0.1:%d", emulatorAddr.Port),
+		"--quiet")
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+	outputStdErrPipe, err := cmd.StderrPipe()
+	if err != nil {
+		return nil, nil, err
+	}
+	err = cmd.Start()
+	killFunc := []func(){
```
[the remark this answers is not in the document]
ccontavalli: ping?

README.md
```diff
@@ -0,0 +1,45 @@
...
+        PLEASE NOTE: do not install using snap/brew/apt-get etc
+        emulators do not work
+    * After following the instructions to install here
+    * run the following command to get access to the emulators
+        > gcloud components install beta 
+    * Add the gcloud binary to the local binaries directory with the following symlink
+        > ln -s $(which gcloud) /usr/local/bin
+                    
+2. Get a service account from <x, Y, Z person>
+    * Put it in credentials/credentials.json
```
ccontavalli, line 19: testdata/credentials.json?

README.md
```diff
@@ -0,0 +1,45 @@
...
+* Running Everything 
+> bazel test //...
+
+##### Adding Tests
+1. Create the test in * _test.go 
+2. Run 
+    > bazel run //:gazelle
+* if your test needs server dependencies, such as astore or minio 
+    1. Tests must be run as local = True 
+    2. Test must also include the target "//credentials:credentials.json"
```
ccontavalli, line 35: same here.

ccontavalli (commented):
almost there! will happily merge after this last round!

README.md
aaahrens: I just delete 2 because gazelle should handle testadata

ccontavalli (approved):

PR #34, opened 2021-03-09
Add kcerts and tls/rsa certificate support

aaahrens:
Does the following things:

- adds golang.org/crypto/ssh package 
- adds CA generation
- adds intermediate cert generation
- adds ssh key generation 
- bullet proof testing

FEEDBACK

When creating an RSA certificate, should there by an option to password encrypt it?

### DIFF
--- a/lib/kcerts/certs.go
+++ b/lib/kcerts/certs.go
@@ -1,0 +1,2 @@
+package kcerts
+
--- a/lib/kcerts/certs.go
+++ b/lib/kcerts/certs.go
@@ -1,2 +1,202 @@
 package kcerts
 
+import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"errors"
+	"math/big"
+	"net"
+	"time"
+)
+
+//GenerateNewCARoot returns the new
+func GenerateNewCARoot(opts *CertOptions) (*x509.Certificate, []byte, *rsa.PrivateKey, error) {
+	if !opts.isValid {
+		return nil, nil, nil, errors.New("must call Validate() on certificate options before creating a CA")
+	}
+	rootTemplate := x509.Certificate{
+		SerialNumber: big.NewInt(1),
+		Subject: pkix.Name{
+			Country:      opts.Country,
+			Organization: opts.Organization,
+			CommonName:   "Root CA",
+		},
+		NotBefore:             opts.Before,
+		NotAfter:              opts.After,
+		KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageCRLSign,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		BasicConstraintsValid: true,
+		IsCA:                  true,
+		MaxPathLen:            2,
+		IPAddresses:           opts.IPAddresses,
+	}
+	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	certBytes, err := x509.CreateCertificate(rand.Reader, &rootTemplate, &rootTemplate, &privateKey.PublicKey, privateKey)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	b := pem.Block{Type: "CERTIFICATE", Bytes: certBytes}
+	certPEM := pem.EncodeToMemory(&b)
+
+	return &rootTemplate, certPEM, privateKey, nil
+}
+
+//GenerateIntermediateCertificate
+func GenerateIntermediateCertificate(opts *CertOptions, RootCa *x509.Certificate, RootCaPrivateKey *rsa.PrivateKey) (*x509.Certificate, []byte, *rsa.PrivateKey, error) {
+	intermediatePrivateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	if !opts.isValid {
+		return nil, nil, nil, errors.New("must call Validate() on certificate options before creating a DCA")
+	}
+	intermediate := x509.Certificate{
+		SerialNumber: big.NewInt(1),
+		Subject: pkix.Name{
+			Country:      opts.Country,
+			Organization: opts.Organization,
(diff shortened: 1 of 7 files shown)


lib/kcerts/certs.go
```diff
@@ -0,0 +1,204 @@
...
+	if o.After.Sub(currTime).Hours() < 24*365 { // hours in a year
+		return errors.New("duration of the CA is too low")
+	}
+	if len(o.Organization) == 0 {
+		return errors.New("must set organization")
+	}
+	if len(o.Country) == 0 {
+		return errors.New("must set countries of origin for en_Lang and i8n support")
+	}
+	o.isValid = true
```
ccontavalli, line 198: (optional: I'd be ok with merging as is, and iterate later - let me know your thoughts)

I'm not a big fan of o.isValid. The reason is that 1) user of the API must remember to call it, and 2) parameters can be changed after validation, accidentally, and 3) it's tricky to provide optional computed parameters (less relevant here, but somewhat). Eg:

```
opts := NewOptions()
opts.Before = ...
opts.After = ...
opts.... =

opts.Validate()
opts.Before = ...
``` 
I've not been consistent either in the rest of the code base, but I'm a fan of the "Functional Options" pattern (google it, a lot of good presentations and talks), that can prevent errors here and allow computed defaults. 

This is how your code would look like:
```
// Attributes are private to this package.
// Could even mark the structure private? Up to you.
type CertOptions struct {
	country      []string
	organization []string
	before       time.Time
	after        time.Time
	ipAddresses  []net.IP
}

type Modifier func (*CertOptions) error

func WithCountries(cs []string) Modifier {
	return func (o *CertOptions) error {
          o.Country = cs
          return nil
	}
}


func NewOptions(mods ...Modifier) (*CertOptions, error) {
  opts := &CertOptions{
    Before: time.Now(), // Some defaults if necessary
  }
  for _, mod := range mods {
    if err := mod(opts); err != nil {
      return nil, err
    }
  }
  // Update defaults based on other defaults, if necessary.
  // eg, bad idea here, but if opts.After == 0 { opts.After = opts.Before + 7 * 24 * time.Hour }, default 7 days validity.
  if err := Validate(opts); err != nil {
    return nil, err
  }
  return opts, nil
}
```

This way, it is impossible to generate an invalid CertOptions, and no need to check for validation further down.
It also allows to provide reasonable default in the API.

If parameters are mandatory, just make them as mandatory arguments for the `NewOptions` function, so build time failure rather than run time.

lib/kcerts/certs.go
```diff
@@ -0,0 +1,204 @@
...
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"errors"
+	"math/big"
+	"net"
+	"time"
+)
+
+// GenerateNewCARoot returns the new certificate anchor for a chain. This should ideally only be called once as rotating
```
ccontavalli, line 15: `go fmt` in the directory?

ccontavalli (commented):
Great majority of the comments is style nits. Address them, and I think this can be merged in.

Only serious comment is about "functional options". I have a preference for that pattern, for reason explained, and consistency with the repo. But I'd be ok merging as is and changing it if we see a need over time.

lib/kcerts/certs.go
aaahrens: The reason why I generally prefer the `func(r *Reciever) *Reciever` configuration method is actually for onboarding/devtooling.

Mainly, when I hit `kcerts.NewOptions().` the period would autocomplete in vim/vscode/idea etc and give the developer all the options available without having to trawl through the original code. 

I kinda screwed up not having the exported funcs use validate. I should not be using isValid under any circumstance and it's really bad smell. 

I'm fine using the functional modifiers, I find it incredibly irritating that even in the grpc and google-client package they use both methodologies.

aaahrens: If you're not swayed by the tooling argument I'll just convert to the functional modifiers, either way is standard practice and I have no issues with using it.

lib/kcerts/certs_test.go
```diff
@@ -0,0 +1,126 @@
...
+func TestCertSuite(t *testing.T) {
+	opts := kcerts.NewOptions().
+		NotValidBefore(time.Now().Add(-5 * time.Minute)).
+		ValidUntil(time.Now().AddDate(3, 0, 0)).
+		WithIpAddresses([]net.IP{net.ParseIP("127.0.0.1"), net.ParseIP("0.0.0.0")}).
+		WithOrganizations([]string{"Test Corp"}).
+		WithCountries([]string{"US"})
+
+	err := opts.Validate()
+	if err != nil {
```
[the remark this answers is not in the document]
aaahrens: Update to using the assert library.

Just a side note: I usually liberally use assert when table testing for stack traces in test tables. But generally for unit testing I find that the default testing lib is enough.

lib/kcerts/certs_test.go
```diff
@@ -0,0 +1,122 @@
...
+		WithIpAddresses([]net.IP{net.ParseIP("127.0.0.1"), net.ParseIP("0.0.0.0")}).
+		WithOrganizations([]string{"Test Corp"}).
+		WithCountries([]string{"US"})
+
+	rootCert, rootPem, rootPrivateKey, err := kcerts.GenerateNewCARoot(opts)
+	if err != nil {
+		assert.Error(t, err)
+	}
+	intermediateCert, intermediatePem, intermediatePrivateKey, err := kcerts.GenerateIntermediateCertificate(opts, rootCert, rootPrivateKey)
+	if err != nil {
```
ccontavalli, line 31: assert.Nil(t, err)

lib/kcerts/certs.go
```diff
@@ -145,58 +135,69 @@ func GenerateSSHKeyPair(privateKey *rsa.PrivateKey) ([]byte, []byte) {
...
-	return nil
 }
 
-func NewOptions() *CertOptions {
-	return &CertOptions{}
+func NewOptions(mods ...Modifier) (*certOptions, error) {
+	co := &certOptions{}
+	for _, mod := range mods {
+		err := mod(co)
+		if err != nil {
```
ccontavalli, line 198: minor nit:
```
if err := mod(co); err != nil {
  return nil, err
}
```
is often considered better golang style.

ccontavalli (approved):
Looks great, thank you!

PR #39, opened 2021-03-11
[WIP] Machinist Draft

aaahrens:


machinist/node.go
```diff
@@ -0,0 +1,161 @@
...
+		pollResponse, err := pollStream.Recv()
+		if err != nil {
+			n.Kill(err)
+		}
+		n.PongChannel <- pollResponse
+	}
+
+}
+
+func (n *Node) BeginPolling() {
```
ccontavalli, line 60: some of this code was already checked in / semi-working. I was expecting to see a diff over the existing files, rather than entirely new files? Not sure how the new code bodes with the existing one otherwise.

machinist/node.go
```diff
@@ -0,0 +1,161 @@
...
+	n.KillChannel <- NodeErr{}
+}
+
+func (n Node) Stop() {
+	n.KillChannel <- NodeErr{}
+}
+
+func WithInviteToken(token string) NodeModifier {
+	return func(node *Node) error {
+		data, err := base64.RawStdEncoding.DecodeString(token)
```
ccontavalli, line 122: when we last chat, I was expecting the invitation token to be a short string? this looks like it'll be a large blob of data?

machinist/server.go
```diff
@@ -0,0 +1,258 @@
...
+type invitationToken struct {
+	Addresses  []string
+	Port       int
+	CRT        string
+	PrivateKey string
+	RootCA     string
+}
+
+func (s Server) GenerateInvitation(tags map[string]string, name string) ([]byte, error) {
+	nats, err := net.Interfaces()
```
ccontavalli, line 109: The set of ip addresses (or domain names) will need to come from a configuration file. If we run this in a docker container, on GCP, or even on our own internal firewall, the IP addresses of the interfaces will have nothing to do with the address configured on the network interfaces of the machines.

(eg, on GCP, it'll be a public GCP IP address, firewall will have internal 10 addresses rather than  public ones, docker runs by default - similar to k8s? - with a different IP per container, and has its own magic to remap it).

I strongly recommend breaking this down in two separate functions, one that generates the invitation, and another to ... read/compute/... the list of names/ips.

machinist/server.go
```diff
@@ -0,0 +1,258 @@
...
+		if err != nil {
+			return err
+		}
+		server.NetAddr = na
+		return nil
+	}
+}
+
+
+type PortDescriptor struct {
```
ccontavalli, line 240: this should go in a test library? why is this necessary?

PR #40, opened 2021-03-19
bazel/utils: Add bazel hook for clang-format

stefanoen:
The clang_format rule can be used to define an executable target which
will format all files matching the specified patterns using the specified
style.

Example of use:

**WORKSPACE**
```
http_file(
    name = "clang-format",
    urls = ["https://clang-format-url",],
    executable = True,
)
```

**BUILD.bazel**
```
clang_format(
    name = "cc_format",
    style = "Google",
    pattern_to_format = "*.cc",
    clang_format = "@clang-format//file",
)
```

Running `bazelisk run :cc_format` will format (Google style) all `*.cc` files in the Bazel repository.


bazel/utils/clang_format.bzl
```diff
@@ -0,0 +1,48 @@
...
+        is_executable = True,
+    )
+
+    return DefaultInfo(
+        files = depset([out_file]),
+        runfiles = ctx.runfiles(files = [clang_format_file]),
+        executable = out_file,
+    )
+
+clang_format = rule(
```
ccontavalli, line 22: `doc = ...` for the entire rule? How is one expected to use this rule?

bazel/utils/run_clang_format.template
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
ccontavalli, line 1: I recommend for templates to have the correct extension, so that editors still pick up the language correctly without manual work, eg, rename to `run_clang_format.template.sh`?

bazel/utils/clang_format.bzl
```diff
@@ -0,0 +1,48 @@
...
+        ),
+        # TODO: Support a list of patterns to format.
+        "pattern_to_format": attr.string(
+            mandatory = True,
+            doc = "The name pattern of the files to format.",
+        ),
+        "clang_format": attr.label(
+            allow_single_file = True,
+            mandatory = True,
+            doc = "Label of the clang-format executable to use.",
```
ccontavalli, line 39: where is `clang-format` going to come from? Do you have WORKSPACE stuff that can pull it in? If yes, can you recommend the use in the docstring here? If no, how do you plan to bring it in?

bazel/utils/clang_format.bzl
```diff
@@ -0,0 +1,48 @@
...
+    implementation = _clang_format,
+    attrs = {
+        "style": attr.string(
+            mandatory = True,
+            # From clang-format-7 --help output.
+            values = ["LLVM", "Google", "Chromium", "Mozilla", "WebKit", "file"],
+            doc = "The coding style to apply.",
+        ),
+        # TODO: Support a list of patterns to format.
+        "pattern_to_format": attr.string(
```
ccontavalli, line 32: I would just call this "pattern" (or patterns). How do we do ".cc", ".h", ".cpp", ".hpp", ".c" file if only one pattern is supported? Well, I guess we can write a more advanced pattern for find? Is that what you had in mind?

bazel/utils/clang_format.bzl
```diff
@@ -0,0 +1,48 @@
...
+
+clang_format = rule(
+    implementation = _clang_format,
+    attrs = {
+        "style": attr.string(
+            mandatory = True,
+            # From clang-format-7 --help output.
+            values = ["LLVM", "Google", "Chromium", "Mozilla", "WebKit", "file"],
+            doc = "The coding style to apply.",
+        ),
```
ccontavalli, line 30: Sorry I'm expanding the scope for this... but I would strongly recommend this supported passing a Label pointing to a `.clang-format` file. Eg, we have one such file in our `internal` repository. And further... we will necessarily have different styles per directory. Eg, kernel code needs to be formatted differently.

bazel/utils/run_clang_format.template
```diff
@@ -0,0 +1,14 @@
...
+
+# Retrieve the absolute path from the relative path passed by Bazel.
+CLANG_FORMAT=$(realpath {clang-format})
+
+# $BUILD_WORKSPACE_DIRECTOR is made available by Bazel to executables
+# (see https://docs.bazel.build/versions/master/user-manual.html#run).
+if ! cd "$BUILD_WORKSPACE_DIRECTORY"; then
```
ccontavalli, line 8: Should this actually start from ... the directory where the rule is? Problem: we have `sdk/lib` with c++ files formatted one way, and `sdk/driver` with .c files formatted a different way. How to exclude some directories? how to use different styles per directory?

Not sure what the simplest solution is, but one that comes to mind is to actually have a dict of {"directory": "clang-format label file to use"}, start the find from the specified directory.

But still, the example above would work very poorly, when one directory is a subdirectory of another :/

Maybe not specify the `.clang-format` file as a label at all, just ... find `.clang-format` files first, and then apply clang-format recursively in those directories?

bazel/utils/clang_format.bzl
stefanoen: Right now the solution is to have some mirror that distribute the clang-format executable, and to download that from the WORKSPACE file using `http_file`. I have added a description of that in the docstring.

p.s. I've searched for some sort of release channel for clang-format, but I couldn't find any. IMHO, this should be good enough for now.

bazel/utils/clang_format.bzl
stefanoen: So, as of now the solution is to simply have multiple `clang_format` rules instantiations. Kinda ugly, but I'm not exactly sure about all our possible use cases. Anyway, yes, the simplest solution would be to pass a regex to the `find` command (this requires a small change in the bash script).

bazel/utils/clang_format.bzl
stefanoen: I've added an additional label to support multiple `.clang-format` files.

bazel/utils/run_clang_format.template
stefanoen: See https://github.com/enfabrica/enkit/pull/40#discussion_r599084808. My idea is that this should be enforced per-language, not per-directory.

ccontavalli (approved):

PR #41, opened 2021-03-23
lib/kflags, lib/client: fix initialization bug.

ccontavalli:
Problem:
Current flag initialization logic has the default flag values
set in the .go code, (possibly) overridden by env variable,
(possibly) overridden by parameters downloaded via http/astore.
Finally, the cobra handler parses the flags, and the final
value is set.

Every time flags change values, some objects need to be
re-created. For example: if log verbosity is changed, the
logger object needs to be re-created (well, or updated).

Before this change:
- the logger object was re-created when the DEFAULTS were
  changed, but not when the final set of flags was loaded.

  Pretty much causing some flags to be ignored when passed
  from the command line (but not when defaulted from env
  or from http configurations).

After this change:
- actually plumbed the object initialization logic with
  current flag parsers.

### DIFF
--- a/lib/client/client.go
+++ b/lib/client/client.go
@@ -122,7 +122,7 @@ func (bf *BaseFlags) IdentityErrorHandler(message string) kflags.ErrorHandler {
 }
 
 func (bf *BaseFlags) IdentityStore() (identity.IdentityStore, error) {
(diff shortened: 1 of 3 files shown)


aaahrens (approved):
LGTM

PR #42, opened 2021-03-23
proxy/ptunnel: re-load credentials at every connection attempt.

ccontavalli:
Problem:
When credentials expire, even if refreshed, "enkit tunnel" is unable
to recover the session. This is because credentials are never reloaded
from disk.

In this change:
- have commands/tunnel.go set things up so credentials are re-loaded
  every time we attempt a connection, so the latest are always used.
- ... but stick to the last known valid credentials if loading fails.

Additionally:
- clarify logging slightly, so that the problem is clear for someone
  looking at the log.


aaahrens (approved):
LGTM, although we should figure out testing for this later on

PR #43, opened 2021-03-23
proxy/ptunnel,nasshp: significantly improved cleanup and error handling.

ccontavalli:
Background:
Before this PR, the cleanup code on tunnel close was sketchy at best,
potentially leaking goroutines, leaving connections opened, and not
really propagating close() events correctly.

In this PR:
- try to propagate close and shutdown events correctly from web sockets.
  If the web socket stops being readable, do a shutdown for write.
  If the web socket stops being writable, do a shutdown for read.
  This resulted in adding a couple utilities in knetwork/shutdown.go.

- close the client connection if the websocket dies.
  Apart from the shutdown, the new code tries to close the socket
  correctly when all is done.

- close the websocket with the server, and try to ensure the corresponding
  routines. This is very tricky to get right, as the websockets library
  used doesn't really have non-blocking or interruptible read / write
  calls. The code relies on a) injecting error, eg, ensuring that the
  next poll fails, and b) closing the web socket.

Additionally:
- make the Waiter object in blocking window much safer to close/fail.

  Specifically:
  - before this change, if Fail() was called with no reader, it would
    block forever (simple channel semantics). If multiple threads were
    waiting on the window (illegal by contract, but not enforced), only
    one would fail. If a reader raced and used the object after failing
    it, the reader would still move forward like nothing happened.

  - after this change, Fail() will mark the object as failed forever,
    without blocking. Any reader/waiter will unblock immediately (yes,
    even multiple readers) and get the error immediately. If the
    object is re-used, the reader will still get the error immediately.

### DIFF
--- a/lib/knetwork/BUILD.bazel
+++ b/lib/knetwork/BUILD.bazel
@@ -1,0 +1,8 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
+
+go_library(
+    name = "go_default_library",
+    srcs = ["shutdown.go"],
+    importpath = "github.com/enfabrica/enkit/lib/knetwork",
+    visibility = ["//visibility:public"],
+)
--- a/lib/knetwork/shutdown.go
+++ b/lib/knetwork/shutdown.go
@@ -1,0 +1,55 @@
+package knetwork
+
+import (
+	"io"
+)
+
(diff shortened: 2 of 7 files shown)


aaahrens (approved):
LGTM

PR #45, opened 2021-03-23
lib/goroutine, lib/multierror: minor API improvements.

ccontavalli:
Specifically:
WaitAll() now returns a multierror rather than a plain array
of errors, only if an error was actually received.

multierror is expanded to have support for sparse arrays of errors,
eg, an array with some entries being nil (all good) and some entries
being actual errors.

Why?
1 - This allows WaitAll() to be used just like any other function,
    it just returns an error, or nil if there are no errors.
    Before, it always returned an array.

2 - This allows to chain WaitFirstError() and WaitAll() arbitrarily
    Like: WaitAll(WaitFirstError(func1, WaitAll(func5, func6)), WaitAll(func3, func4))
    and still have error propagation work correctly.

The cost on multierror should be negligible: people that were using
a contiguous array should have no additional cost. Sparse arrays are
slightly more costly as they need to be scanned to determine if they
contain an error or not.

### DIFF
--- a/lib/goroutine/BUILD.bazel
+++ b/lib/goroutine/BUILD.bazel
@@ -5,4 +5,5 @@ go_library(
     srcs = ["gochan.go"],
     importpath = "github.com/enfabrica/enkit/lib/goroutine",
     visibility = ["//visibility:public"],
+    deps = ["//lib/multierror:go_default_library"],
 )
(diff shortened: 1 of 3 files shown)


aaahrens (approved):
This is LGTM to me. 

Ideally later on I want to move to logrus for structured logging.

PR #48, opened 2021-03-24
Working generic SPA added to tunnel frontend

aaahrens:
So this PR is essentially the rules_nodejs example repo that's been updated to be in a subdirectory and updated for the most recent CRA. 

The dev server runs and you can talk to backends with it, build and test also work.  

Talking points:
- It's possible to use a shared package.json for all projects and write bazel rules to configure environment variables. Not sure if it's worth the overhead? 
- Not sure if we should add an extra config option later on that appends rewrite-proxy options to the frontend.

Was working on getting the minified build embedded in a go binary on prod deployment, but that'll be another largish rule later on in the week after adding support for tagging. (note: need tagging to avoid rebuilding on every dev restart)


ccontavalli:
This looks like a great beginning. The only comments I would add are related to file paths, and using a single `yarn.lock` and packages file for the whole project (don't want different parts of the repo to use different versions of the same libraries).

My thinking right now is to merge this as is, and then adjust as we actually implement the interface? If this sounds reasonable, I'll just merge this in.

aaahrens:
I'm fine to merge as it is. 

The one thing I'm worried about in having a shared package is propagating errors if different libraries are in conflict with each other. Generally I'm in favor of it though.

ccontavalli:
> The one thing I'm worried about in having a shared package is propagating errors if different libraries are in conflict with each other. Generally I'm in favor of it though.

Yes. I can see both sides. My general take is that "forking out" (eg, to use a different library version) is generally easier than "forking in" (eg, to all use the same library). Further, sharing forces us to... try to rely on the same dependencies, which in turn a) makes code sharing simpler (eg, imagine we have a .ts library that requires npm package version X, and another .ts library requiring the same package version Y, and now having a project requiring both libraries), b) if a bug is fixed in one place, it's fixed everywhere, c) if we need to update in one place, we'll generally try to update the entire repo, d) if somebody wants a library to do something similar to a library we already have, there will be natural pressure to ... not introduce new libraries.

However, we still retain the ability to fork if we need to. Of course, this is very subjective, and we may change our mind in the future, but working this way has worked very well for me in the past.

PR #49, opened 2021-03-26
bazel/linux/defs.bzl: Fix kernel_module bug

stefanoen:
As a default "srcs" parameter, consider only the makefile parameter and
all the .c and .h files belonging to the package where the kernel_module
rule has been instatiated.

Currently, if the user doesn't specify the "srcs" parameter of the
kernel_module rule, the implementation will consider all files belonging
to the package to be the source files for that kernel module.
If for some reason, a file with the same name of one of the intermediate
build files pollutes the package, bazel will try to copy/link that over in
the build dir. This seems to cause a compilation failure since that file
could already exist there and that filesystem is read-only?

The bug can be reproduced, for example, by copying the module_name.mod.c
in the package source tree and modifying it (or the kernel module sources).
The error will look like this:

redacted: Read-only file system
make[2]: *** [redacted/Makefile.modpost:111: redacted/Module.symvers] Error 1
make[1]: *** [redacted/Makefile:1772: single_modpost] Error 2
make: *** [redacted/Makefile:185: __sub-make] Error 2

### DIFF
--- a/bazel/linux/defs.bzl
+++ b/bazel/linux/defs.bzl
@@ -226,7 +226,9 @@ def kernel_module(*args, **kwargs):
     Args:
       srcs: list of labels, specifying the source files that constitute the kernel module.
             If not specified, kernel_module will provide a reasonable default including all
-            files that are typically part of a kernel module.
+            files that are typically part of a kernel module (i.e., the specified makefile
+            and all .c and .h files belonging to the package where the kernel_module rule
+            has been instantiated, see https://docs.bazel.build/versions/master/be/functions.html#glob).
       module: string, name of the output module. If not specified, kernel_module will assume
             the output module name will be the same as the rule name. Also, it normalizes the
             name ensuring it has a '.ko' suffix.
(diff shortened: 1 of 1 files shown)


bazel/linux/defs.bzl
```diff
@@ -237,17 +239,19 @@ def kernel_module(*args, **kwargs):
       kernels: list of kernel (same as above). kernel_module will instantiate multiple
             kernel_module_rule, one per kernel, and ensure they all build in parallel.
     """
+    if "makefile" not in kwargs:
+        kwargs["makefile"] = "Makefile"
+
     if "srcs" not in kwargs:
-        kwargs["srcs"] = native.glob(include = ["**/*"], exclude = BUILD_LEFTOVERS, allow_empty = False)
+        include = ["**/*.c", "**/*.h", kwargs["makefile"]]
+        exclude = BUILD_LEFTOVERS + ["**/*.mod.c"]
```
ccontavalli, line 247: Can you fix BUILD_LEFTOVERS instead? Eg, in theory *.mod.c is already in there? It's probably not working either because of the missing "**/" or the "[co]" argument.

stefanoen: Are you okay with the changes to the include pattern? Or do you prefer to keep it `**/*`?

PR #52, opened 2021-03-29
Adding in ssh utilities for Machinist

aaahrens:
Adds smaller ssh-agent functionality, and the ability to write the known hosts to add a cert authority


lib/kcerts/ssh.go
```diff
@@ -0,0 +1,116 @@
...
+		return err
+	}
+	defer knownHostsFile.Close()
+	for _, dns := range hosts {
+		publicFormat := fmt.Sprintf("%s %s %s", CAPrefix, dns, caPublic)
+		if !strings.Contains(string(existingKnownHostsContent), publicFormat) {
+			_, err = knownHostsFile.WriteString(publicFormat)
+			if err != nil {
+				fmt.Println("error is not nil", err.Error())
+				return err
```
ccontavalli, line 53: I'm a big fan of the [new error handling introduced in go1.13](https://blog.golang.org/go1.13-errors). 

Let's say the disk is full, and this error is printed in the logs. You go debug it, all you'll see is "no space on device".

My recommendation is to return:
`fmt.Errorf("could not add key %s to known_hosts file: %w", err)`

(note the '%w', and err passed as is) 
This would generate a very readable and much more informative error message (now I know the implication of no space on disk), while still allowing easy programmatic access by using the `Unwrap` or `errors.Is`, `errors.As` API.

aaahrens: At some point I'm going to have to bump the go version.
Been using the go's errors.Wrap instead of the fmt for a while, super nice. 
https://pkg.go.dev/github.com/pkg/errors#Wrap

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		return err
+	}
+	knownHostsFile, err := os.OpenFile(qualifiedKnownHosts, os.O_APPEND|os.O_WRONLY, 0644)
+	if err != nil {
+		return err
+	}
+	defer knownHostsFile.Close()
+	for _, dns := range hosts {
+		publicFormat := fmt.Sprintf("%s %s %s", CAPrefix, dns, caPublic)
```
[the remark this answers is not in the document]
aaahrens: I thought this too, but each call to WriteString actually appends a newline, I think it has to do with the epipecheck at the end of the write function. 

If you want to run the ssh_test a couple of times and triple check, go ahead. (right now it writes to the users home known_hosts file)

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,116 @@
...
+		if !strings.Contains(string(existingKnownHostsContent), publicFormat) {
```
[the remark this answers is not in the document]
aaahrens: I actually disagree because most linters actually forbid noop continue and breaks in for loops. But implmented.

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,116 @@
...
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+)
+
+const CAPrefix = "@cert-authority"
+const SSHDir = ".ssh"
+const KnownHosts = "known_hosts"
```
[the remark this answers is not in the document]
aaahrens: The reason why I'm fine with how it is now is because since each go test is essentially an isolate we can just change the homedir in each test to some tmp file and call it good.  That way it mimics somewhat its real case. 

If you think eventually kcerts is going to move to more of an ssh/ecdsa/rsa utility package I would definitely refactor into a service later on.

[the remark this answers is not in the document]
aaahrens: Especially if we end up isolating ALL ssh dependencies into $HOME/.enkit (which is a cool idea). That way the end user never has to deal with local .ssh screwing up with other projects.

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,116 @@
...
+// For each entry, it adds an additional line and does not concatenate
+func AddSSHCAToClient(publicKey ssh.PublicKey, hosts []string) error {
+	hDir, err := homedir.Dir()
+	if err != nil {
+		return err
+	}
+	hDir = hDir + "/"
+	if _, err := os.Stat(hDir + SSHDir); os.IsNotExist(err) {
+		return fmt.Errorf("ssh directory %s does not exist, please create it", hDir+SSHDir)
+	}
```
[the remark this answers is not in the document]
aaahrens: See below

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,116 @@
...
+	qualifiedKnownHosts := hDir + SSHDir + "/" + KnownHosts
+	if _, err := os.Stat(qualifiedKnownHosts); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
aaahrens: See below

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,116 @@
...
+		return fmt.Errorf("ssh authorized hosts file %s does not exist, please create it", qualifiedKnownHosts)
+	}
+	caPublic := string(ssh.MarshalAuthorizedKey(publicKey))
+	existingKnownHostsContent, err := ioutil.ReadFile(qualifiedKnownHosts)
```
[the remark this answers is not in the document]
aaahrens: Regarding both the os.Stat and ReadFile 

I disagree with this because there isn't a valuable reason to ignore the errors. If we ignore the errors and don't os.Stat: The linter complains, it provides less info for the end user, and can possibly obfuscate errors. Generally I like to provide as much stdout info as possible especially when mucking about files users routinely touch. 

That being said, it should try and make them if it doesn't exist, so I implemented that with proper errors.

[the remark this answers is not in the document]
ccontavalli: > I disagree with this because there isn't a valuable reason to ignore the errors.
> If we ignore the errors and don't os.Stat: The linter complains, it provides less info for the end user, and can possibly obfuscate errors. Generally I like to provide as much stdout info as possible especially when mucking about files users routinely touch.

I don't disagree with the goals, but I think the code here now is significantly more complex than it could be, and is actually making the errors harder.

Doing Stat() and then the operation (eg, Read...(), Open...(), Write...()) is common in some languages, but should generally be frowned upon as an anti-pattern: Stat() can fail (or succeed) for reasons different from that of the operation, and introduces race conditions (in between Stat() and Open(), for example, the file may have been removed). Instead of one possible error, you now have two, and both need to be managed (as, for example, a Stat() succeeding and indicating a file exists and is readable does not imply the next Open() will also succeed and allow you to read the file - fs acls, syscall filtering, ...).

Instead of using Stat() to check if a file can be written or read, just try to write or read into the file, and *check the returned error code*.

In the suggestion I had made above, I just wanted your code to work correctly in the absence of the known hosts file (perfectly legal), absence of directory, empty file, ...

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,125 @@
...
+
+// AddSSHCAToClient adds a public key to the $HOME/.ssh/known_hosts in the ssh-cert x509.1 format.
+// For each entry, it adds an additional line and does not concatenate
+func AddSSHCAToClient(publicKey ssh.PublicKey, hosts []string) error {
+	hDir, err := homedir.Dir()
+	if err != nil {
+		return err
+	}
+	sshDir := filepath.Join(hDir, SSHDir)
+	if _, err := os.Stat(sshDir); os.IsNotExist(err) {
```
ccontavalli, line 31: see larger explanation below - but this block, line 31 to 37, is unnecessary. If the directory does not exist, your attempt to read the `known_hosts` file will probably return "No such file or directory", which is perfectly clear.

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,125 @@
...
+		fmt.Printf("ssh directory %s does not exist, attempting to create it", sshDir)
+		err = os.Mkdir(sshDir, 0700)
+		if err != nil {
+			return fmt.Errorf("could not create dir %s err: %w", sshDir, err)
+		}
+	}
+	knownHosts := filepath.Join(sshDir, KnownHosts)
+	if _, err := os.Stat(knownHosts); os.IsNotExist(err) {
```
ccontavalli, line 39: Same for this block, line 39 to 45. I would eliminate it entirely. If the file does not exist, the read on line 47 will fail, with file does not exist.

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,125 @@
...
+		fmt.Printf("ssh authorized hosts file %s does not exist, creating it now \n", knownHosts)
+		err = os.Mkdir(knownHosts, 0644)
+		if err != nil {
+			return fmt.Errorf("could not create file %s: %w", knownHosts, err)
+		}
+	}
+	caPublic := string(ssh.MarshalAuthorizedKey(publicKey))
+	existingKnownHostsContent, err := ioutil.ReadFile(knownHosts)
```
ccontavalli, line 47: Now: if the read fails here, it can be for many different reasons. However, the file is optional: it's ok if it doesn't exist, ssh may have never created it (fresh installation). I would change the condition to only fail if the error is != than does not exist:

```
if err != nil && !os.IsNotExist(err) {
  return fmt.Errorf("...")
}
```
and...

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return fmt.Errorf("error reading %s: %w", knownHosts, err)
+	}
+	knownHostsFile, err := os.OpenFile(knownHosts, os.O_APPEND|os.O_WRONLY, 0644)
```
ccontavalli, line 51: here is the critical part: you must be able to write the file, and create the directory if it does not exist. I would implement it like this (note os.O_CREAT, and 0600):

```
knownHostsFile, err := os.OpenFile(knownHosts, os.O_APPEND|os.O_WRONLY|os.O_CREAT, 0600)
if err != nil {
  if !os.IsNotExist(err) {
    return fmt.Errorf("could not create known_hosts file: %w", err)
  }
  if err := os.Mkdir(knownHosts, 0700); err != nil {
    return fmt.Errorf("could not create .ssh directory: %w", err)
  }
}
```

(open returns ENOENT, equivalent to NotExist, if some of the path components do not exist, https://play.golang.org/p/sesFoHhvFyz, and `man 2 open`)

lib/kcerts/ssh.go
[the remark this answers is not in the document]
ccontavalli: It's not the WriteString, it is the caPublic that contains a `\n` at the end. `MarshalAuthorizedKey` appends a \n at the end of the string, see https://play.golang.org/p/jHPir_Je6n_S

If we're relying on that, I'd add a comment here.

// caPublic terminates with a '\n', added by ...

lib/kcerts/ssh.go
aaahrens: Implemented, but earlier in the code we do guarantee that it exists and has been made

lib/kcerts/ssh.go
aaahrens: I've implemented this in a slightly different fashion that requires less nesting.

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,121 @@
...
+func StartSSHAgent() error {
+	if os.Getenv("SSH_AUTH_SOCK") != "" {
+		return nil
+	}
+	cmd := exec.Command("ssh-agent", "-s")
+	out, err := cmd.Output()
+	if err != nil {
+		return err
+	}
+	reader := bufio.NewScanner(bytes.NewReader(out))
```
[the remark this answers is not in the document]
aaahrens: The parsing is very frgaile. I had tried before to eval the output but for some reason bash kept failing on me. Didn't think of the regex, I'll try again.

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,124 @@
...
+		if os.IsNotExist(err) {
+			return fmt.Errorf("could not create known_hosts file: %w", err)
+		}
+		return err
+	}
+	existingKnownHostsContent, err := ioutil.ReadAll(knownHostsFile)
+	if err != nil {
+		return fmt.Errorf("error reading %s: %w", knownHosts, err)
+	}
+	defer knownHostsFile.Close()
```
ccontavalli, line 53: this should be moved on line 49, otherwise we will leak the file descriptor if ReadAll afterward fails.

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// FindSSHAgent Will start the ssh agent in the interactive terminal if it isn't present already as an environment variable
+// Currently only outputs the env and does not persist it across terminals
+func FindSSHAgent(store cache.Store, ttl time.Duration) (string, int, error) {
+	envSSHSock := os.Getenv("SSH_AUTH_SOCK")
+	envSSHPID := os.Getenv("SSH_AGENT_PID")
+	if envSSHSock != "" && envSSHPID != "" {
+		fmt.Println("found in env")
+		pid, err := strconv.Atoi(envSSHPID)
```
ccontavalli, line 85: am ok fixing later, but add a TODO in case: the environment variables may be totally bogus.

Eg, `SSH_AGENT_PID` and `SSH_AUTH_SOCK` may be pointing to an `ssh-agent` that is long dead or long gone.
This happens more frequently than you think, pretty much every time you ssh on a machine with agent forwarding if you use `screen` and `tmux`: your screen session will reattach you to an old shell, likely pointing to an old agent / pid that no longer exists.

My take here: a) rather than return error, just ignore the environment `ssh-agent` if anything is fishy about it (eg, pid is not an integer), b) ideally, try to use the socket before assuming that the agent is available / alive.

lib/kcerts/ssh.go
aaahrens: nice catch, done

lib/kcerts/ssh.go
aaahrens: Added in todo, leaving this chain open in case you find anything else

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,197 @@
...
+	if envSSHSock != "" && envSSHPID != "" {
+		pid, err := strconv.Atoi(envSSHPID)
+		if err != nil {
+			return "", 0, fmt.Errorf("%s is not a valid pid: %w", envSSHPID, err)
+		}
+		return envSSHSock, pid, nil
+	}
+
+	sshEnkitCache, isFresh, err := store.Get(SSHCacheKey)
+	if err != nil {
```
ccontavalli, line 93: here and below: I would refactor the error handling so that the cache is entirely optional.

Eg, if we fail to write in cache because we are out of disk space, or disk is faulty, or ... or if we fail to read a cache file (corrupted, ...) we should still be able to ssh into other boxes.

Further, we have the same problem as above: the cache may be bogus. For example: started an agent, enjoyed for a while, machine was rebooted... now cache points to an agent that no longer exists.

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,197 @@
...
+		if err != nil && !os.IsNotExist(err) {
+			return "", 0, fmt.Errorf("error opening cache: %w", err)
+		}
+		if err == nil {
+			cacheContent, err := ioutil.ReadAll(f)
+			if err != nil {
+				return "", 0, fmt.Errorf("error reading cache: %w", err)
+			}
+			var sshCache sshCache
+			if err := json.Unmarshal(cacheContent, &sshCache); err != nil {
```
ccontavalli, line 108: rather than `os.OpenFile`, `ioutil.ReadAll` followed by `json.Unmarshal`, I recommend you use: `marshal.UnmarshalFile(filepath.Join(sshEnkitCache, "ssh.json"), &sshCache)`, from `import "github.com/enfabrica/enkit/lib/config/marshal"`.

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,197 @@
...
+	pid, err := strconv.Atoi(rawPId[1])
+	if err != nil {
+		return "", 0, fmt.Errorf("error processing ssh agent pid %s: %w", resultPID, err)
+	}
+	cacheToWrite := &sshCache{
+		Sock:      rawSock[1],
+		PID:       pid,
+		TimeStamp: time.Now().Add(ttl),
+	}
+	b, err := json.Marshal(cacheToWrite)
```
ccontavalli, line 152: same comment as above, recommend using `marhsal.MarshalFile(filepath.Join(sshEnkitCache, "ssh.json"), cacheToWrite)`.

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,197 @@
...
+	envSSHPID := os.Getenv("SSH_AGENT_PID")
+	if envSSHSock != "" && envSSHPID != "" {
+		pid, err := strconv.Atoi(envSSHPID)
+		if err != nil {
+			return "", 0, fmt.Errorf("%s is not a valid pid: %w", envSSHPID, err)
+		}
+		return envSSHSock, pid, nil
+	}
+
+	sshEnkitCache, isFresh, err := store.Get(SSHCacheKey)
```
ccontavalli, line 92: you probably need a `defer store.Rollback(sshEnkitCache)` after error checking, to ensure that garbage is cleaned up if there's an error before Commit. (Rollback is guaranteed to leave the data in place)

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,197 @@
...
+
+type sshCache struct {
+	PID       int       `json:"pid"`
+	Sock      string    `json:"sock"`
+	TimeStamp time.Time `json:"time_stamp"`
+}
+
+// FindSSHAgent Will start the ssh agent in the interactive terminal if it isn't present already as an environment variable
+// Currently only outputs the env and does not persist it across terminals
+func FindSSHAgent(store cache.Store, ttl time.Duration) (string, int, error) {
```
ccontavalli, line 80: meta: I really like this function and the improvements you've made. I would however break this down in multiple functions. Something like this:

```
type Agent struct {
  Pid int
  Socket string
}

func FindSSHAgent(store cache.Store) (*Agent, error) {
  agent, err := LoadAgentFromEnv() // Parses env variables to find an agent.
  if err == nil && agent.IsValid() {
    return agent, nil
  }

  agent, cache, err = LoadAgentFromCache(store)
  defer store.Rollback(cache)
  if err == nil && agent.IsValid() {
    return agent, nil
  }

  agent, err = CreateAgent()
  if err != nil {
    return err
  }

  StoreAgentInCache(agent, cache)
  return nil
}
```

also: both re-using an existing agent or loading from cache (or storing in cache) are "optimizations", the code should not fail or return error if those things fail.

ccontavalli: By the way: if you do this, the ttl parameter is no longer needed? Basically: if you find a working agent, no matter what, you can reuse it.

ccontavalli: `IsValid` could be implemented as something like:
```
	conn, err := net.Dial("unix", agent.Socket)
	defer conn.Close()
        return err != nil
```

you could even have an `Open()` on the `Agent` object to later on use it.

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,199 @@
...
+func FindSSHAgent(store cache.Store, logger *klog.Logger) (*SSHAgent, error) {
+	// TODO(adam): sniff out dead agents and sneak pids
+	agent := FindSSHAgentFromEnv()
+	if agent != nil && agent.Valid() {
+		return agent, nil
+	}
+	agent, err := FetchSSHAgentFromCache(store)
+	if err != nil {
+		logger.Warn(err.Error())
+		if errors.Is(SSHAgentCacheInvalid, err) {
```
ccontavalli, line 111: I would remove this block, from line 111 to 115. It's none of the caller's business to cleanup after FetchSSHAgentFromCache own mess, and ... due to how the cache is structured, sure enough this file will be overwritten the first time we succeed writing in the cache, don't bother removing the entry.

(fun fact: if one was using nfs or bind mounts, the cached file may be valid in one terminal, invalid in another - I recommend being lazy here)

lib/kcerts/cache.go
```diff
@@ -0,0 +1,63 @@
...
+	SSHAgentCacheInvalid = errors.New("the cache is corrupted or invalid json")
+)
+
+func FetchSSHAgentFromCache(store cache.Store) (*SSHAgent, error) {
+	sshEnkitCache, isFresh, err := store.Get(SSHCacheKey)
+	if err != nil {
+		return nil, fmt.Errorf("error fetching cache: %w", err)
+	}
+	defer store.Rollback(sshEnkitCache)
+	agent := &SSHAgent{}
```
ccontavalli, line 27: Following my suggestion below about a "Close" method:

`&SSHAgent{Close: func() {}}`

(when the agent is read from cache, there's nothing to close)

lib/kcerts/cache.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+		return agent, err
+	}
+	return nil, SSHAgentNoCache
+}
+
+func WriteAgentToCache(store cache.Store, agent *SSHAgent) error {
+	sshEnkitCache, _, err := store.Get(SSHCacheKey)
+	if err != nil {
+		agent.failed = true
```
ccontavalli, line 40: following my advice below for a `Close` member... remove this line.
(if we failed to write the agent file, it has to be closed as normal)

lib/kcerts/cache.go
```diff
@@ -0,0 +1,63 @@
...
+		return fmt.Errorf("error fetching cache: %w", err)
+	}
+	defer store.Rollback(sshEnkitCache)
+	err = marshal.MarshalFile(filepath.Join(sshEnkitCache, SSHCacheFile), agent)
+	if err != nil {
+		agent.failed = true
```
ccontavalli, line 46: following my advice below for a Close member... remove this line.
(if we failed to write the agent file, it has to be closed as normal)

lib/kcerts/cache.go
```diff
@@ -0,0 +1,63 @@
...
+		return fmt.Errorf("error writing to cache: %w", err)
+	}
+	_, err = store.Commit(sshEnkitCache)
+	if err != nil {
+		agent.failed = true
```
ccontavalli, line 51: change this condition to be:
```
if err == nil {
  // Now that the agent is cached, it can be used in the future or in other terminals.
  // Don't close it.
  agent.Close = func() {}
}
```

lib/kcerts/cache.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	return err
+}
+
+// DeleteSSHCache deletes the SSH cache
+func DeleteSSHCache(store cache.Store) error {
```
ccontavalli, line 57: See my suggestion below: I would remove this method, as I would not invoke it from the code.

If you want to keep it (I'm ok with it, for API completeness), I strongly recommend using `Exists` instead of `Get`.
(due to how the cache was designed for atomicity, `Get` either returns an existing entry, or creates a new one in one go - it's a bit wasteful to create an entry just to remove it 3 lines below).

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,199 @@
...
+	conn, err := net.Dial("unix", a.Socket)
+	defer conn.Close()
+	return err == nil
+}
+
+// CloseIfNotCached will kill the agent if it failed to write to cache. Otherwise is a NoOp
+func (a *SSHAgent) CloseIfNotCached() error {
+	if a.failed {
+		cmd := exec.Command("kill", "-HUP", strconv.Itoa(a.PID))
+		cmd.Stdout = os.Stdout
```
ccontavalli, line 94: If you follow my advice of creating a Close attribute above, CloseIfNotFailed can be removed entirely, get rid of it.
You'll be left with: `Close` attribute (that depending on how the agent was created it'll close it in different ways) and `Kill`, to actually kill the agent.

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,199 @@
...
+// FindSSHAgentFromEnv
+func FindSSHAgentFromEnv() *SSHAgent {
+	envSSHSock := os.Getenv("SSH_AUTH_SOCK")
+	envSSHPID := os.Getenv("SSH_AGENT_PID")
+	if envSSHSock != "" && envSSHPID != "" {
+		pid, err := strconv.Atoi(envSSHPID)
+		if err != nil {
+			return nil
+		}
+		return &SSHAgent{PID: pid, Socket: envSSHSock}
```
ccontavalli, line 137: `&SSHAgent{Close: func () {}, PID: pid, Socket: envSSHSock}`

(nothing to do if we try to close this agent, does not belong to us)

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	// The second element is the raw value we want
+	rawPID := resultPID[1]
+	rawSock := resultSock[1]
+
+	pid, err := strconv.Atoi(rawPID)
+	if err != nil {
+		return nil, fmt.Errorf("error processing ssh agent pid %s: %w", resultPID, err)
+	}
+	return &SSHAgent{Socket: rawSock, PID: pid}, nil
```
ccontavalli, line 163: Change this in:
```
agent := &SSHAgent{Socket: rawSock, PID: pid}
agent.Close = func () { agent.Kill() }
return agent
```

(when this agent is closed, we need to kill the PID)

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,199 @@
...
+func (a SSHAgent) Valid() bool {
+	conn, err := net.Dial("unix", a.Socket)
+	defer conn.Close()
+	return err == nil
+}
+
+// CloseIfNotCached will kill the agent if it failed to write to cache. Otherwise is a NoOp
+func (a *SSHAgent) CloseIfNotCached() error {
+	if a.failed {
+		cmd := exec.Command("kill", "-HUP", strconv.Itoa(a.PID))
```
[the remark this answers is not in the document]
aaahrens: The sighup is usually the fastest way I've found for ssh-add to know that the agent got killed, but that's pretty anecdotal

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,199 @@
...
+	return sshDir, nil
+}
+
+// AddSSHCAToClient adds a public key to the $HOME/.ssh/known_hosts in the ssh-cert x509.1 format.
+// For each entry, it adds an additional line and does not concatenate
+func AddSSHCAToClient(publicKey ssh.PublicKey, hosts []string, sshDir string) error {
+	caPublic := string(ssh.MarshalAuthorizedKey(publicKey))
+	knownHosts := filepath.Join(sshDir, KnownHosts)
+	knownHostsFile, err := os.OpenFile(knownHosts, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0644)
+	defer knownHostsFile.Close()
```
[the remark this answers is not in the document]
aaahrens: good catch

lib/kcerts/cache.go
```diff
@@ -36,19 +38,27 @@ func FetchSSHAgentFromCache(store cache.Store) (*SSHAgent, error) {
 
 func WriteAgentToCache(store cache.Store, agent *SSHAgent) error {
 	sshEnkitCache, _, err := store.Get(SSHCacheKey)
+
+	killFunc := func() {
```
ccontavalli, line 40: no, do the opposite here: if writing succeeds, than set the "Close" to a noop function. Otherwise, leave it as is.
In the function that creates an agent, set the "Close" to "Kill" the agent.

Rationale:
As the code is written now, if one uses the API in different ways, it can lead to buggy / wrong results.

For example, let's say I change the code (or use the functions) so that `WriteAgentToCache` is always invoked, even when I've loaded the agent from the environment, or when I've read it from cache. If `WriteAgentToCache` fails, we end up killing *innocent* agents!! Some poor agents that don't belong to us.
(from documentation, this use of the functions is perfectly legitimate).

Another example: I create an agent invoking the function below. Then I never write to cache. I call Close() assuming the agent will be terminated, but it won't! As it's WriteAgentToCache that kills the agent in case of error.

Basically:
- the function that creates the SSHAgent owns the agent, and knows how to kill it.
- WriteAgentToCache, if it succeeds, should prevent the agent from being killed, as it's now part of the cache.
if it fails, it should terminate the agent as whoever owns it intended for it to be terminated.

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,197 @@
...
+func (a SSHAgent) Valid() bool {
+	conn, err := net.Dial("unix", a.Socket)
+	defer conn.Close()
+	return err == nil
+}
+
+// FindSSHAgent Will start the ssh agent in the interactive terminal if it isn't present already as an environment variable
+// Currently only outputs the env and does not persist it across terminals
+func FindSSHAgent(store cache.Store, logger *klog.Logger) (*SSHAgent, error) {
+	// TODO(adam): sniff out dead agents and sneak pids
```
ccontavalli, line 102: nit: this TODO is now addressed

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func (a SSHAgent) Valid() bool {
+	conn, err := net.Dial("unix", a.Socket)
+	defer conn.Close()
+	return err == nil
+}
+
+// FindSSHAgent Will start the ssh agent in the interactive terminal if it isn't present already as an environment variable
+// Currently only outputs the env and does not persist it across terminals
```
ccontavalli, line 100: nit: comment is out of date.

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	logger.Infof("%s", WriteAgentToCache(store, newAgent))
+	return newAgent, nil
+}
+
+// FindSSHAgentFromEnv
+func FindSSHAgentFromEnv() *SSHAgent {
+	envSSHSock := os.Getenv("SSH_AUTH_SOCK")
+	envSSHPID := os.Getenv("SSH_AGENT_PID")
+	if envSSHSock != "" && envSSHPID != "" {
```
ccontavalli, line 126: nit: if you invert the if, eg `if envSSHSock == "" || envSSHPID == "" { return nil }` you reduce the newting.

lib/kcerts/ssh.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	sshDir := filepath.Join(hDir, SSHDir)
+	if err := os.Mkdir(sshDir, 0700); err != nil && !os.IsExist(err) {
+		return "", fmt.Errorf("could not create directory %s: %w", sshDir, err)
+	}
+	return sshDir, nil
+}
+
+// AddSSHCAToClient adds a public key to the $HOME/.ssh/known_hosts in the ssh-cert x509.1 format.
+// For each entry, it adds an additional line and does not concatenate
```
ccontavalli, line 48: nit: period at end of comment.

ccontavalli: here and below.

lib/kcerts/cache.go
```diff
@@ -0,0 +1,67 @@
...
+		return nil, fmt.Errorf("error fetching cache: %w", err)
+	}
+	if sshEnkitCache == "" {
+		return nil, SSHAgentNoCache
+	}
+	agent := &SSHAgent{
+		Close: func() {},
+	}
+	if err := marshal.UnmarshalFile(filepath.Join(sshEnkitCache, SSHCacheFile), &agent); err != nil {
+		return nil, fmt.Errorf("%s: %w", SSHAgentCacheInvalid, err)
```
ccontavalli, line 33: ping on the comment I had added here. The SSHAgentCacheInvalid here... is pretty much just a string? I would recommend removing SSHAgentCacheInvalid. 

Basically: the function returns "an agent, and an error", an error is returned if the cache is invalid (eg, line 24 is a form of invalid cache), unless it's SSHAgentNoCache.

ccontavalli (approved):
All nits remaining. Please address them (recommend looking at function documentation and comments in general, add period, grammar), and merge. No need to wait for further review from me.

PR #54, opened 2021-03-29
Add CLI parameters to auth server regarding CA certificates

aaahrens:
Blocking: PR #52 needed to generate the final stages of the certificate (aka private key and public signed cert) 

This is hand tested with local configs to the best of my ability.

### DIFF
--- a/astore/server/auth/auth.go
+++ b/astore/server/auth/auth.go
@@ -7,6 +7,7 @@ import (
 	"github.com/enfabrica/enkit/astore/common"
 	"github.com/enfabrica/enkit/astore/rpc/auth"
 	"golang.org/x/crypto/nacl/box"
+	"golang.org/x/crypto/ssh"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"io"
@@ -24,6 +25,10 @@ type Server struct {
 
 	authURL string
 	limit   time.Duration
+
+	caSigner              ssh.Signer
+	principals            []string
+	marshalledCAPublicKey []byte
 }
 
 type Jar struct {
@@ -96,8 +101,12 @@ func (s *Server) Token(ctx context.Context, req *auth.TokenRequest) (*auth.Token
 		}
 
 		return &auth.TokenResponse{
-			Nonce: nonce[:],
-			Token: box.Seal(nil, []byte(token), &nonce, (*[32]byte)(clientPub), (*[32]byte)(s.serverPriv)),
+			Nonce:       nonce[:],
+			Token:       box.Seal(nil, []byte(token), &nonce, (*[32]byte)(clientPub), (*[32]byte)(s.serverPriv)),
+			Capublickey: s.marshalledCAPublicKey,
+			// Always trust the CA for now since the DNS gets resolved behind tunnel and therefore the client doesnt know
+			// which to trust
+			Cahosts:     []string{"*"},
 		}, nil
 	}
 
--- a/astore/server/auth/factory.go
+++ b/astore/server/auth/factory.go
@@ -1,7 +1,10 @@
 package auth
 
 import (
+	"errors"
 	"fmt"
+	"golang.org/x/crypto/ssh"
+	"io/ioutil"
 	"math/rand"
 	"strings"
 	"time"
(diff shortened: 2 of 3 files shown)


astore/server/auth/factory.go
```diff
@@ -24,6 +29,8 @@ func DefaultFlags() *Flags {
 
 func (f *Flags) Register(set kflags.FlagSet, prefix string) *Flags {
 	set.DurationVar(&f.TimeLimit, prefix+"time-limit", f.TimeLimit, "How long to wait at most for the user to complete authentication, before freeing resources")
+	set.StringVar(&f.Principals, prefix+"principals", f.Principals, "Allowed principals and the ability to auth,comma separated string")
```
ccontavalli, line 32: nit: change "auth,comma" to "auth, comma" (space)

astore/server/auth/factory.go
```diff
@@ -24,6 +29,8 @@ func DefaultFlags() *Flags {
...
+	set.StringVar(&f.PathToCA, prefix+"ca", f.PathToCA, "Path to the certificate authority private file")
```
ccontavalli, line 33: For files, use `set.ByteFileVar`, defined in `lib/kflags/flags.go`. By identifying the variable as the path to a file (and not just as a string), the file will be loadable / embeddable correctly, and flag parsing can take care of checking if the file is loadable.

astore/server/auth/factory.go
```diff
@@ -61,6 +73,33 @@ func WithTimeLimit(limit time.Duration) Modifier {
 	}
 }
 
+func WithCA(path string) Modifier {
+	return func(server *Server) error {
+		fileContent, err := ioutil.ReadFile(path)
```
ccontavalli, line 78: see comment above, this would become unnecessary

astore/server/auth/factory.go
ccontavalli: Also, what is the meaning of principals in this context? Can you provide an example? (in the help string above, not in comments on the PR)

astore/server/auth/auth.go
```diff
@@ -96,8 +101,12 @@ func (s *Server) Token(ctx context.Context, req *auth.TokenRequest) (*auth.Token
 		}
 
 		return &auth.TokenResponse{
-			Nonce: nonce[:],
-			Token: box.Seal(nil, []byte(token), &nonce, (*[32]byte)(clientPub), (*[32]byte)(s.serverPriv)),
+			Nonce:       nonce[:],
+			Token:       box.Seal(nil, []byte(token), &nonce, (*[32]byte)(clientPub), (*[32]byte)(s.serverPriv)),
+			Capublickey: s.marshalledCAPublicKey,
+			// Always trust the CA for now since the DNS gets resolved behind tunnel and therefore the client doesnt know
```
ccontavalli, line 107: nit: s/doesnt/doesn't/

astore/server/auth/auth.go
```diff
@@ -96,8 +101,12 @@ func (s *Server) Token(ctx context.Context, req *auth.TokenRequest) (*auth.Token
...
+			// which to trust
+			Cahosts:     []string{"*"},
```
ccontavalli, line 109: This should also return the private key for the user, correct? I assume that's coming in the next PRs?

aaahrens: Yup, need #52 to generate the Keys from the ca.Signer

astore/server/auth/factory.go
```diff
@@ -29,8 +28,8 @@ func DefaultFlags() *Flags {
 
 func (f *Flags) Register(set kflags.FlagSet, prefix string) *Flags {
 	set.DurationVar(&f.TimeLimit, prefix+"time-limit", f.TimeLimit, "How long to wait at most for the user to complete authentication, before freeing resources")
-	set.StringVar(&f.Principals, prefix+"principals", f.Principals, "Allowed principals and the ability to auth,comma separated string")
-	set.StringVar(&f.PathToCA, prefix+"ca", f.PathToCA, "Path to the certificate authority private file")
+	set.StringVar(&f.Principals, prefix+"principals", f.Principals, "Authorized ssh users which the ability to auth, in a comma separated string e.g. \"john,root,admin,smith\"")
```
ccontavalli, line 31: uhm .. now I'm confused though. When I authenticate with the auth server, I need a certificate for myself, eg, `carlo`, since I'm `carlo@enfabrica.net`. In the future, I may get `extra roles` (like root or similar) based on the group I belong to.

Are those additional principals, on top of my user name, I presume? If yes, just update the comment accordingly.
If not, let's chat :)

ccontavalli (approved):
If we're on the same page re comment below, update the help string, and merge, don't wait for further approval from me.

ccontavalli (approved):

PR #55, opened 2021-03-30
Work in Progress: License Manager

ahungrynacho:
Design Doc
https://docs.google.com/document/d/1TNqbBprpcNU9tTHVCFzRwaQoHlGFdjkw221C5p9UsAw/edit?usp=sharing

- Will eventually implement a queuing mechanism for handling more license requests than licenses available in FIFO order
- Currently, when all licenses are being used, a pool of clients will send a license request every 60 seconds to the license manager server. FIFO ordering is not guaranteed.
- Have a non-blocking keepalive heartbeat periodically ping the license manager server every 60 seconds in the background while the main xilinx/cadence command runs in the foreground. If the keepalive heartbeat disconnects, automatically release the license and stop the xilinx/cadence build/simulation from running.
- The clients will always initiate a request while the license manager server will always be listening for requests. Because of this, even when a license becomes available after being released, the server will have to wait for a polling request from a client to allocate the license. To keep track of multiple polling license requests from clients waiting for an available license while also maintaining order, an array can be used where a unique job ID corresponding to each polling client are the elements while the indices of the array are the order in which the licenses should be allocate per client.

Sample logs from server side after running `bazel run :cmac_rx --test_arg="-waves"`
```
2021/04/16 22:13:25 Client n4djd2gwfyts2uj5a2hbpuwvyfps45cz from bbhuynh@127.0.0.1:8080 waiting for 1 xilinx vivado 
2021/04/16 22:13:25 xilinx job queue: n4djd2gwfyts2uj5a2hbpuwvyfps45cz 
2021/04/16 22:13:25 License acquired by n4djd2gwfyts2uj5a2hbpuwvyfps45cz 
2021/04/16 22:13:25 xilinx vivado licenses in use: 1/3 
2021/04/16 22:13:25 Client n4djd2gwfyts2uj5a2hbpuwvyfps45cz removed from the xilinx queue 
2021/04/16 22:13:25 Client oi2ut7aqtkogxxjj8nwwfxnfik5onw3x from bbhuynh@127.0.0.1:8080 waiting for 1 xilinx vivado 
2021/04/16 22:13:25 xilinx job queue: oi2ut7aqtkogxxjj8nwwfxnfik5onw3x 
2021/04/16 22:13:25 License acquired by oi2ut7aqtkogxxjj8nwwfxnfik5onw3x 
2021/04/16 22:13:25 xilinx vivado licenses in use: 2/3 
2021/04/16 22:13:25 Client k37wiruipc8r3npzdqeqgd5aj4c75m86 from bbhuynh@127.0.0.1:8080 waiting for 1 xilinx vivado 
2021/04/16 22:13:25 xilinx job queue: oi2ut7aqtkogxxjj8nwwfxnfik5onw3x, k37wiruipc8r3npzdqeqgd5aj4c75m86 
2021/04/16 22:13:25 Client oi2ut7aqtkogxxjj8nwwfxnfik5onw3x removed from the xilinx queue 
2021/04/16 22:13:25 Client gyaenjjmyt2duwmzzhrm6pqqe2bsz227 from bbhuynh@127.0.0.1:8080 waiting for 1 xilinx vivado 
2021/04/16 22:13:25 xilinx job queue: k37wiruipc8r3npzdqeqgd5aj4c75m86, gyaenjjmyt2duwmzzhrm6pqqe2bsz227 
2021/04/16 22:13:30 License acquired by k37wiruipc8r3npzdqeqgd5aj4c75m86 
2021/04/16 22:13:30 xilinx vivado licenses in use: 3/3 
2021/04/16 22:13:30 Client k37wiruipc8r3npzdqeqgd5aj4c75m86 removed from the xilinx queue 
2021/04/16 22:15:02 License released by oi2ut7aqtkogxxjj8nwwfxnfik5onw3x 
2021/04/16 22:15:02 Client oi2ut7aqtkogxxjj8nwwfxnfik5onw3x disconnected 
2021/04/16 22:15:02 xilinx vivado licenses in use: 2/3 
2021/04/16 22:15:05 License acquired by gyaenjjmyt2duwmzzhrm6pqqe2bsz227 
2021/04/16 22:15:05 xilinx vivado licenses in use: 3/3 
2021/04/16 22:15:05 Client gyaenjjmyt2duwmzzhrm6pqqe2bsz227 removed from the xilinx queue 
2021/04/16 22:15:14 License released by n4djd2gwfyts2uj5a2hbpuwvyfps45cz 
2021/04/16 22:15:14 Client n4djd2gwfyts2uj5a2hbpuwvyfps45cz disconnected 
2021/04/16 22:15:14 xilinx vivado licenses in use: 2/3 
2021/04/16 22:15:36 License released by k37wiruipc8r3npzdqeqgd5aj4c75m86 
2021/04/16 22:15:36 Client k37wiruipc8r3npzdqeqgd5aj4c75m86 disconnected 
2021/04/16 22:15:36 xilinx vivado licenses in use: 1/3 
2021/04/16 22:19:20 License released by gyaenjjmyt2duwmzzhrm6pqqe2bsz227 
2021/04/16 22:19:20 Client gyaenjjmyt2duwmzzhrm6pqqe2bsz227 disconnected 
2021/04/16 22:19:20 xilinx vivado licenses in use: 0/3 
2021/04/16 22:21:47 Client sxoqa6i3v2unaej686bqemfda6ytmmnh from bbhuynh@127.0.0.1:8080 waiting for 1 cadence xcelium 
2021/04/16 22:21:47 cadence job queue: sxoqa6i3v2unaej686bqemfda6ytmmnh 
2021/04/16 22:21:47 License acquired by sxoqa6i3v2unaej686bqemfda6ytmmnh 
2021/04/16 22:21:47 cadence xcelium licenses in use: 1/3 
2021/04/16 22:21:47 Client sxoqa6i3v2unaej686bqemfda6ytmmnh removed from the cadence queue 
2021/04/16 22:22:37 License released by sxoqa6i3v2unaej686bqemfda6ytmmnh 
2021/04/16 22:22:37 Client sxoqa6i3v2unaej686bqemfda6ytmmnh disconnected 
2021/04/16 22:22:37 cadence xcelium licenses in use: 0/3 
```


manager/common/license.go
```diff
@@ -0,0 +1,62 @@
...
+type Server struct {
+}
+
+func (s *Server) Acquire(ctx context.Context, request *rpc_license.AcquireRequest) (*rpc_license.AcquireResponse, error) {
+	log.Printf("Received request for %d %s feature %s \n", request.Quantity, request.Vendor, request.Feature)
+	if licenseCounter.counter < licenseCounter.total {
+		log.Printf("Licenses in use: %d/%d \n", licenseCounter.increment(int(request.Quantity)), licenseCounter.total)
+		return &rpc_license.AcquireResponse{Available: true, Waiting: false, Missing: false}, nil
+	} else {
+		return &rpc_license.AcquireResponse{Available: false, Waiting: true, Missing: false}, nil
```
ahungrynacho, line 44: The job ID should be returned to the client inside the response body so that if the client needs to poll for a license, the connection can be uniquely identified from other clients.
```
jobId := 0
...
jobId += 1
jobId = jobId % MAX_INT // prevent integer overflow
```

manager/client/acquire/main.go
```diff
@@ -0,0 +1,50 @@
...
+	"os"
+	"time"
+	"log"
+	"context"
+	"google.golang.org/grpc"
+	"strconv"
+	rpc_license "github.com/enfabrica/enkit/manager/rpc"
+)
+
+func main() {
```
ccontavalli, line 13: As mentioned over chat, I'm very worried about having an `acquire` command and a separate `release` command. Whenever the build is killed, or fails, or there's a coding error in a rule, .. if we fail to call `release`, we leak licenses, I fear this won't work reliably.

My ideal would be to have a command `with-license` that wraps another command, something like:

```
./with-license vendor feature feature -- vivado -mode tcl ...
```

.. eg, runs the command after "--" passing all the remaining flags to it, after grabbing the licenses. When the command exits, the licenses are automatically freed. No need to call a separate release command.

If you implement the suggested changes in the `.proto` file, we end up with:
- having to call `Acquire` until it succeeds, giving us a `ticket`.
- having to call `KeepaliveRequest` (aka `RenewLease` if you accept the suggestion) every 30 seconds.
- having to call `Release` with the ticket (at the end of the process).

To implement something like I described, the code could look something like:

```
func main() {
  // ... flag parsing ...
  vendor := ...
  features := []string{...}
  command := []string{... leftover argv parsing ...}
  
  // ... grpc initialization ...

  // Try to grab a license.
  // Use import "github.com/lib/retry"
  var response rpc_license.AcquireResponse
  if err := retry.New(retry.WithWait(60), retry.WithAttempts(30)).Run(func () error {
    response, err = client.Acquire(context.Background(), &rpc_license.AcquireRequest{Vendor: vendor, Feature: features, Quantity: int32(quantity)})
    return err  
  }); err != nil {
    log.Fatalf("could not acquire license - exiting")
  }

  // If we got here, it means we successfully got a license. Retrying succeeded.
  // Now: 0) make sure we release the license as quickly as possible once done,
  // 1) keep the license alive, and 2) run the command.

  // defer causes the function to invoke at return, no matter what.
  defer Release(client, response.Ticket)
  // go starts a separate thread.
  go Refresh(client, response.Ticket, 60 * time.Second)  

  // import "os/exec"
  cmd := exec.Command(command[0], command[1:]...)
  cmd.Stdin = os.Stdin
  cmd.Stdout = os.Stdout

  // Finally run the command.
  if err := cmd.Run(); err != nil {
    if exitError, ok := err.(*exec.ExitError); ok {
        os.Exit(exitError.ExitCode())
    }
    // Running the command failed with an unknown exit code. log something.
    os.Exit(255)
  }
  return // this will exit 0.
}
```

Release and Refresh could be implemented as something like:
```
func Refresh(client *rpc_license.LicenseClient, ticket string, delay time.Duration) {
  for {
    if err := client.RenewLease(context.Background(), &rpc_license.KeepaliveRequest{ticket: ticket}); err != nil {
      time.Sleep(5 * time.Second)  // we failed to renew the ticket, retry faster to avoid expiry!
      // TODO: stop the loop if it's a fatal error. Eg, maybe the server restarted, and does not know about the ticket!
    } else {
      time.Sleep(dealy / 2) // renew the ticket well before it expires.
    }
  }
}

func Release(client, ticket) {
  client.Release(...) // no real need to handle errors here. Worst case, the lease will expire and the license freed.
}
```

manager/client/acquire/main.go
```diff
@@ -0,0 +1,50 @@
...
+	timeout := 1800 * time.Second
+	interval := 60 * time.Second
```
ccontavalli, line 15: For argv parsing, the rest of the codebase here is using cobra with a bit of magic plumbing. For this command specificially, I recommend you do the same as the astore main, [here](https://github.com/enfabrica/enkit/blob/master/astore/server/main.go#L252). If you want to, we can look at it together, but if you follow the advice above, you can use the same pattern with `DefaultFlags` to register the `retry` library flags as well, so the attempts etc are configured from the command line.

manager/common/license.go
ccontavalli: See recommendations for changes to the interface below. I would actually recommend you return a random string, and just keep it in a hash table.

To generate the string, you can use something like [GenerateUID here](https://github.com/enfabrica/enkit/blob/master/astore/server/astore/astore.go#L55) (happy to move this in more of a library function).

A simple `map[string]*Client` for example, where client is something like:
```
type Client struct {
  IP string // extracted from grpc connection
  Owner string // ...
  Grabbed time.Time // when the license was grabbed
  Refreshed time.Time // when the license was last refreshed
  Interval time.Duration // max wait time for a refresh
}
```

... to "return licenses once the timer expires", and "refresh the timers", you can use something like (the pinger thread here)[https://github.com/enfabrica/enkit/blob/master/lib/icmp/pinger.go#L450], which runs a separate thread, which just waits for the next timer to expire, and invokes a callback. This should also be moved in a library, like `lib/goroutine`.

manager/client/acquire/main.go
ccontavalli: (I actually recommend you do those changes last, my first recommendation would be to focus on the interface below, and mechanics above)

manager/client/main.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	if waiting >= timeout {
+		log.Fatalf("Failed to acquire %d %s feature %s license after %d seconds \n", quantity, vendor, feature, waiting)
+	}
+	return hash, nil
+}
+
+func main() {
+	var quantity int32 = 1
+	host, port := os.Args[1], os.Args[2]
```
ahungrynacho, line 92: I will integrate the cobra command line parsing library in a subsequent commit, just FYI.

manager/common/license.go
```diff
@@ -1,62 +1,175 @@
...
+	astoreServer "github.com/enfabrica/enkit/astore/server/astore"
 )
 
 type LicenseCounter struct {
-    mutex sync.Mutex
-    counter int
-    total int
+    counterMutex sync.Mutex
+    counter map[string]int // current number of licenses used per vendor
+	clientMutex sync.Mutex
```
ahungrynacho, line 17: Will fix up formatting with `gofmt -w license.go`

manager/common/license.go
```diff
@@ -0,0 +1,175 @@
...
+			status = err
+			break
+		}
+	}
+	inUse := licenseCounter.decrement(vendor, 1, hash)
+	log.Printf("%s %s licenses in use: %d/%d \n", vendor, feature, inUse, licenseCounter.total[vendor])
+	return status
+}
+
+func (s *Server) Polling(stream rpc_license.License_PollingServer) error {
```
ccontavalli, line 134: It took me some time to understand how the code works, there are parts that are not clear to me.

Correct me if wrong:
- client opens a Polling stream. On this stream, the client sends PollingRequests every few seconds. If the requested license is available, then acquired: true is returned, and the stream is closed. Otherwise, acquired: false is returned.
- client opens a KeepAlive stream. When this stream is closed, the license is released.

The keepalive messages exchanged in this stream... serve no purpose? as there is no timeout? it's only based on the Read() returning an error?

Also, the hash is ... not necessary as is, right? Eg, given that a stream must be kept open, the state may as well be tied to the stream, without a hash? in the queue you could put a client object or the stream itself, right?

In my previous comment, I was recommending either using streams, *or* passing around a hash.

My  take:
- we only need a hash, a cookie/token exchanged between client and server, if we either don't use streaming, eg, simple rpc messages, stateless, or if we want to support re-connecting if the stream fails/is closed.
- don't fully understand why we need two streams here.

We could get away with just one, right? Which would also make the code resilient to a client that crashes/never sends a keepalive? And we could remove the hash entirely if we use streaming?

(well, unless we add a bit of code to re-connect without losing a license, which could be valuable)

ahungrynacho: Ok, I can consolidate my code to just have a single bidirectional stream. I do not want to implement any reconnect logic because sometimes disconnections are intentional like when a user on the client side does `CTRL+C` when the bazel test/build is in progress but the user wants to abort.

manager/common/license.go
```diff
@@ -0,0 +1,175 @@
...
+			if err != nil {
+				log.Fatalf("Failed to generate hash for %s@%s", recv.User, ip)
+			}
+			length := licenseCounter.enqueue(hash, &Client{IP: ip, User: username, Start: start})
+			log.Printf("Client %s from %s@%s received at %s waiting at position %d \n", hash, username, ip, start, length)
+		} else {
+			hash = recv.Hash
+		}
+		log.Printf("Received request for %d %s feature %s from client %s \n", recv.Quantity, recv.Vendor, recv.Feature, hash)
+		if licenseCounter.counter[recv.Vendor] < licenseCounter.total[recv.Vendor] && licenseCounter.queue[0] == hash {
```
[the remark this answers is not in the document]
ahungrynacho: oopsie

manager/common/license.go
```diff
@@ -1,40 +1,41 @@
...
 
 type LicenseCounter struct {
-    counterMutex sync.Mutex
-    counter map[string]int // current number of licenses used per vendor
-	clientMutex sync.Mutex
-	clients map[string]*Client // client connection metadata
-    total map[string]int // total licenses available per vendor
-	queue []string
+	counterMutex sync.Mutex
+	counter      map[string]int // current number of licenses used per vendor
```
ccontavalli, line 17: I recommend having counter and total part of the same map. A single map cannot go out of sync :)
eg
```
license map[string]*License
```

```
type License struct {
  Total int
  Used int
}
```

manager/BUILD.bazel
```diff
@@ -0,0 +1,29 @@
+load("@io_bazel_rules_docker//docker/util:run.bzl", "container_run_and_commit")
+load("@io_bazel_rules_docker//container:container.bzl", "container_image", "container_push")
+
+container_run_and_commit(
+    name = "server-image",
+    image = "@ubuntu-20.04//image",
```
ccontavalli, line 6: why ubuntu-20.04 for the license manager? No need to pull in a distro at all for a pure go binary docker container.
Use distroless instead: https://github.com/GoogleContainerTools/distroless/blob/main/examples/go/BUILD

It's both easier than what you are doing, creates much smaller images, with less security issues, which don't need to be updated (eg, if we base on ubuntu-20.04, eventually we'll need to udpate).

manager/client/main.go
```diff
@@ -0,0 +1,102 @@
...
+	"os/exec"
+	"os/user"
+	"strings"
+	"time"
+)
+
+func run(timeout time.Duration, cmd string, args ...string) {
+	channel := make(chan bool)
+	defer close(channel)
+	go func(status chan bool, cmd string, args ...string) {
```
ccontavalli, line 20: ugh. No need for this either! `os/exec` supports timeouts natively, look at `CommandContext`, I found an example here:
https://medium.com/@vCabbage/go-timeout-commands-with-os-exec-commandcontext-ba0c861ed738

(further down the page, look for CommandContext)

manager/common/license.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func (c *LicenseCounter) increment(vendor string, num int, hash string) int {
+	c.counterMutex.Lock()
+	defer c.counterMutex.Unlock()
+	if c.licenses[vendor].Used < c.licenses[vendor].Total {
+		log.Printf("License acquired by %s \n", hash)
+		c.licenses[vendor].Used += num
+	} else {
+		log.Fatalf("Cannot acquire more licenses than available \n")
```
ccontavalli, line 30: in increment, decrement, etc:
1) invert the if, remove the else.
2) Don't go Fatal, just log.Errorf

Example:

```
if c.licenses[vendor].Used >= c.licenses[vendor].Total {
  log.Errorf(... cannot acquire ...)
  return;
}

c.licenses[vendor].Used += num
```

if you crash and burn with the Fatal, all builds in progress will fail/start failing. Further, docker will restart the server, and make debugging harder, making everything flaky.

Might as well "log the error" and keep going. Will be easier to debug, and will make the system more stable.

manager/common/license.go
```diff
@@ -0,0 +1,192 @@
...
+			if err != nil {
+				log.Printf("Failed to generate hash for %s@%s", recv.User, ip)
+				return err
+			}
+			client := Client{IP: ip, User: username, Start: start, Quantity: recv.Quantity, Vendor: vendor, Feature: recv.Feature}
+			licenseCounter.enqueue(hash, &client)
+		} else {
+			hash = recv.Hash
+		}
+		_, total, available := licenseCounter.peekLicense(vendor)
```
ccontavalli, line 156: please validate the vendor before using it, or make `peekLicense`, `increment`, `dequeue` tolerant to invalid vendors.

If I fat finger the vendor in the cli, I can crash the server: the lookup in the hash table will return nil, null pointer dereference, crash.

manager/common/license.go
```diff
@@ -0,0 +1,192 @@
...
+		if available && licenseCounter.peekQueue(vendor, hash) {
+			log.Printf("%s %s licenses in use: %d/%d \n", vendor, recv.Feature,
+				licenseCounter.increment(vendor, int(recv.Quantity), hash),
+				total)
+			licenseCounter.dequeue(hash, vendor)
+			err := stream.Send(&rpc_license.PollingResponse{Acquired: true, Hash: hash})
+			if err != nil {
+				log.Printf("Failed to send message back to client %s: %s \n", hash, err)
+			} else {
+				break
```
ccontavalli, line 166: should the break happen error or not? don't think the stream is valid after a failed send? would need to check the docs. Generally on send or receive error the stream is closed / client dropped, there's clearly a communication problem.

manager/client/main.go
ahungrynacho: Huh, I didn't know that the exec library natively supported timeouts. That is interesting.

manager/client/main.go
```diff
@@ -0,0 +1,93 @@
...
+    if err != nil {
+        log.Fatalf("Job \"%s %s\" failed with error %s", cmd, strings.Join(args, " "), err)
+    } else {
+		log.Printf("Job completed successfully: %s %s \n", cmd, strings.Join(args, " "))
+    }
+}
+
+func polling(client rpc_license.LicenseClient, username string, quantity int32, vendor string, feature string,
+	cmd string, args ...string) {
+	timeout := 1800 * time.Second
```
aaahrens, line 34: It would be a really good idea to use a context.WithTimeout here instead of setting it as a manual timeout. 
Reason: -> you can pass the context down a chain of async commands and they're all reading from the same timeout.
                  also, it looks clearer and can be used with grpc-web

manager/rpc/license.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
+
+package license;
+
+message PollingRequest{
+	string vendor = 1; // Xilinx or Cadence
```
aaahrens, line 6: If you already know these values, they should be enums. Means you don't have spend time string validating.

manager/common/license.go
```diff
@@ -0,0 +1,191 @@
...
+	Vendor   string
+	Feature  string
+}
+
+type License struct {
+	Total int
+	Used  int
+}
+
+var clients = map[string]*Client{}
```
aaahrens, line 109: you can use 
```
var (
      A = 3
      b = 4
)
```
way of declaring multiple variables

aaahrens (changes requested):
So I won't poke holes too much in the grpc polling sections because the CLI isn't finished and neither is testing. As has already been pointed out, there are plenty of ways to nil pointer the project.

The only issue I see architecturally is that for all tests, the implementation right now would be all e2e tests. Usually that's valid, but in this case licenses take money and time. 

The way I would mitigate this and make it so that 95% of the code is tested as unit-ish test: 

under /manager declare an interface like 
```
type LicenseManager interface {
    Type() string
    Allocate() error
    // You get the idea here, and the Server would accept N number of Managers per type of 
    // licenses it manages.
}
```
The counter you wrote can be its own struct, its pretty general.
The manager could be something like 
```
type Server struct {
     Xilinx LicenseManager
     Cadence LicenseManager
}
```

and then in lib/xilinx 

```
type XilinxManager struct {
          *queue.Counter // general queue library you wrote
          // rest of interface implementation of LicenseManager
}

```

So during testing, you can punt in testing structs of LicenseManagers that can timeout or deterministically return items, and run ideally 1 or 2 behaviour driven tests to make sure the concrete implementations work, minimizing real license allocation.

This is just a suggestion, you could make the queue object be the interface. (or do one of the 12 billion other possibilities)

Just generally accept interfaces as parameters and return structs.

manager/client/main.go
```diff
@@ -0,0 +1,103 @@
...
+	"log"
+	"os"
+	"os/exec"
+	"os/user"
+	"strings"
+	"time"
+)
+
+func run(timeout time.Duration, cmd string, args ...string) {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
ccontavalli, line 20: go fmt? indentation seems wonky

manager/client/main.go
```diff
@@ -0,0 +1,103 @@
...
+    defer cancel()
+    job := exec.CommandContext(ctx, cmd, args...)
+    err := job.Run()
+    if ctx.Err() == context.DeadlineExceeded {
+		log.Fatalf("Job failed to complete after %d seconds: %s %s \n", timeout, cmd, strings.Join(args, " "))
```
ccontavalli, line 25: genuinely asking: will the log.Fatal actually kill the runaway child process? eg, I would not expect the **go** process exiting to kill its children. I would have expected a `job.Process.Kill()` to be necessary to ensure it is actually killed before the log.Fatal.

manager/client/main.go
```diff
@@ -0,0 +1,103 @@
...
+    }
+    if err != nil {
+        log.Fatalf("Job \"%s %s\" failed with error %s", cmd, strings.Join(args, " "), err)
+    } else {
```
ccontavalli, line 29: no need for else here.

manager/common/license.go
```diff
@@ -0,0 +1,204 @@
...
+			licenseCounter.del(hash, vendor)
+			return nil
+		}
+		if err != nil {
+			licenseCounter.del(hash, vendor)
+			return err
+		}
+		if !licenseCounter.validLicense(recv.Vendor, recv.Feature) {
+			errMsg := fmt.Sprintf("\"%s %s\" is an unsupported vendor feature combination", recv.Vendor, recv.Feature)
+			return grpcStatus.Errorf(grpcCodes.InvalidArgument, errMsg)
```
ccontavalli, line 148: no need for `Sprintf` here, `Errorf` calls sprintf internally:
```
return grpcStatus.Errorf(grpcCodes.InvalidArgument, "'%s %s' is an unsuppoaterd vendor / feature combination", ...);
```

manager/client/main.go
```diff
@@ -0,0 +1,103 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	rpc_license "github.com/enfabrica/enkit/manager/rpc"
+	grpcCodes "google.golang.org/grpc/codes"
```
ccontavalli, line 7: why rename them? (usually renaming is only done if there is a naming conflicts in the import, but don't see any here)

manager/common/license.go
```diff
@@ -0,0 +1,204 @@
...
+	s.rng = rand.New(rand.NewSource(rand.Int63()))
+	for {
+		recv, err := stream.Recv()
+		// remove client hash from queue when connection is broken
+		if err == io.EOF {
+			licenseCounter.del(hash, vendor)
+			return nil
+		}
+		if err != nil {
+			licenseCounter.del(hash, vendor)
```
ccontavalli, line 143: this seems a bit suspicious. `vendor` can change value from one loop to the next, but only the last value of vendor is deleted from the counter handler?

ahungrynacho: I assumed that I would need to remove the pending request from the queue in case the user terminates the job request before a license is allocated to it. But I should introduce a check inside the `licenseCounter.del()` method to check whether the vendor exists. The vendor will not change in between iterations because the server is receiving the same repeated request from the client when the client polls the server with a license request.

manager/client/main.go
ahungrynacho: I chose to rename it to be consistent with the `grpcStatus` package I renamed since I categorized all these packages as part of the grpc library. Not required, but just my preference.

manager/client/main.go
ahungrynacho: `ctx.Done()` is called internally after the timeout error
https://stackoverflow.com/questions/61001275/is-it-safe-to-leave-out-ctx-done-when-using-the-go-context-package

manager/common/license.go
ccontavalli: Genuinely asking, is it ok to call `licenseCounter.del` with empty hash and empty vendor? It can happen from multiple threads, if we have an error on the first Recv.

Also, "the vendor will not change in between iterations" is entirely up to what the client is sending? Having server code that can end up in inconsistent state depending on the client behavior worries me. 

Impact here is low, so I'm ok with this PR, it's a first step. But writing the code so that it is resilient to this kind of problems is not significantly harder, I claim the code here would even be simpler if eg, only considered the vendor on first message, eliminated the hash, ...

ccontavalli (commented):
So, I'm still a bit uncomfortable about this PR, mostly due to the complexity of some of the code, and assumptions.

However, at this point I would iterate on a merged version, one step at a time, rather than on this PR, already has been pending too long.

Of all comments here, please address the one about the `for loop never looping`, and feel free to merge in.

manager/rpc/license.proto
```diff
@@ -0,0 +1,20 @@
...
+	string feature = 2; // Vivado, Xcelium, etc
+	int32 quantity = 3; // Number of licenses
+	string user = 4; // Username
+	string hash = 5; // Job ID of previous request by the same client
+}
+
+message PollingResponse{
+	bool acquired = 1; // status of the acquired license
```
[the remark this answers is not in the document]
ahungrynacho: I chose not to use the grpc status codes for this because returning a status code from the Polling() function under `common/license.go` will terminate the grpc connection. I want to maintain the grpc connection while notifying the client from the server that a license is not available yet. In the event that the client requests for an invalid vendor or feature, I terminate the grpc connection from the server side with a `InvalidArgument` code sent to the client:
```
        if !licenseCounter.validLicense(recv.Vendor, recv.Feature) {
            return grpcStatus.Errorf(grpcCodes.InvalidArgument, "\"%s %s\" is an unsupported vendor feature combination", recv.Vendor, recv.Feature)
        }
```

manager/common/license.go
```diff
@@ -0,0 +1,207 @@
...
+type Server struct {
+	rng *rand.Rand
+}
+
+func (s *Server) Polling(stream rpc_license.License_PollingServer) error {
+	hash := ""
+	vendor := ""
+	var status error
+	s.rng = rand.New(rand.NewSource(rand.Int63()))
+	for {
```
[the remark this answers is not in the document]
ahungrynacho: 1. If the connection with the client terminates prematurely before a license is allocated, remove the client from the queue but do not decrement the license counter.
2. If the connection with the client terminates prematurely AFTER the license has been allocated or if the client finishes its job gracefully (implying that the client has also been removed from the queue already), only decrement the license counter.

Because disconnection or failure on the client side can only happen in these 2 states, I cannot call `defer` because I need to handle these two different scenarios differently.

manager/common/license.go
```diff
@@ -0,0 +1,207 @@
...
+			err := stream.Send(&rpc_license.PollingResponse{Acquired: false, Hash: hash})
+			if err != nil {
+				log.Printf("Failed to send message back to client %s: %s \n", hash, err)
+				licenseCounter.del(hash, vendor)
+				return err
+			}
+		}
+	}
+	// license acquired
+	for {
```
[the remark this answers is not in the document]
ahungrynacho: After removing the `for {}` loop, the `stream.Recv()` blocking call is necessary to wait for the client to close the grpc connection when it finishes the job gracefully or terminates prematurely (ctrl+c) before releasing the license by calling `licenseCounter.decrement()`

PR #57, opened 2021-03-30
adding in basic node polling with tags and names, including e2e testing

aaahrens:
This implements polling with names and tags on machinist. 

This does not include the shape of the cli or any additional package main type functions, it is purely the controllers. 

I also deleted vestigial folders from the initial draft, which are going to handled by other parts of the application, for example assets will be handled by webpack/react itself.


machinist/mserver/cmd/main.go
```diff
@@ -0,0 +1,27 @@
...
+		Short: "controller is a server in charge of controlling workers",
+	}
+
+	oauthFlags := oauth.DefaultRedirectorFlags()
+	oauthFlags.Register(&kcobra.FlagSet{command.Flags()}, "")
+
+	command.RunE = func(cmd *cobra.Command, args []string) error {
+		return Start(oauthFlags)
+	}
+
```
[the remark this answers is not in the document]
ccontavalli: (as a side note, we should probably have a common `Run` for CLI utilities, and a common `Run` for server binaries, right now it's a bit confusing)

machinist/mserver/cmd/main.go
```diff
@@ -0,0 +1,27 @@
+package main
+
+import (
+	"github.com/enfabrica/enkit/lib/kflags/kcobra"
+	"github.com/enfabrica/enkit/lib/oauth"
+	"github.com/spf13/cobra"
+)
+
+func Start(oauthFlags *oauth.RedirectorFlags) error {
+	return nil
```
ccontavalli, line 10: shouldn't this start an mserver object :)? I'd recommend using the 'lib/server' Run function, to set up the mux and actually run it.

machinist/mserver/controller.go
```diff
@@ -1,16 +1,15 @@
...
 	"sync"
 )
 
-type Worker struct {
-}
-
 type Controller struct {
-	workers sync.Map
+	workers        sync.Map
+	ConnectedNodes map[string][]string
```
ccontavalli, line 11: A couple notes here:
- `workers` was meant to be your `ConnectedNodes` and `sync.RWMutex` together? a `sync.Map` is pretty much an uglier map, with built in locking. You may want to revert the ConnectedNodes and addition of the lock?
- `ConnectedNodes` ... would probably want it to be private? so lowercase. Especially given that it requires a lock to be accessed, so someone with access to Controller should only read it through accessors.
- Rather than `map[string][]string`, I recommend `map[string]*Node` (eg, a struct that records the tags, when the last ping was received, the ip address of the node, ... etc, basis for the web page showing the list of nodes, for example)

machinist/mserver/mserver.go
```diff
@@ -0,0 +1,47 @@
...
+	Controller    *Controller
+	sync.Mutex
+	*machinist.SharedFlags
+}
+
+func (s *server) Flags() *machinist.SharedFlags {
+	return s.SharedFlags
+}
+
+func (s *server) Start() error {
```
ccontavalli, line 37: same comment as another PR. Start() generally starts something and returns, Run() or Serve() block. This blocks, so I would rename it.

machinist/mserver/flags.go
```diff
@@ -0,0 +1,23 @@
...
+		for _, mod := range mods {
+			if err := mod(s); err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+}
+
+func WithController(controller *Controller) Modifier {
```
ccontavalli, line 18: why does a `machinist server` require knowing a controller?

aaahrens: It doesn't. It populates the server with an empty one at the start. It gives the option of passing one in however to introspect state e.g. for testing or boiling up scope.

machinist/mnode/mnode.go
```diff
@@ -0,0 +1,69 @@
...
+	initialRequest := &machinist_rpc.PollRequest{
+		Req: &machinist_rpc.PollRequest_Register{
+			Register: &machinist_rpc.ClientRegister{
+				Name: n.Name,
+				Tag:  n.Tags,
+			},
+		},
+	}
+	if err := pollStream.Send(initialRequest); err != nil {
+		return errors.Wrap(err, "unable to send initial request")
```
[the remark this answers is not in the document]
aaahrens: It uses the same interface as standard errors, and I wrote this section a while ago and didn't realize I was using it, whoops <.<

machinist/mserver/flags.go
aaahrens: It doesn't. It populates the server with an empty one at the start. It gives the option of passing one in however to introspect state e.g. for testing or boiling up scope.

machinist/mserver/cmd/main.go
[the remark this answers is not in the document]
aaahrens: I noted in the body of the PR that this doesn't include the shape or the cli or any plugins, this mainly has to do with logic only just to keep the size of the pr reasonable

machinist/mserver/controller.go
aaahrens: I don't want to use sync.Map until generics hit in 2 months or so just because I don't want to be typecasting everything.

You're right about the private access with a getter and the storing a struct, this was just for prototyping sake. Changed

machinist/mserver/mserver.go
aaahrens: Changed to run

machinist/mserver/cmd/main.go
aaahrens: See above about CLI shape

machinist/mserver/controller.go
aaahrens: *Changed the connected node accessors and naming. Also, concurrent reading on maps is supported.

machinist/mserver/controller.go
```diff
@@ -7,11 +7,22 @@ import (
 )
 
 type Controller struct {
-	workers        sync.Map
-	ConnectedNodes map[string][]string
+	connectedNodes map[string]*Node
 	sync.RWMutex
 }
 
+func (en *Controller) Nodes() []*Node {
```
ccontavalli, line 14: nit, but important: I generally assume that public methods of a function are ~always safe to call, or prominently indicate they aren't.

Nodes() here can only be called by someone holding the lock. Can you mark the method private AND add a comment indicating that the caller is expected to hold the Controller Lock?

machinist/mserver/controller.go
```diff
@@ -7,11 +7,22 @@ import (
 )
 
 type Controller struct {
-	workers        sync.Map
-	ConnectedNodes map[string][]string
+	connectedNodes map[string]*Node
```
ccontavalli, line 10: Looks better, thanks. Now... related to the comment below, about public methods:
- by embedding sync.RWMutex directly, anyone can call `Controller.Lock()` and `Controller.Unlock`. Is that desired? Generally public interfaces of an object should not require the **caller** to know how to perform locking on the **callee** (makes changing and optimizing locking later harder, makes all the callsites potentially causing deadlocks/mishandling)
- further, as the code grows, I generally end up with some parts of an object *not* protected by lock (eg, read only, set at initialization) and other parts read write, or under different locks.

My recommendation would be to change to:
```
type Controller struct {
  connectedNodes map[string]*Node // All access protected by c_lock below.
  c_lock sync.RWMutex
}
```

(eg, name locks, so they are not accessible from outside, add comments so it's clear what's locked by what)

machinist/mserver/controller.go
ccontavalli: Or make it public, with proper locking.

machinist/mserver/controller.go
```diff
@@ -34,7 +45,11 @@ func (en *Controller) HandlePing(stream machinist.Controller_PollServer, ping *m
 
 func (en *Controller) HandleRegister(stream machinist.Controller_PollServer, ping *machinist.ClientRegister) error {
 	en.Lock()
-	en.ConnectedNodes[ping.Name] = ping.Tag
+	n := &Node{
+		Name: ping.Name,
+		Tags: ping.Tag,
+	}
+	en.connectedNodes[ping.Name] = n
```
ccontavalli, line 52: nit:
```
n := &Node{
...
}
en.Lock()
en.connectedNodes[ping.Name] = n
en.Unlock()
```

better even would be to have an "Add" method on the Controller object? or have a locked connectedNodes object.
That way you could use defer (which is panic safe)

machinist/machinist_e2e_test.go
```diff
@@ -52,10 +52,11 @@ func TestJoinServerAndPoll(t *testing.T) {
...
 	}()
 	time.Sleep(2 * time.Second)
-	assert.Equal(t, 1, len(mController.ConnectedNodes))
+	assert.Equal(t, 1, len(mController.Nodes()))
+	assert.NotNil(t, mController.Node("test-01"))
 
-	for k, v := range mController.ConnectedNodes {
-		assert.Equal(t, []string{"big", "heavy"}, v)
-		assert.Equal(t, "test-01", k)
+	for _, v := range mController.Nodes() {
```
ccontavalli, line 58: Nodes() cannot be accessed without locking? See next comments.

machinist/mserver/controller.go
```diff
@@ -7,11 +7,22 @@ import (
...
+func (en *Controller) Nodes() []*Node {
+	var nodes []*Node
+	for _, v := range en.connectedNodes {
+		nodes = append(nodes, v)
+	}
+	return nodes
+}
+
+func (en *Controller) Node(name string) *Node {
+	return en.connectedNodes[name]
```
ccontavalli, line 23: same here, locking, name, ...

machinist/mserver/controller.go
ccontavalli: looking how controlledNodes is used so far, I actually recommend having the locking directly in Nodes(), Node(), and adding a "Set()" method (also doing locking). `c_lock` would then only be used by those few methods.

(I would also consider creating a dedicated NodeSet data structure, for example, so the public Controller methods don't get muddied with the private NodeSet methods, and you can *guarantee* and *enforce* using the locks on NodeSet, via its public methods - any Controller method can otherwise access c_lock directly) (entirely optional now, as Controller, well, there's not much there anyway).

machinist/mserver/factory.go
```diff
@@ -2,7 +2,7 @@ package mserver
 
 func NewController() (*Controller, error) {
 	en := &Controller{
-		ConnectedNodes: make(map[string][]string),
+		connectedNodes: make(map[string]*Node),
```
ccontavalli, line 5: nit: make is necessary only in a handful of cases, generally recommend using plain initialization when make is not necessary:
```
connectedNodes: map[string]*Node{},
```

machinist/mnode/mnode.go
```diff
@@ -0,0 +1,69 @@
...
+	// Dial func will override any existing options to connect
+	DialFunc func() (*grpc.ClientConn, error)
+}
+
+func (n *Node) Flags() *machinist.SharedFlags {
+	return n.SharedFlags
+}
+
+func (n *Node) Init() error {
+	if n.DialFunc != nil {
```
ccontavalli, line 26: nit: would invert the if to reduce nesting. Given this still needs to be implemented, I'm ok merging as is, and addressing later.

ccontavalli (approved):

PR #58, opened 2021-03-30
Adds DNS server to knetwork 

aaahrens:
Open to making a new package, but logically it makes sense that it would live in knetwork.

Tested, and currently only has functionality to add records.


lib/knetwork/BUILD.bazel
```diff
@@ -1,8 +1,22 @@
-load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
```
ccontavalli, line 1: meta: could you move this under a subdirectory of knetwork? like knetwork dns?

rationale: knetwork right now is pulled in by the tunnel CLI interface. No reason to bring in a build time dependency on a DNS server library?

lib/knetwork/dns.go
```diff
@@ -0,0 +1,191 @@
...
+	"sync"
+)
+
+var (
+	DNSEntryNotExistError = errors.New("the following entry did not exist")
+	DNSEntryExistError    = errors.New("the following entry did exist")
+)
+
+func NewDNS(mods ...DNSModifier) (*DnsServer, error) {
+	defaultLogger, err := klog.New("default", klog.FromFlags(*klog.DefaultFlags()))
```
ccontavalli, line 19: most libraries in the repo don't create a logger. They default to `&NilLogger`, and expect the caller to provide one.
I recommend to follow the same pattern. To all effects, the caller will have a logger and pass it in.

lib/knetwork/dns.go
```diff
@@ -0,0 +1,191 @@
...
+		s.host = ip
+		return nil
+	}
+}
+
+type DnsServer struct {
+	dnsServer *dns.Server
+	sync.RWMutex
+	routeMap map[string][]string
+	Port     int
```
ccontavalli, line 76: nit: I tend to group public attributes together, and private attributes together. Generally, if you read the code of a library you're probably trying to find out what the public attributes are, the private ones are mostly noise.

lib/knetwork/dns.go
```diff
@@ -0,0 +1,191 @@
...
+	Logger   *klog.Logger
+	domains  []string
+	Listener net.Listener
+	host     string
+}
+
+func (s *DnsServer) Start() error {
```
ccontavalli, line 83: `Start()` will block forever, by the fact that it calls ListenAndServe() below? If that's the case, I'd recommend calling it `Run()` instead.

Mild convention: `Start()` starts something and returns. `Run()` loops forever or blocks until completion.

lib/knetwork/dns.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	return s.dnsServer.ListenAndServe()
+}
+
+func (s *DnsServer) Stop() error {
+	return s.dnsServer.Shutdown()
+}
+
+func (s *DnsServer) AddEntry(name string, ips []string) error {
+	if s.routeMap[dns.Fqdn(name)] != nil {
```
ccontavalli, line 105: you also need a lock to read the map, don't you? otherwise you may end up reading the map while it is being written into, which would crash?

lib/knetwork/dns.go
```diff
@@ -0,0 +1,191 @@
...
+		return fmt.Errorf("%w: %s", DNSEntryExistError, name)
+	}
+	s.Lock()
+	s.routeMap[dns.Fqdn(name)] = ips
+	s.Unlock()
+	return nil
+}
+
+// ReplaceEntry will hard replace an entry. Consider it a force AddEntry
+func (s *DnsServer) ReplaceEntry(name string, ips []string) error {
```
ccontavalli, line 115: s/Replace/Set/ maybe? Adds it if not there, overwrite if it already exists.

lib/knetwork/dns.go
```diff
@@ -0,0 +1,191 @@
...
+	s.Lock()
+	s.routeMap[name] = ips
+	s.Unlock()
+	return nil
+}
+
+// AppendToEntry will add ips to an existing entry. Any collisions are automatically handled. It will not automatically
+// add the entry if it does not exist
+func (s *DnsServer) AppendToEntry(name string, ips []string) error {
+	if s.routeMap[dns.Fqdn(name)] == nil {
```
ccontavalli, line 125: same locking comment as above.

lib/knetwork/dns.go
```diff
@@ -0,0 +1,191 @@
...
+	s.Lock()
+	s.routeMap[fqdnName] = appendIfNotPresent(s.routeMap[fqdnName], ips)
+	s.Unlock()
+	return nil
+}
+
+// RemoveIPFromEntry will remove ips from an entry if the entry and name exists. I will not error if the ip to delete
+// is not found in th records
+func (s *DnsServer) RemoveIPFromEntry(name string, ips []string) error {
+	if s.routeMap[name] == nil {
```
ccontavalli, line 138: not implemented? and locking.

lib/knetwork/dns.go
```diff
@@ -0,0 +1,191 @@
...
+		return fmt.Errorf("%w: %s", DNSEntryNotExistError, name)
+	}
+	return nil
+}
+
+func (s *DnsServer) RemoveEntry(name string) error {
+	return nil
```
ccontavalli, line 145: ?

lib/knetwork/dns.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+}
+
+func (s *DnsServer) ParseDNS(m *dns.Msg) {
+	for _, q := range m.Question {
+		switch q.Qtype {
+		case dns.TypeA:
+			ips := s.routeMap[q.Name]
+			for _, ip := range ips {
+				rr, err := dns.NewRR(fmt.Sprintf("%s A %s", q.Name, ip))
```
ccontavalli, line 168: don't parse the string every time you need to generate a response, parse the string when adding it to the map. Keep the rr in the map directly.

Rationale:
1) If an invalid record is added, you want to find out at addition time, not when the response is generated.
2) Parsing and sprintf is expensive, no reason to spend the cycles here.

Personally, I would change `routeMap map[string][]string` to instead be `routeMap map[string]*Record`, and then you can define record as:

```
type Record struct {
  IPs []dns.RR
  IPv6s []dns.RR
  TXTs []string
}
```
(IPs is the list of things returned for an 'A' Query, IPv6s for 'AAAA', TXTs for TXT query. TXT may come particularly handy to add annotations about the IPs returned - eg, description of machine, or records for scripts to use)

lib/knetwork/dns.go
```diff
@@ -0,0 +1,191 @@
...
+					m.Answer = append(m.Answer, rr)
+				}
+			}
+		}
+	}
+}
+
+// generic func
+func appendIfNotPresent(s1, s2 []string) (inter []string) {
+	hash := make(map[string]bool)
```
ccontavalli, line 179: fyi, if simpler, especially now that I'm asking to change the RR: up to a ... don't know, <= 100 values (which is certainly the case here), it is probably cheaper to have nested for loops. This code is not performance critical, so don't really care, but having RRs and TXTs as proposed above is going to make it harder to use a hash table.

ccontavalli (commented):
Very good code! Only one real comment here.

lib/knetwork/BUILD.bazel
aaahrens: Moved to package kdns

lib/knetwork/dns.go
aaahrens: Defaults to NilLogger now

lib/knetwork/kdns/records.go
```diff
@@ -0,0 +1,203 @@
...
+		rc.aRecordOps <- recordOperation{Type: RecordDelete, Keywords: keywords}
+	case dns.TypeTXT:
+		rc.txtRecordOps <- recordOperation{Type: RecordDelete, Keywords: keywords}
+	default:
+		rc.ErrorChan <- fmt.Errorf("%w: %v", UnSupportedTypeErr, recordType)
+	}
+}
+
+// NewRecordController create a new controller for a specified origin, or basename;
+func NewRecordController() *RecordController {
```
ccontavalli, line 177: ok, let me see if I understand the design correctly:
1) for each name, eg, `worker-01.corp.enfabrica.net` there is a RecordController
2) a record controller holds two go coroutines, one for A records, one for TXT records.
3) each goroutine has a private list of records
4) the goroutine keeps waiting for: a) operations to edit the set of records, b) requests to read the records.
5) when records are edited (a), they are changed in the same thread. When they are read, they are sent over the channel.

What worries me here is 1) the leaking of goroutines, as records are added and deleted, 2) the # of goroutines, eg, if we have 100 names, there's going to be ... 3x goroutines? 3) an operation like a simple enumeration of the hosts (for a web interface? or to add prometheus monitoring?) will require querying all names? waking them all. 4) the moment we address those issues (eg, reclaiming goroutines, copying the array passed around - or having locks around those arrays) I think this code will be much more complex (cleanup is hard to do without deadlocks and weird race conditions).

I'm impressed by this design, truly - but in terms of complexity/debugging/concurrency issues, your original design using a simple map with locking was significantly simpler.

If I had to guess, at most will have 10s of queries per second reading, at most we'll have a couple changes per day, we will have however O(100s) of different names.  I would personally favor reducing complexity and lines of code over performance here

At this level, an `RWMutex` should be ok? A `sync.Map` would introduce some annoying type casting, but would probably lead to the same level of scalability than you have here? Getting rid of a global mutex, while still being very simple. We could scale up on goroutines in the DNS listener (or is it using a pool already?).

lib/knetwork/kdns/dns.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func (s DnsServer) HandleControllers() {
+	controllerMap := map[string]*RecordController{}
+	for {
+		select {
+		case o := <-s.requestControllerChan:
+			if controllerMap[o.Origin] == nil {
+				c := NewRecordController()
+				go logControllerError(s.Logger, c.ErrorChan)
```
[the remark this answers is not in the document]
aaahrens: Error chan was for logging yeah, but passing the logger does make sense, implemented

lib/knetwork/kdns/dns.go
```diff
@@ -0,0 +1,125 @@
...
+func logControllerError(log logger.Logger, errChan chan RecordControllerErr) {
+	for {
+		e := <-errChan
+		log.Errorf("%s", e)
+	}
+}
+
+// AddEntry will append a
+func (s *DnsServer) AddEntry(name string, rr dns.RR) {
+	cleanedName := dns.Fqdn(name)
```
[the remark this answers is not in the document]
aaahrens: The raw query (like what gets thrown in by the dns server) is by FQDN unfortunately. Everything in the internal DNS server and in NewRR seems to be using FQDN.

lib/knetwork/kdns/records.go
aaahrens: Yes you understand the design correctly. 
And going over the concern in order:
1) They don't leak, see the added tests. All the transient channels are buffered, and the unbuffered ones get cleaned up, 

2) Correct. I'm ok with this.

3. Because of situations like this is why I rewrote it for nonblocking records. The scraping done by external parties is arbitrary, ttl is user configurable and the records are beings used in other systems. Any sort of deadlock would screw up multiple systems. With a mutex and a runaway script it would be easy to congest the dns.

4. I'm not sure if you caught on, but the pure read time without any operations is actually constant. Even a sync.Map (which I looked into because it does offer nicer sharding, which is nice and does scale better for higher read scenarios) begins significantly slowing down at 14000 records roughly. Which might seem like a lot until it isn't. 

4 cont. I won't argue it's simpler, but I will argue that it's tested. And goroutine ownership isn't exactly out of left field. If this gets merged I might take the first half of your comment and make it the readme even. I generally don't like speculating as to how much people will use X service because smart people tend to be very creative (and not cache friendly).

lib/knetwork/kdns/records.go
```diff
@@ -0,0 +1,203 @@
...
+		return append(DeleteIfContains(operation.Keywords, src), operation.Records...)
+	case RecordDelete:
+		return DeleteIfContains(operation.Keywords, src)
+	case RecordWriteForce:
+		return operation.Records
+	}
+	return src
+}
+
+func (rc *RecordController) watchARecords() {
```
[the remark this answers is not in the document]
aaahrens: Added a close defer and a Stop method just in case, but unbuffered channels are actually part of garbage collection. 

`You needn't close every channel when you've finished with it. It's only necessary to close a channel when it is important to tell the receiving goroutines that all data have been sent. A channel that the garbage collector determines to be unreachable will have its resources reclaimed whether or not it is closed. (Don't confuse this with the close operation for open files. It is important to call the Close method on every file when you've finished with it.)`

lib/knetwork/kdns/records.go
```diff
@@ -0,0 +1,203 @@
...
+	var aRecords []dns.RR
+	for {
+		select {
+		case operation := <-rc.aRecordOps:
+			aRecords = handleOperation(aRecords, operation)
+		case rc.aRecords <- aRecords:
```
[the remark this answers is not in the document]
aaahrens: handleOperation is entirely immutable, it never modifies the array in place. I believe the dns server does as well. The only proof I have that it's threadsafe is the race condition is on during testing.

lib/knetwork/kdns/dns.go
```diff
@@ -0,0 +1,125 @@
...
+	}{Return: returnChan, Origin: dns.Fqdn(origin)}
+	return <-returnChan
+}
+
+func (s DnsServer) HandleControllers() {
+	controllerMap := map[string]*RecordController{}
+	for {
+		select {
+		case o := <-s.requestControllerChan:
+			if controllerMap[o.Origin] == nil {
```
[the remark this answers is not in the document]
aaahrens: This is a good catch, and correct. implemented.

lib/knetwork/kdns/records.go
ccontavalli: Did you measure the performance? In 4, even if the code looks constant, it won't be: a simple enumeration will cause goroutines to be woken up and put to sleep constantly, while reading and writing into a channel. Waking and putting goroutines to sleep based on channel reading/writing ... I expect it to be significantly more costly than `sync.Map` mutexes - consider that coroutine management and channel still require some form of locking, atomics, and/or preempting.

If I run this microbenchmark here: https://gist.github.com/craigmj/5770480, similar to your code? It uses a channel to perform map operations with a serving goroutine, 10 goroutines on 8 CPU:
```
carlo@penguin:~/test$ ./test 
In parallel on 8 CPUs with 10 goroutines
GoMap:       1.356325302s
GoMap1Chan:  1.283012764s
SyncMap:     410.67364ms
```

But even this does not really matter? If we ever get to 14k records, we'll have a full team of developers, and we'll likely have other problems than this (mutliple server instances? monitoring? sharding? ...).

IMHO, what can bite us more than anything else are bugs like leaks of goroutine and concurrency, which there were a couple of in this PR. Also, ability to iterate over the code, adding functionalities as we learn what we need exactly. My recommendation would to focus on a) simplicity, b) defensiveness/reliability - more than performance.

Given the investment in time work, I'm ok moving forward with this once we address the pending comments, but I would argue that as we add features - or find other bugs - we should strive to simplify this code.

Some references:
https://medium.com/@deckarep/the-new-kid-in-town-gos-sync-map-de24a6bf7c2c - shows that a simple map with RWMutex can take about ~20 Million lookups per second, does not change with core numbers (scales very poorly). sync.Map can take about 4 Million lookups per second, but scales linearly with cores (eg, at about 4-5 cores, beats a normal map).

lib/knetwork/kdns/dns.go
```diff
@@ -117,9 +141,13 @@ func (s *DnsServer) HandleIncoming(writer dns.ResponseWriter, incoming *dns.Msg)
 // ParseDNS will only handle dns requests from domains that it is specified to handle. I will modify the *dns.Msg inplace
 func (s *DnsServer) ParseDNS(m *dns.Msg) {
 	for _, q := range m.Question {
-		rrs := s.ControllerForName(q.Name).FetchRecords(q.Qtype)
-		for _, txt := range rrs {
-			m.Answer = append(m.Answer, txt)
+		c := s.ControllerForName(q.Name)
+		if c != nil {
```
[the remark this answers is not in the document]
aaahrens: I would rather throw out REFUSED instead, which forces the client to try another DNS server. returning NXDOMAIN would make the client not retry,  which is bad considering we want the client to fall back to google. 


https://github.com/miekg/dns/blob/c99ea652e37a78dd34941023c1d4cc32c0654ecd/serve_mux.go#L105 it does this by default if the domains not handled are inputted in.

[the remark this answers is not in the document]
ccontavalli: Either way, not answering here is bad, we need to reply. 

I have conflicting feelings about NXDOMAIN and REFUSED. Let's chat more in person.

lib/knetwork/kdns/records.go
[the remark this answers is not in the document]
ccontavalli: don't see this solved?

lib/knetwork/kdns/records.go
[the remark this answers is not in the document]
ccontavalli: don't see this solved either?

ccontavalli (approved):
This PR still makes me very uncomfortable, I still think there is a lot of complexity for little benefit. I see us fighting bugs and paying a high cost to add simple features moving forward for no strong benefit.

That said, I think we need to move forward. We can iterate on the merged code.

Before merging, please fix:
1) The two simple comments.
2) Either implement or put TODOs in place for REFUSED, NXDOMAIN, deletion of controller, and using close() to shutdown the servers.

PR #60, opened 2021-03-31
[WIP][Manual] Machnist client-side auth

aaahrens:


astore/client/auth/auth.go
```diff
@@ -103,5 +113,42 @@ func (c *Client) Login(username, domain string, o LoginOptions) (string, error)
...
+	agent, err := kcerts.FindSSHAgent(o.Store, o.L)
+	if err != nil {
+		return "", err
+	}
+	defer agent.Close()
+	file, err := ioutil.TempFile("/tmp", "en")
+	if err != nil {
+		return "", err
+	}
+	err = marshal.MarshalFile(file.Name(), tres.Key)
```
ccontavalli, line 137: No! Don't write the private key in /tmp! Also, no reason to fork off to `ssh-add`, just use the golang ssh-agent client API directly (add a Connect() method to the SSHAgent object you created in previous PR?). Code could look like this ([a full example here](https://pkg.go.dev/golang.org/x/crypto/ssh/agent#NewClient)):
```
	conn, err := net.Dial("unix", agent.Socket)
        ... err ...
        // from import "golang.org/x/crypto/ssh/agent"
	agentClient := agent.NewClient(conn)
        agentClient.Add(...)
```

ccontavalli: also... we need something a bit smarter to add the key: we need to garbage collect old keys? Make sure to have a Lifetime specified).

If we don't garbage collect keys, this whole thing will stop working in like 5 days (reason: last time I checked how this worked, the ssh client presented one key at a time to the server, but the server refused to challenge more than X keys - X == 5? to avoid DoS and similar).

astore/client/auth/auth.go
```diff
@@ -103,5 +113,42 @@ func (c *Client) Login(username, domain string, o LoginOptions) (string, error)
 		return "", fmt.Errorf("could not decrypt returned token")
 	}
 
+	publicKey, _, _, _, err := ssh.ParseAuthorizedKey(tres.Capublickey)
```
ccontavalli, line 116: Meta, this is quite a bit of code, I'd personally move it into a function.

ccontavalli, line 116: Side note, unrelated to your PR: loop from 88 to 103 should probably use the Retry library (did not have it when this code was written).

astore/client/auth/auth.go
```diff
@@ -38,6 +46,8 @@ type LoginOptions struct {
 	// the token and the next. This is meant to prevent busy loops DoSsing
 	// the server, while allowing fast retries in the normal case.
 	MinWait time.Duration
+	Store   cache.Store
+	L       *klog.Logger
```
ccontavalli, line 50: call it Log, L can be easily mistaken for eg, a Lock.

ccontavalli, line 50: also, pass in a `logger.Logger`, eg, `Log logger.Logger`, a generic logger.

lib/client/commands/login.go
```diff
@@ -84,16 +85,23 @@ func (l *Login) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Right now recreating the logger because ssh requires a klog.Logger and it only gets called once and immediately is
+	// coerced into an interface type, not sure if I should just typecast here or deal with propagation
```
ccontavalli, line 89: Basically, libraries are written so they work with any logging library that implements the `logger.Logger` interface.
Just change the type of the struct to be `logger.Logger`, pass in the logger defined by base.

(otherwise, the flags passed on the command line to control logging won't work - if the libraries you just recently merged in take a concrete klog, they should be changed to take an abstract logger instead - may have missed it during review)

astore/client/auth/auth.go
aaahrens: I'll implement this in another PR, but yeah good point especially since the browser lib is pretty fickle

lib/kcerts/ssh.go
```diff
@@ -96,9 +97,27 @@ func (a SSHAgent) Valid() bool {
 	return err == nil
```
ccontavalli, line 97: just notice the line above, probably need to add an `if err != nil` before `defer` and `close`? I suspect the code as is will work, but not sure the fact that close works on a nil conn is specified behavior.

ccontavalli (commented):
The only real comment is about the `Store` initialization, the rest are nice to have / suggestions.

astore/client/auth/auth.go
```diff
@@ -38,6 +43,7 @@ type LoginOptions struct {
 	// the token and the next. This is meant to prevent busy loops DoSsing
 	// the server, while allowing fast retries in the normal case.
 	MinWait time.Duration
+	Store   cache.Store
```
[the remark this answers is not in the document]
aaahrens: Initialized here:
https://github.com/enfabrica/enkit/pull/60/files#diff-cb178c6f41f85a63f8e69bb257973b12e49f34cc72c6b95e0a4bfdecd0c978baR90

lib/kcerts/ssh.go
```diff
@@ -96,9 +97,27 @@ func (a SSHAgent) Valid() bool {
...
 }
 
+func (a SSHAgent) AddCertificates(privateKey *rsa.PrivateKey, publicKey ssh.PublicKey, ttl uint32) error {
+	conn, err := net.Dial("unix", a.Socket)
+	if err != nil {
+		return err
+	}
+	defer conn.Close()
+	cert, ok := publicKey.(*ssh.Certificate)
```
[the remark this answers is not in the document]
aaahrens: You can pass in a ssh.Certificate in the publicKey section of this function, considering ssh.PubKey is an interface. The intention behind accepting the interface is because it's standard to accept interfaces and means the caller doesn't have to typecast considering there's no way to marshal a certificate (always have to typecast a pub key).

lib/kcerts/ssh.go
```diff
@@ -96,9 +97,27 @@ func (a SSHAgent) Valid() bool {
...
+	if !ok {
+		return fmt.Errorf("public key is not a valid ssh certificate")
+	}
+	agentClient := agent.NewClient(conn)
```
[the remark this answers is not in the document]
aaahrens: I would rather not do this right now because it would affect other parts of Close() functionality, including  when it gets reassigned. I like it how it is right now where it cannot leak open connections/all commands op -> run -> finish.

lib/kcerts/ssh.go
aaahrens: It does work as intended because the defer would swallow the nil exception, but it should catch the error, at least to be idiotmatic. Added in a err handle.

astore/client/auth/auth.go
```diff
@@ -98,10 +104,39 @@ func (c *Client) Login(username, domain string, o LoginOptions) (string, error)
...
 	}
 
-	decrypted, ok := box.Open(nil, []byte(tres.Token), nonce.ToByte(), servPub.ToByte(), priv)
+	decrypted, ok := box.Open(nil, tres.Token, nonce.ToByte(), servPub.ToByte(), priv)
 	if !ok {
 		return "", fmt.Errorf("could not decrypt returned token")
 	}
 
-	return string(decrypted), nil
+	caPublicKey, _, _, _, err := ssh.ParseAuthorizedKey(tres.Capublickey)
```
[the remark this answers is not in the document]
aaahrens: technically this doesn't help for testability since during testing we're just going to execute the cobra command, but implemented for readability at least

ccontavalli: very important, just forgot about this: if I cut an enkit version today, login will break, as the server will not return a `Capbulickey`. Please make this code conditional, something like:

```
if len(tres.Capublickey) == 0 || len(tres.Cert) == 0 {
  ...
  return immediately;
}
```

This will also allow to do rollbacks cleanly in case we need to.

astore/client/auth/auth.go
```diff
@@ -98,10 +105,48 @@ func (c *Client) Login(username, domain string, o LoginOptions) (string, error)
 		return "", fmt.Errorf("server returned invalid nonce, please try again - %s", err)
 	}
 
-	decrypted, ok := box.Open(nil, []byte(tres.Token), nonce.ToByte(), servPub.ToByte(), priv)
+	decrypted, ok := box.Open(nil, tres.Token, nonce.ToByte(), servPub.ToByte(), priv)
 	if !ok {
 		return "", fmt.Errorf("could not decrypt returned token")
 	}
+	err = loadSSHKey(tres, o.Store, o.Logger)
```
ccontavalli, line 112: nit:
```
if err := loadSshKey(...); err != nil {
}
```
is more idiomatic.

lib/kcerts/ssh.go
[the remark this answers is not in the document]
ccontavalli: ack, no strong feelings here.

lib/kcerts/ssh.go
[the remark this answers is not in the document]
ccontavalli: sgtm

astore/client/auth/auth.go
aaahrens: yeah I just caught that XD

ccontavalli (approved):

ccontavalli (approved):

astore/server/auth/factory.go
```diff
@@ -79,6 +78,9 @@ func WithTimeLimit(limit time.Duration) Modifier {
 
 func WithCA(fileContent []byte) Modifier {
 	return func(server *Server) error {
+		if len(fileContent) == 0 {
+			return nil
```
ccontavalli, line 82: does server has a `Log`/`Logger` member? If yes, add a:
```
server.Warnf("CA file not found - will operate without certificates")
```

astore/server/auth/factory.go
```diff
@@ -92,9 +94,6 @@ func WithCA(fileContent []byte) Modifier {
 func WithPrincipals(raw string) Modifier {
 	return func(server *Server) error {
 		splitString := strings.Split(raw, ",")
```
ccontavalli, line 96: Use the username as principal? Not in this PR?

aaahrens: It's in #61

astore/server/auth/factory.go
aaahrens: It did not, now it does. Although passing it in isn't supported right now with the way the auth server is instantiated, it'll include the normal logger at a later date. Probably during spring cleaning.

ccontavalli (approved):

PR #65, opened 2021-04-07
add rpc for host certificates

aaahrens:


astore/rpc/auth.proto
```diff
@@ -26,9 +26,20 @@ message TokenResponse {
   repeated string cahosts = 6; // List of hosts the CA should be trusted for
 }
 
+message HostCertificateRequest {
+  repeated string principals = 1; // List of principals you want to be able too ssh into this server
```
ccontavalli, line 30: as per chat, this is a list of hosts? Change comment.

astore/rpc/auth.proto
```diff
@@ -26,9 +26,20 @@ message TokenResponse {
...
+  bytes hostcert = 2; // The signed host certificate
+}
+
 service Auth {
   // Use to retrieve the url to visit to create an authentication token.
   rpc Authenticate(AuthenticateRequest) returns (AuthenticateResponse) {}
   // Use to retrieve an authentication token.
   rpc Token(TokenRequest) returns (TokenResponse) {}
+
+  rpc HostCertificate(HostCertificateRequest) returns (HostCertificateResponse) {}
```
ccontavalli, line 44: comment? "// Use to retrieve an SSH certificate for a host."

astore/rpc/auth.proto
```diff
@@ -26,9 +26,20 @@ message TokenResponse {
   repeated string cahosts = 6; // List of hosts the CA should be trusted for
 }
 
+message HostCertificateRequest {
+  repeated string principals = 1; // List of principals you want to be able too ssh into this server
+}
+
+message HostCertificateResponse {
+  bytes capublickey = 1; // The CA public key
```
ccontavalli, line 34: nit: end comments with a period, here and below.

astore/rpc/auth.proto
```diff
@@ -26,9 +26,20 @@ message TokenResponse {
   repeated string cahosts = 6; // List of hosts the CA should be trusted for
 }
 
+message HostCertificateRequest {
+  repeated string hosts = 1; // List of dns names you want this server to be authorized to be. Does not have to be valid TLDs.
```
ccontavalli, line 30: As per comment in the chat: I think this should sign a certificate provided directly by the client, with the private key never sent over the wire.

I would s/HostCertificate/SignCertificate/g

ccontavalli, line 30: nit: the list of hosts should also include the IP addresses of the machine? If yes, could you update the comment? Should it include loopback as well?

aaahrens: It does not include the ip addresses of the host machines unfortunately. If you want to connect via a raw ip address but still have a valid host certificate it'll give you the "want to trust this host yes/no" prompt ssh always has.

To mitigate it I'm not sure at what stage ssh does the dns lookup for certificates, it might be a non issue and reads the machinist dns. The dns names dont even need to be FQDNs.

aaahrens: see comment in chat.

ccontavalli (approved):

PR #66, opened 2021-04-08
Change ssh command to include environments from new ssh agent

aaahrens:


proxy/ptunnel/commands/ssh.go
```diff
@@ -83,7 +84,11 @@ func (r *SSH) Run(cmd *cobra.Command, args []string) error {
 	ecmd.Stdin = os.Stdin
 	ecmd.Stdout = os.Stdout
 	ecmd.Stderr = os.Stderr
-
+	agent, err := kcerts.FindSSHAgent(r.BaseFlags.Local, r.Log)
```
ccontavalli, line 87: can you make this feature flag protected? Have a flag like "use internal agent" defaulting to true that can be disabled? Reason: I may have my own agent running with a different set of keys loaded, I need to be able to avoid using the one provided by enkit.

As a side note, we should provide a way for users to load extra keys/run extra commands in the `enkit` owned agent, eg, `enkit agent -- rsync -avz ...` or `enkit agent -- ssh-add -l`, useful for debugging as well.
(all that `enkit agent` does is set up the agent in the environment, and run the specified command)

proxy/ptunnel/commands/ssh.go
```diff
@@ -83,7 +84,11 @@ func (r *SSH) Run(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return err
+	}
+	ecmd = agent.AddToCmd(ecmd)
```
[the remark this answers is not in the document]
aaahrens: seems good, implemented

proxy/ptunnel/commands/ssh.go
aaahrens: I'll add the agent command as a separate PR but I think that's a good idea. 

At the very least the agent should just set the env in the host for the ssh agent so ssh-add can be used directly, or ssh directly. But that's all configurable.

ccontavalli (approved):

PR #68, opened 2021-04-08
kbuild/utils: Buildroot patch

stefanoen:
This buildroot patch allows the creation of a rootfs image that takes
care of automatically running all KUnit tests at startup.

See the patch commit message for more info.

### DIFF
--- a/kbuild/utils/README.md
+++ b/kbuild/utils/README.md
@@ -5,7 +5,17 @@ In short, starting from a compiled custom kernel, it can generate a .tar.gz with
 This is particularly useful for generating .tar.gz packages to be used for testing,
 since most of the kernel config used for testing are usually disabled in the debian and ubuntu repositories (e.g. KUnit).
 
-# Workflow
+The *buildroot_2020.11.2-Add-scripts-and-config-for-kunit-rootfs-image.patch* patch can be applied to buildroot 2020.11.2 release (should work for newer releases as well, the patch doesn't change buildroot internal) to generate a slim 60MiB image, suitable for executing KUnit tests defined through the *kernel_test* bazel rule.
+This patch does the following:
+* Provide buildroot with a sane default configuration for the kernel testing use case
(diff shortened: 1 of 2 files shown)


ccontavalli:
Can you add a bit more in a README? (or patch description?) about how this patch is supposed to be used?

I assume it is applied against a clean kernel tree? Is the rootfs than created by a normal `make` in that tree? or are extra steps necessary? What is the created image based on?

stefanoen:
Agreed, without a README this is sort of cryptic (I was planning on writing one, then I forgot).

The general idea is that this provides a default buildroot config which runs the custom script `kunit_post_build.sh` at the end of the build process, hooking the custom script `run_kunit_tests.sh` to run when the image is launched.

---

While I come up the README, I'm leaving this info (which were scattered inside the patch). 

Patch commit message:
```
Add scripts and config for our rootfs image

Adds the .config and scripts used to generate our slim 60 MiB image that
bypasses the login prompt and automatically loads the kernel modules
exposed through hostfs (triggering KUnit tests execution).

The kunit_post_build.sh script, as its name suggests, is run after building
the base rootfs image. It makes it possible to bypass the login prompt and
creates a new startup script required for running KUnit tests.

Instead, the run_kunit_tests.sh script takes care of running the tests
accordingly to our model.

Read the comments inside both scripts for more info.
```

kunit_post_build.sh documentation
```
# Script to be run after building the filesystem but before generating the
# image (to do that set the config BR2_ROOTFS_POST_BUILD_SCRIPT to point
# to this file).
# This does the following:
# - Modify the default getty prompt to a simple shell, bypassing the login
#   prompt.
# - Creates a new startup script that mounts the hostfs and loads all kernel
#   modules available there (triggering kunit tests execution).
```

run_kunit_tests.sh documentation
```
# Simple script to mount what has been exposed through hostfs and then load
# all the kernel modules made available.
```

ccontavalli (approved):

PR #69, opened 2021-04-12
Slight edit to the auth.proto interface and impl of the interface. Also slight edit to the kcerts package

aaahrens:


astore/rpc/auth.proto
```diff
@@ -28,6 +28,7 @@ message TokenResponse {
 
 message HostCertificateRequest {
   bytes hostcert = 1; // The public key of the host that will be returned as signed by the CA
+  repeated string hosts = 2; // A list of DNS names you wish for the host to have.
```
ccontavalli, line 31: this can also be IP addresses, right? nit: edit the comment.

astore/server/auth/auth.go
```diff
@@ -38,6 +38,22 @@ type Server struct {
 	log                   logger.Logger
 }
 
+func (s *Server) HostCertificate(ctx context.Context, request *auth.HostCertificateRequest) (*auth.HostCertificateResponse, error) {
+	b, _ := pem.Decode(request.Hostcert)
+	pubKey, err := ssh.ParsePublicKey(b.Bytes)
```
ccontavalli, line 43: from the documentation of `pem.Decode`, `b` could be nil here? In that case, `b.Bytes` would probably crash?

I recommend adding a ceck `if b == nil { return .. ` some error.

ccontavalli (approved):
Please fix the comments, and then feel free to merge.

astore/rpc/auth.proto
aaahrens: SSH CA's have the same limitations as TLS CA's where they can only trust dns names. IP's are invalid in principals (actually they're just a NoOp). SSHing into a box with valid certificates will always provoke a trust this host prompt.

astore/server/auth/auth.go
aaahrens: Ah nice catch, fixed

PR #70, opened 2021-04-13
Change Auth Flow to not send Priv Key over the wire 

aaahrens:
I still need to test this but the code is here to review.

Public Key is send with the token request and not with the auth request:
Reason: The way that the auth requests cookie is sent would requrie significant refactoring in main.go of astore server.
Downside:  A public key isn't tied to a per login flow.

### DIFF
--- a/astore/rpc/auth.proto
+++ b/astore/rpc/auth.proto
@@ -15,15 +15,15 @@ message AuthenticateResponse {
 
 message TokenRequest {
   string url = 1; // URL returned by server.
+  bytes publickey = 2; // Public key to be signed by the server. Optional.
 }
 
 message TokenResponse {
   bytes nonce = 1; // Nonce used for encryption.
   bytes token = 2; // Encrypted token. Requires the private key corresponding to the public key supplied to open.
-  bytes key = 3; // Private Key used to login
-  bytes cert = 4; // Certificate signed to be used with the Private Key
-  bytes capublickey = 5; // CA Public Key to be added to the authenticated client
-  repeated string cahosts = 6; // List of hosts the CA should be trusted for
+  bytes cert = 3; // Certificate signed to be used with the Private Key, is a signed version of the public key sent in the TokenRequest
+  bytes capublickey = 4; // CA Public Key to be added to the authenticated client
+  repeated string cahosts = 5; // List of hosts the CA should be trusted for
 }
 
 message HostCertificateRequest {
--- a/astore/rpc/auth.proto
+++ b/astore/rpc/auth.proto
@@ -21,9 +21,9 @@ message TokenRequest {
 message TokenResponse {
   bytes nonce = 1; // Nonce used for encryption.
   bytes token = 2; // Encrypted token. Requires the private key corresponding to the public key supplied to open.
-  bytes cert = 4; // Certificate signed to be used with the Private Key, is a signed version of the public key sent in the TokenRequest
-  bytes capublickey = 5; // CA Public Key to be added to the authenticated client
-  repeated string cahosts = 6; // List of hosts the CA should be trusted for
+  bytes cert = 4; // Certificate signed to be used with the Private Key, is a signed version of the public key sent in the TokenRequest.
+  bytes capublickey = 5; // CA Public Key to be added to the authenticated client.
+  repeated string cahosts = 6; // List of hosts the CA should be trusted for.
 }
 
 message HostCertificateRequest {
--- a/astore/rpc/auth.proto
+++ b/astore/rpc/auth.proto
@@ -21,9 +21,9 @@ message TokenRequest {
 message TokenResponse {
   bytes nonce = 1; // Nonce used for encryption.
   bytes token = 2; // Encrypted token. Requires the private key corresponding to the public key supplied to open.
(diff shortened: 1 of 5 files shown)


astore/rpc/auth.proto
```diff
@@ -15,15 +15,15 @@ message AuthenticateResponse {
...
 }
 
 message TokenResponse {
   bytes nonce = 1; // Nonce used for encryption.
   bytes token = 2; // Encrypted token. Requires the private key corresponding to the public key supplied to open.
-  bytes key = 3; // Private Key used to login
-  bytes cert = 4; // Certificate signed to be used with the Private Key
-  bytes capublickey = 5; // CA Public Key to be added to the authenticated client
-  repeated string cahosts = 6; // List of hosts the CA should be trusted for
+  bytes cert = 3; // Certificate signed to be used with the Private Key, is a signed version of the public key sent in the TokenRequest
```
ccontavalli, line 24: I believe it does not matter in this instance, but generally you never re-number fields, that way protocol buffer maintains wire compatibility, addressing some compatibility / versioning issues.

astore/server/auth/auth.go
```diff
@@ -132,8 +131,17 @@ func (s *Server) Token(ctx context.Context, req *auth.TokenRequest) (*auth.Token
 				Token: box.Seal(nil, []byte(authData.Cookie), &nonce, (*[32]byte)(clientPub), (*[32]byte)(s.serverPriv)),
 			}, nil
 		}
+		// if the ca signer was present, continuing with public keys
+		b, _ := pem.Decode(req.Publickey)
+		if b == nil {
+			return nil, errors.New("public key not sent on a CA enabled server")
```
ccontavalli, line 137: Make this block of code optional on receiving req.PublicKey. I'd like to roll out the new CA server even before clients are updated. If you make this code not return error when req.PublicKey is not provided, it is up to the client to ask for the key to be signed or not.

lib/kcerts/ssh.go
```diff
@@ -209,7 +209,7 @@ func GenerateUserSSHCert(ca ssh.Signer, certType uint32, principals []string, tt
 	return priv, cert, nil
 }
 
-func makeKeys() (*rsa.PrivateKey, ssh.PublicKey, error) {
+func MakeKeys() (*rsa.PrivateKey, ssh.PublicKey, error) {
```
ccontavalli, line 212: I'd recommend defaulting to ed25519. Keys are smaller, faster to generate, faster to use, and arguably more secure.

ccontavalli (commented):
Looks like a great start, minor nits.

lib/kcerts/ssh.go
aaahrens: I agree but the refactor is large enough that it would warrant its own PR, and I'd like to do it after we get auth server testing up (just to make sure there aren't any weird ssh bugs)

astore/server/auth/auth.go
aaahrens: Sure thing, implemented

astore/rpc/auth.proto
aaahrens: you know what, I did this knowing it wouldn't change anything, but it's the principal of the matter. I reverted the numbering.

lib/kcerts/ssh.go
ccontavalli: FWIW, the CERT currently used by the auth I deployed last week is an ED25519 cert, it worked so far? it at least looked like it.

astore/rpc/auth.proto
```diff
@@ -15,13 +15,13 @@ message AuthenticateResponse {
...
   string url = 1; // URL returned by server.
+  bytes publickey = 2; // Public key to be signed by the server. Optional.
 }
 
 message TokenResponse {
   bytes nonce = 1; // Nonce used for encryption.
   bytes token = 2; // Encrypted token. Requires the private key corresponding to the public key supplied to open.
-  bytes key = 3; // Private Key used to login
-  bytes cert = 4; // Certificate signed to be used with the Private Key
+  bytes cert = 4; // Certificate signed to be used with the Private Key, is a signed version of the public key sent in the TokenRequest
```
ccontavalli, line 24: nit: periods at end of sentences. Capitalization, line 128.

lib/kcerts/ssh.go
ccontavalli: (I agree on a separate PR, BTW)

astore/server/auth/auth.go
```diff
@@ -125,15 +124,22 @@ func (s *Server) Token(ctx context.Context, req *auth.TokenRequest) (*auth.Token
 		if _, err := io.ReadFull(s.rng, nonce[:]); err != nil {
 			return nil, status.Errorf(codes.Internal, "could not generate nonce - %s", err)
 		}
-		// if the ca signer is nil that means the CA was never passed in flags
-		if s.caSigner == nil {
+		b, _ := pem.Decode(req.Publickey)
+		// if the ca signer is nil that means the CA was never passed in flags, if the request never sent a public key
+		// then so ssh certs will be sent back
```
ccontavalli, line 129: same as above.

ccontavalli: also below (comments are full sentences)

PR #74, opened 2021-04-21
[DO NOT MERGE] Technically working. Blocking on Pr #72

aaahrens:
Upstream changes for #72 . Technically working with some limitations. 

Adds in a couple of commands to help debug. Also some janky PAM stuff is going on. that might be need to be edited.


machinist/mnode/commands.go
```diff
@@ -0,0 +1,85 @@
...
+		newN, err := New(config)
+		if err != nil {
+			return nil, err
+		}
+		return newN, err
+	}
+
+	// Global Relate Flags
+	c.PersistentFlags().StringArrayVar(&config.Tags, "tags", []string{}, "the list of tags you want this node to have. Setting this will unset the cache")
+	c.PersistentFlags().StringVar(&config.Name, "name", "no-name", "the name of this node. If a node already exists with this name, polling the machinist server will fail")
```
ccontavalli, line 30: default to: `os.Hostname` for `--name`

machinist/mnode/commands.go
```diff
@@ -0,0 +1,85 @@
...
+	c.PersistentFlags().StringArrayVar(&config.DnsNames, "dns-names", []string{"localhost"}, "the list of dns names you want this node to have")
```
ccontavalli, line 31: uhm ... what happens when multiple hosts register as localhost? shouldn't this list be empty?

Also, I would call the flag something like `--aliases` or `--extra-names` ... dns is an implementation detail :) eg, allocation won't care about DNS, but will work by alias as well?

machinist/mnode/commands.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// NSS Autouser flags
+	c.PersistentFlags().StringVar(&config.LibNssConfLocation, "nss-autouser-conf", "/etc/nss-autouser.conf", "the file location of libnss autouser configuration file")
+
+	// Pam Flags
+	c.PersistentFlags().StringVar(&config.PamSecurityLocation, "pam-account-script-file", "/etc/security/pam_script_acct", "the location where to save the machinist host key")
+	c.PersistentFlags().StringVar(&config.PamSSHDLocation, "pam-sshd-file", "/etc/pam.d/sshd", "the location where to save PAM sshd configuration")
+
+	// SSHD Related Flags
+	c.PersistentFlags().BoolVar(&config.AutoRestartSSHD, "auto-restart-ssh", true, "if enroll is is successful, auto restart sshd by calling service sshd-restart")
```
ccontavalli, line 78: nit: s/is is/is/

machinist/mnode/assets/install_libpam-script.sh
```diff
@@ -0,0 +1,14 @@
+declare -A osInfo;
```
ccontavalli, line 1: arrays are not POSIX shell features, only a subset of shells support arrays.
Start the script with:

```
#!/usr/bin/env bash
```

or even `/bin/bash`, most systems have bash installed today.

machinist/mnode/mnode.go
```diff
@@ -67,3 +76,91 @@ func (n *Node) BeginPolling() error {
 		}
 	}
 }
+
+// Todo(adam): perform rollbacks if enroll fails
```
ccontavalli, line 80: nit: s/Todo/TODO/

machinist/mnode/commands.go
```diff
@@ -0,0 +1,85 @@
...
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			n, err := factoryFunc()
+			if err != nil {
+				return err
+			}
+			return n.Enroll()
+		},
+	}
+	// General Flags
```
ccontavalli, line 67: nit: end comments with a period.

machinist/mnode/mnode.go
```diff
@@ -67,3 +76,91 @@ func (n *Node) BeginPolling() error {
...
+	}
+	if err := ioutil.WriteFile(n.config.HostKeyLocation, pemBytes, 0644); err != nil {
+		return err
+	}
+	return nil
+}
+
+func anyFileExist(names ...string) (string, bool) {
+	for _, name := range names {
+		if _, err := os.Stat(name); err != nil {
```
ccontavalli, line 158: nit:
```
if _, err := os.Stat(name); err != nil && os.IsNotExist(err) {
  continue
}
```

reduce nesting by using an end.

machinist/mnode/mnode.go
```diff
@@ -67,3 +76,91 @@ func (n *Node) BeginPolling() error {
...
+			if os.IsNotExist(err) {
+				continue
+			}
```
ccontavalli, line 161: minor issue: Stat could fail for any number of reasons. Eg, ssh is not installed, the parent directory does not exist, or is not writeable, ... If any of those failures happens, the code will output a message saying that the file already exists, very confusing.

I would personally use `lib/multierror` here, and change the function to ONLY return error. No error: none of the files exists. error: some of the file exists, or something else went sideways.

Basically, if `IsNotExist(err) true` -> no error added. Otherwise, error added. Should also simplify the logic, and show all errors in one go, which is more user friendly (eg, now, let's say two files exist, I remove one, run again, error again).

machinist/mnode/templates.go
```diff
@@ -0,0 +1,116 @@
...
+	l.Infof("Successfully fetched the binary, installing now")
+	// TODO(adam): configure arm support for filepath writing and nss building
+	if err = ioutil.WriteFile("/lib/x86_64-linux-gnu/libnss_autouser.so.2", nssBinary, 0600); err != nil {
+		return err
+	}
+	out, err := ioutil.ReadFile("/etc/nsswitch.conf")
+	if err != nil {
+		return err
+	}
+	if !strings.Contains(string(out), "passwd:         files autouser systemd") {
```
ccontavalli, line 68: If you want to do this simply, you could run a .sh script or a simple regex replace with something like:

`passwd:(\s+)files(\s+.*)$` to replace with `passwd:\1files autouser\2`

machinist/mnode/mnode.go
ccontavalli, line 80: (not a requirement for this PR)

A simple way to implement rollback could be:
1) Array of functions to undo, simple `defer` invoking each function in order.
2) Each file modified, you have a function saving the original in a /tmp/ directory. Eg, a function like "SaveForRollback(path) which saves a file in /tmp/ directory, returns callback to invoke in defer that puts back the original file in place".
3) Each file created, you have a function deleting it.

(should not be too hard to implement)

ccontavalli (commented):
This looks very reasonable, only minor nits in the PR. Would be happy to approve in short order.

machinist/mnode/commands.go
aaahrens: So Dns Names is a misnomer, I was actually using these as ssh-names. Since this isn't in userland, I want it to be pretty specific for the operator. I renamed it to be more clear

We use localhost as a default because of the proxy, and because localhost is used for a lot of bastion type configs

machinist/mnode/templates.go
aaahrens: I'm going to do this later. It's changed from files to compat in ubuntu 20.04, and I feel like nsswitch should be set by the ansible script or provisioner.

machinist/mnode/mnode.go
aaahrens: Yeah not difficult at all, but it would significantly increase the LOC of this PR. So I'll add it later.

machinist/mnode/assets/install_libpam-script.sh
aaahrens: Done

machinist/mnode/mnode.go
```diff
@@ -67,3 +77,90 @@ func (n *Node) BeginPolling() error {
...
+	if err := ioutil.WriteFile(n.config.HostKeyLocation, pemBytes, 0644); err != nil {
+		return err
+	}
+	return nil
+}
+
+func anyFileExist(names ...string) error {
+	var errs []error
+	for _, name := range names {
+		if _, err := os.Stat(name); err != nil {
```
ccontavalli, line 161: os.IsNotExist is supposed to be ignored. This code likely no longer works on a clean machine? Test it?

ccontavalli (approved):
Modulo the comment below, looks good. Please address the comment, test the code, and merge.

PR #75, opened 2021-04-22
Fully Support 25519 Keys as CA base and have certificates revolve around them

aaahrens:
Adds support for ED25519 Certificates. Doing this now because machinist hits a bump when using certs on OpenSSH  > 8.1.

This means it should work for all forms of sshd(8)

Unfortunately it might get a little hairy if we also want to support sha256 RSA, but not sure if it's strictly necessary to do so. I think making a configurable signer and cert provider shouldn't be too hard to do.


astore/client/auth/auth.go
```diff
@@ -74,14 +74,18 @@ func (c *Client) Login(username, domain string, o LoginOptions) (string, error)
...
+		}
+	}()
+	privateKey, pubKey, err := kcerts.MakeKeys(kcerts.GenerateDefault)
 	if err != nil {
 		return "", fmt.Errorf("error generating ssh credentials: %w", err)
 	}
 	treq := &auth.TokenRequest{
 		Url:       ares.Url,
-		Publickey: pubKey,
+		Publickey: ssh.MarshalAuthorizedKey(pubKey),
```
ccontavalli, line 88: no, this should not work? `kcerts.MakeKeys` now returns a marshalled key directly?

astore/client/auth/auth.go
```diff
@@ -74,14 +74,18 @@ func (c *Client) Login(username, domain string, o LoginOptions) (string, error)
...
 	fmt.Printf("\t%s\n\nTo complete authentication with @%s.\n"+
 		"I'll be waiting for you, but hurry! The request may timeout.\nHit Ctl+C with no regrets to abort.\n", ares.Url, domain)
-	browser.OpenURL(ares.Url)
-	pubKey, privateKey, err := kcerts.MakeKeys(kcerts.GenerateDefault)
+	go func() {
+		if err := browser.OpenURL(ares.Url); err != nil {
+			o.Logger.Infof("error opening url, you must open it yourself")
+		}
+	}()
+	privateKey, pubKey, err := kcerts.MakeKeys(kcerts.GenerateDefault)
```
ccontavalli, line 82: I changed MakeKeys in my PR to follow the same convention as other library functions, now returns `pubKey, privateKey, err`, very unlucky that none of the uses below caused the compiler to freak out.

astore/client/auth/auth.go
ccontavalli: as a side note, this function as is should not be working. Did you forget to push another commit with fixes? How did you test this?

astore/server/auth/factory.go
```diff
@@ -83,13 +87,30 @@ func WithCA(fileContent []byte) Modifier {
...
 		}
-		signer, err := ssh.ParsePrivateKey(fileContent)
+		caPrivateKey, err := ssh.ParseRawPrivateKey(fileContent)
 		if err != nil {
 			return fmt.Errorf("Could not parse CA key - %w", err)
 		}
-		server.caSigner = signer
-		server.marshalledCAPublicKey = ssh.MarshalAuthorizedKey(signer.PublicKey())
-		return nil
+		// Todo(adam): make parsing existing keys cleaner
```
ccontavalli, line 94: nit: s/Todo/TODO, period at end of comment.

lib/kcerts/ssh.go
```diff
@@ -268,8 +205,101 @@ func SignPublicKey(ca ssh.Signer, certType uint32, principals []string, ttl time
...
 	return cert, nil
 }
+
+func OpenSSHEncode21559PrivateKey(key ed25519.PrivateKey) ([]byte, error) {
+	b := openSSHMarshalED25519PrivateKey(key)
+	return pem.EncodeToMemory(&pem.Block{Type: "OPENSSH PRIVATE KEY", Bytes: b}), nil
+
+}
+
+// Note: copited from https://github.com/mikesmitty/edkey/blob/master/edkey.go
```
ccontavalli, line 224: we can't just copy in this code, see LICENSE here: https://github.com/mikesmitty/edkey/blob/master/LICENSE, as a bare minimum, we should put it in a directory, and have the LICENSE file in it.

Better: can we import the library like any other go dependency? WORKSPACE and such.

lib/kcerts/ssh.go
```diff
@@ -268,8 +205,101 @@ func SignPublicKey(ca ssh.Signer, certType uint32, principals []string, ttl time
...
+// TODO(adam): support passwords
+
+//* Writes ed25519 private keys into the new OpenSSH private key format.
+//I have no idea why this isn't implemented anywhere yet, you can do seemingly
```
ccontavalli, line 228: nit: space between // and comment.

This sounds very suspicious, though: `ssh.MarshalAuthorizedKey` works just fine on ed25519 keys. The function [here](https://github.com/golang/crypto/blob/83a5a9bb288b/ssh/keys.go#L281) just invokes the [code here](https://github.com/golang/crypto/blob/83a5a9bb288b/ssh/keys.go#L572).

Why do we need this?

lib/kcerts/keys.go
```diff
@@ -0,0 +1,158 @@
...
+// Only keys supported by x/crypto/ssh.NewPublicKey are supported.
+type SSHKeyGenerator func() (*PrivateKey, ssh.PublicKey, error)
+
+// MakeKeys uses the speficied key generator to create a pair of ssh keys.
+//
+// The first return value is a marshalled version of the public key,
+// a binary blob suitable for transmission.
+// The second return value is the private key in the original format.
+// This is generally directly usable with functions like agent.Add.
+func MakeKeys(generator SSHKeyGenerator) (*PrivateKey, ssh.PublicKey, error) {
```
ccontavalli, line 96: ah! You changed it back to return public, private. The comment needs to be updated then.

lib/kcerts/keys.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil, err
+	}
+	return res, nil
+}
+
+type PrivateKeyProvider interface {
+	Signer() crypto.Signer
+	SigningAlgo() string
+	Raw() interface{}
+	SSHPemEncode() ([]byte, error)
```
ccontavalli, line 30: why use PEM encoding at all at this point? ssh has its own marshal / unmarshal, that seems to work just fine, and does not land us in uncharted territory (eg, ED25519 not having a PEM supported encoding, ssh-keygen -m PEM -e ed25519 key fails).

lib/kcerts/keys.go
```diff
@@ -0,0 +1,158 @@
...
+	"crypto/ed25519"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+	"golang.org/x/crypto/ssh"
+	"strings"
+)
+
+type PrivateKey struct {
```
ccontavalli, line 14: This encapsulation of PrivateKeyProvider... seems a bit complex/overkill? I think I understand why you do it, but there's a single use case of it.

Personally, I would eliminate this `PrivateKey` struct definition, and let the [one caller here](https://github.com/enfabrica/enkit/pull/75/files#diff-c5353184ccf4457a5a29aaa79afa329fd3d795f1a475f55806d070b80646040bR208), just call NewSSHSigner.

(I think it would make the code easier to understand both for those touching the library, and users of Signer, who can clearly see the need of conversion)

lib/kcerts/keys.go
ccontavalli: Assuming it works (and I believe it was working for me, in previous PR - it just required a new server release to not have the PEM encoding in the way).

astore/client/auth/auth.go
ccontavalli: never mind, just noticed that MakeKeys is back to returning non-marshalled keys.

lib/kcerts/keys.go
```diff
@@ -0,0 +1,158 @@
...
+type SSHKeyGenerator func() (*PrivateKey, ssh.PublicKey, error)
+
+// MakeKeys uses the speficied key generator to create a pair of ssh keys.
+//
+// The first return value is a marshalled version of the public key,
+// a binary blob suitable for transmission.
+// The second return value is the private key in the original format.
+// This is generally directly usable with functions like agent.Add.
+func MakeKeys(generator SSHKeyGenerator) (*PrivateKey, ssh.PublicKey, error) {
+	private, public, err := generator()
```
ccontavalli, line 97: MakeKeys can now be changed in just:
`return generator()`

?

If that's the case, I would remove the `generator` argument entirely, and have:

var MakeDefaultKeys = MakeRSAKeys

rname GenerateRSA into MakeRSAKeys, GenerateED25519 into MakeED25519?

astore/client/auth/auth.go
```diff
@@ -74,14 +74,18 @@ func (c *Client) Login(username, domain string, o LoginOptions) (string, error)
 	}
 	fmt.Printf("\t%s\n\nTo complete authentication with @%s.\n"+
 		"I'll be waiting for you, but hurry! The request may timeout.\nHit Ctl+C with no regrets to abort.\n", ares.Url, domain)
-	browser.OpenURL(ares.Url)
-	pubKey, privateKey, err := kcerts.MakeKeys(kcerts.GenerateDefault)
+	go func() {
```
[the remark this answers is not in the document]
aaahrens: I put this in #80

lib/kcerts/ssh.go
aaahrens: Those are for public keys. This is for private keys.

lib/kcerts/keys.go
aaahrens: I changed it back, your current style is right

lib/kcerts/keys.go
aaahrens: Ah yeah you're right made it much simpler.

lib/kcerts/keys.go
aaahrens: Changed to be more functional

lib/kcerts/keys.go
aaahrens: ssh only has marshalling for public keys, not private. This is an interface for private

lib/kcerts/keys.go
```diff
@@ -8,22 +8,17 @@ import (
...
 )
 
-type PrivateKey struct {
-	Key PrivateKeyProvider
-}
-
-func (e PrivateKey) Signer() (ssh.Signer, error) {
-	res, err := NewSSHSigner(e.Key.Signer(), e.Key.SigningAlgo())
+func NewSigner(key PrivateKey) (ssh.Signer, error) {
+	res, err := NewSSHSigner(key.Signer(), key.SigningAlgo())
```
ccontavalli, line 14: nit: Make this a one liner?
```
return NewSSHSigner(...)
```

Also... the only thing this function does is call `.Signer()` and `.SigningAlgo()` for us? not sure it has much value, I would consider just removing it.

astore/client/auth/auth.go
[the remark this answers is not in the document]
ccontavalli: I still see the change here? remove?

lib/kcerts/signer.go
```diff
@@ -0,0 +1,37 @@
+package kcerts
+
+import (
```
ccontavalli, line 3: go fmt

lib/kcerts/signer_test.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+var tableTestTypes = []kcerts.SSHKeyGenerator{kcerts.GenerateED25519, kcerts.GenerateRSA}
+
+// TestSha256Signer_PublicKey tests all possible combinations of supported kcerts.PrivateKey signing ssh.PublicKeys
+// It will sign the following ssh certs with the custom algos by their providers
+func TestSha256Signer_PublicKey(t *testing.T) {
+	for _, sourceType := range tableTestTypes {
+		for _, toSignType := range tableTestTypes {
+			_, sourcePrivKey, err := sourceType()
```
ccontavalli, line 20: the idiomatic way for looping over a test to try a bunch of combinations is to use `subtests`, eg, in the innermost loop, have:
```
t.Run(fmt.Sprintf("Source:%v,Sign:%v", sourceType, toSignType"), func (t *testing.T) {
   // code of your test
})
```
there's a number of reasons why this is better, see https://blog.golang.org/subtests

ccontavalli (commented):
really, only very minor nits left. The only thing I think you need to fix now is removing the code that is now in #80.

lib/kcerts/keys.go
aaahrens: I'm going to keep it in here to mimic the ssh.NewSignerFromSigner function. I'm keeping it in because I'm going to be getting more use out of the factory func once mTLS hits.

astore/server/auth/factory.go
```diff
@@ -83,13 +87,30 @@ func WithCA(fileContent []byte) Modifier {
...
-		signer, err := ssh.ParsePrivateKey(fileContent)
+		caPrivateKey, err := ssh.ParseRawPrivateKey(fileContent)
 		if err != nil {
 			return fmt.Errorf("Could not parse CA key - %w", err)
 		}
-		server.caSigner = signer
-		server.marshalledCAPublicKey = ssh.MarshalAuthorizedKey(signer.PublicKey())
-		return nil
+		// TODO(adam): make parsing existing keys cleaner
+		if key, ok := caPrivateKey.(*ed25519.PrivateKey); ok {
```
ccontavalli, line 95: this smells like code that should go in `kcerts`, simple typeswitch that returns kcerts.From... depending on key type.

ccontavalli: with common MarshalAuthorizedkey

ccontavalli (approved):
I'm approving now, but:
1) REMOVE THE BROWSER FIX - it has nothing to do with this PR.
2) Do or don't move the function suggested to kcerts now.
3) Make sure all tests pass!

PR #80, opened 2021-04-27
Adds retry lib support for auth, removes double layering of auth clients, fixes browser.OpenUrl and makes login more testable.  

aaahrens:


astore/client/auth/auth.go
```diff
@@ -1,153 +0,0 @@
...
-	if err != nil {
-		return "", fmt.Errorf("Could not contact the authentication server. Is your connectivity working? Is the server up?\nFor debugging: %w", err)
-	}
-
-	servPub, err := common.KeyFromSlice(ares.Key)
-	if err != nil {
-		return "", fmt.Errorf("server provided invalid key - please retry - %s", err)
-	}
-
-	if username != "" {
```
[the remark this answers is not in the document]
ccontavalli: a good compromise solution could be to have a vector of `func()` invoked to display the auth message to the user.
One opens the browser, another prints on the screen, ... etc. or breaking the function in two steps, and have an API like GetAuthToken() and ActuallyLogin() or something.

lib/kauth/login.go
```diff
@@ -0,0 +1,121 @@
...
+
+// PerformLogin will login with the provider auth client, retry and logger. It does not care about the cache.
+// If you wish to save the result, please call SaveCredentials
+func PerformLogin(authClient auth.AuthClient, l logger.Logger, repeater *retry.Options, username, domain string) (*EnkitCredentials, error) {
+	rng := rand.New(srand.Source)
+	pubBox, privBox, err := box.GenerateKey(rng)
+	if err != nil {
+		return nil, err
+	}
+	sshPub, sshPriv, err := kcerts.MakeKeys(kcerts.GenerateED25519)
```
ccontavalli, line 43: this is not needed until much later, move it down closer to where it is used?
makes the code harder to follow.

lib/kauth/login.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return nil, err
+	}
+	areq := &auth.AuthenticateRequest{
+		Key:    (*pubBox)[:],
+		User:   username,
+		Domain: domain,
+	}
+	l.Infof("Retrieving authentication url.")
+	ares, err := authClient.Authenticate(context.TODO(), areq)
```
ccontavalli, line 53: not for this PR: we need to start implementing timeouts here, context.

lib/kauth/login.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Could not contact the authentication server. Is your connectivity working? Is the server up?\nFor debugging: %w", err)
+	}
+	if username != "" {
```
ccontavalli, line 57: I liked my proposal better of passing a list of functions to "interact with the user". That way, we could do things like disable "opening a browser" from a unit test (does the test open browsers for you? fun?) or fully embed in a non interactive library. I'm ok not doing this in this PR.

lib/kauth/login.go
ccontavalli, line 43: MakeKeys(...) this has changed right? or is about to change? don't forget to update the code after one of the PRs is merged in!

lib/kauth/login.go
```diff
@@ -0,0 +1,121 @@
...
+	browser.Stdout = nil
+	browser.Stderr = nil
+}
+
+type EnkitCredentials struct {
+	Token string
+	// The below fields can be possibly empty if the auth server does not support CA certificates.
+	CaHosts        []string
+	CAPublicKey    string
+	PrivateKey     ed25519.PrivateKey
```
ccontavalli, line 31: can we make this not be explicitly an ed25519.PrivateKey, but eg, the interface you created, or generic ssh object? I don't want any code outside this library to rely on the key being ed25519.

lib/kauth/save.go
```diff
@@ -0,0 +1,45 @@
...
+	"golang.org/x/crypto/ssh"
+	"time"
+)
+
+var NoSSHCredentialsErr = errors.New("no credentials passed in")
+// SaveCredentials saves the passed in credentials to the current ssh-agent. If the credentials are empty, i.e.
+// the EnkitCredentials only contain EnkitCredentials.Token, it will return NoSSHCredentialsErr.
+func SaveCredentials(credentials *EnkitCredentials, store cache.Store, l logger.Logger) error{
+	if len(credentials.CaHosts) == 0 || credentials.SSHCertificate == nil || len(credentials.PrivateKey) == 0 {
+		return NoSSHCredentialsErr
```
ccontavalli, line 18: I would personally just return nil here. Not really a failure, we saved the nil credentials by not writing them anywhere. Makes the calling code simpler.

Also, no user of this library needs to know about this error. The only user ignores it. Add it if we actually ever end up needing it?

lib/kauth/save.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/enfabrica/enkit/lib/logger"
+	"github.com/pkg/errors"
+	"golang.org/x/crypto/ssh"
+	"time"
+)
+
+var NoSSHCredentialsErr = errors.New("no credentials passed in")
+// SaveCredentials saves the passed in credentials to the current ssh-agent. If the credentials are empty, i.e.
+// the EnkitCredentials only contain EnkitCredentials.Token, it will return NoSSHCredentialsErr.
+func SaveCredentials(credentials *EnkitCredentials, store cache.Store, l logger.Logger) error{
```
ccontavalli, line 16: thinking aloud... should this be a method on EnkitCredentials? Rather than a free standing function?

lib/client/commands/login.go
```diff
@@ -79,23 +81,23 @@ func (l *Login) Run(cmd *cobra.Command, args []string) error {
...
 	}
-
-	options := auth.LoginOptions{
-		Context: l.base.Context(),
-		MinWait: l.MinWaitTime,
-		Store:   l.base.Local,
-	}
-
-	token, err := client.Login(username, domain, options)
+	repeater := retry.New(retry.WithWait(l.MinWaitTime), retry.WithRng(l.rng))
```
ccontavalli, line 88: btw, two reasons I wanted to bring in the retry library:
1) So we'd have the fuzzying, WithFuzzy(1 second)
2) So we'd have the retry parameters exposed on the CLI like any other parameter.

not necessary to do for this PR.

lib/kauth/save.go
ccontavalli: either way is fine, do it or don't at your leisure.

lib/kauth/login.go
```diff
@@ -0,0 +1,121 @@
...
+	SSHCertificate *ssh.Certificate
+}
+
+// PerformLogin will login with the provider auth client, retry and logger. It does not care about the cache.
+// If you wish to save the result, please call SaveCredentials
+func PerformLogin(authClient auth.AuthClient, l logger.Logger, repeater *retry.Options, username, domain string) (*EnkitCredentials, error) {
+	rng := rand.New(srand.Source)
```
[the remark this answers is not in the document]
ccontavalli: as you pass an extra parameter, I'm wondering if you should wrap them in a struct with options.
The long list of parameters obfuscates what this does, something most wouldn't care about (this last part is optional)

lib/client/commands/login.go
```diff
@@ -79,23 +81,23 @@ func (l *Login) Run(cmd *cobra.Command, args []string) error {
...
+	enCreds, err := kauth.PerformLogin(auth.NewAuthClient(conn), l.base.Log, repeater, username, domain)
 	if err != nil {
 		return err
 	}
+	if err := kauth.SaveCredentials(enCreds, l.base.Local, l.base.Log); err != nil && err != kauth.NoSSHCredentialsErr {
+		l.base.Log.Warnf("error saving credentials, err: %v", err)
+		return err
+	}
+
+	// TODO(adam): delete below when we are comfortable migrating from the token to pure ssh certificates
```
ccontavalli, line 98: I suspect this file is not going to go fully away. We still need to know the name of the user, so we don't have to ask it every time. I'd like the certificate of the remote server to be stored there as well, so we can eg, pin TLS or grpc auth to this certificate after login.

lib/client/commands/login.go
```diff
@@ -79,23 +81,23 @@ func (l *Login) Run(cmd *cobra.Command, args []string) error {
...
-		Store:   l.base.Local,
-	}
-
-	token, err := client.Login(username, domain, options)
+	repeater := retry.New(retry.WithWait(l.MinWaitTime), retry.WithRng(l.rng))
+	enCreds, err := kauth.PerformLogin(auth.NewAuthClient(conn), l.base.Log, repeater, username, domain)
 	if err != nil {
 		return err
 	}
+	if err := kauth.SaveCredentials(enCreds, l.base.Local, l.base.Log); err != nil && err != kauth.NoSSHCredentialsErr {
```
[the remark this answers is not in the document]
ccontavalli: As per other comment, I recommend just getting rid of this error. It gets ignored anyway.

lib/kauth/save.go
```diff
@@ -0,0 +1,45 @@
...
+	"fmt"
+	"github.com/enfabrica/enkit/lib/cache"
+	"github.com/enfabrica/enkit/lib/kcerts"
+	"github.com/enfabrica/enkit/lib/logger"
+	"github.com/pkg/errors"
+	"golang.org/x/crypto/ssh"
+	"time"
+)
+
+var NoSSHCredentialsErr = errors.New("no credentials passed in")
```
ccontavalli, line 13: don't forget to eliminate this error.

lib/kauth/login.go
```diff
@@ -0,0 +1,121 @@
...
+		fmt.Printf("Dear %s, please visit:\n\n", username)
+	} else {
+		fmt.Printf("Kind human, please visit:\n\n")
+	}
+	fmt.Printf("\t%s\n\nTo complete authentication with @%s.\n"+
+		"I'll be waiting for you, but hurry! The request may timeout.\nHit Ctl+C with no regrets to abort.\n", ares.Url, domain)
+	l.Infof("Authentication url is %s, attempting to open with your Os's default browser", ares.Url)
+	// browser.OpenURL blocks depending on permissions level and OS. By running it in a goroutine, we ensure that
+	// the login process does not get stuck waiting for the browser window to be closed.
+	go func() {
```
ccontavalli, line 67: This PR is not just a refactor! It is solving a bug! Bugfixes should go in THEIR OWN PR.
That way, we can cherrypick, or if we rollback your larger refactor due to a separate bug, we don't lose the dedicate bugfix.

This is pretty minor, but please make sure your merge COMMIT MESSAGE includes the fact that there is a bugfix here. From `git log`, my assumption for this PR is that this should be a noop in terms of behavior.

lib/kauth/login.go
aaahrens: Already pulled

lib/kauth/save.go
aaahrens: I would rather this be a function, just stylistically. Also means we don't need to go through the hoopla of embedding interfaces in the EnCreds

lib/kauth/save.go
aaahrens: Removed the error and updated for the latest pr

lib/kauth/login.go
aaahrens: Updated the title so that when it gets squashed it'll show up.

lib/client/commands/login.go
aaahrens: Setting the user could be something entirely separate. IIRC we need the username for COOKIE_HOST and that should be it? Setting the name sort of like git config and performing actions based on the user-set config file seems okayish to me. 

Rough outline: NasshProxy.ServeCookie gets called. It blocks harder on authentication, Fragment target is now cred.Username + "@" + np.relayHost on authenticator success. 

If I set my username to Brad, tunnel expects Brad@nss-relay, but it fails  since I can't login as brad@host.

lib/kauth/login.go
aaahrens: Ah later, mainly because things like opening from a browser should be more of a fire and forget, and user interaction outside of copy-paste generally is pretty unreliable imo.

lib/client/commands/login.go
ccontavalli: We should chat, I've lost you on the last comment. Your identity should not be tied to what you're connecting to? you are adam@enfabrica.net when connecting to our proxy, @np.relayHost seems very suspicious to me.
If we add external guests support, they will be eg, bob@company.com, or foo@bar.biz?

ccontavalli (approved):

PR #82, opened 2021-04-29
Fixes HostCert Invalid

aaahrens:
If no principals were specified, it would add []string{""} to principals instead of an empty string. This would lead to the certificate containing "", and the valid username. 

While this didn't affect functionality (since its invalid) just removes some dirty output from sshd since it causes some wonky errors.

### DIFF
--- a/astore/server/auth/factory.go
+++ b/astore/server/auth/factory.go
@@ -117,7 +117,9 @@ func WithCA(fileContent []byte) Modifier {
 func WithPrincipals(raw string) Modifier {
 	return func(server *Server) error {
 		splitString := strings.Split(raw, ",")
-		server.principals = splitString
+		if len(splitString) >= 1 && splitString[0] != "" {
+			server.principals = splitString
+		}
 		return nil
 	}
 }


astore/server/auth/factory.go
```diff
@@ -117,7 +117,9 @@ func WithCA(fileContent []byte) Modifier {
 func WithPrincipals(raw string) Modifier {
 	return func(server *Server) error {
 		splitString := strings.Split(raw, ",")
-		server.principals = splitString
+		if len(splitString) >= 1 && splitString[0] != "" {
+			server.principals = splitString
```
ccontavalli, line 121: Good catch! The problem is that empty string will lead to a split of 1, with an empty string, correct?

It took me a couple seconds to be sure your code above was correct. How about this:
```
raw = strings.TrimSpace(raw)
if raw != "" {
  server.principals = strings.Split(raw, ",")
}
```

ccontavalli (approved):
approved either way, pick your poison.

astore/server/auth/factory.go
aaahrens: Ah I'm going to just merge this as is because I'm going to PR kcobra to have the VarStringArray like standard cobra

PR #83, opened 2021-04-29
Fix compat issue with future auth-server release

aaahrens:
So since this got change to not be pem encoded:
https://github.com/enfabrica/enkit/blob/master/astore/server/auth/auth.go#L137 

This change is needed in order to be compatible. The code without this fix works with the current version of auth. Not sure if we want this to be cross compatible or not (e.g. retry with pem encoded).

### DIFF
--- a/lib/kauth/login.go
+++ b/lib/kauth/login.go
@@ -2,7 +2,6 @@ package kauth
 
 import (
 	"context"
-	"encoding/pem"
 	"fmt"
 	"github.com/enfabrica/enkit/astore/common"
 	"github.com/enfabrica/enkit/astore/rpc/auth"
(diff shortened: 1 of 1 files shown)


ccontavalli:
> So since this got change to not be pem encoded:
> https://github.com/enfabrica/enkit/blob/master/astore/server/auth/auth.go#L137
> 
> This change is needed in order to be compatible. The code without this fix works with the current version of auth. Not sure if we want this to be cross compatible or not (e.g. retry with pem encoded).

The change is fine, but I'm not sure I understand this sentence. If I update the auth server, the client will stop working? Because it expects no PEM encoding. But if we update the client now, the client will also stop working?

Given that this version was never released, I'm ok with a breaking change. Let's coordinate a deployment of auth server tomorrow?

aaahrens:
Agreed for the sync up, would probably want to hold onto merging this and deploying both until Monday/Tuesday. Just because of addition PR's and deploying on Friday is bad luck <.<

ccontavalli:
> Agreed for the sync up, would probably want to hold onto merging this and deploying both until Monday/Tuesday. Just because of addition PR's and deploying on Friday is bad luck <.<

We should merge the PTY fix today, see other PR. I'd like if we can push today anyway, we're still small, impact is small, I can roll back easily. And we need to screw up at least once to be reminded not to push on Friday.

ccontavalli (approved):

PR #84, opened 2021-04-29
Adds SSH commands to enkit, under enkit ssh agent --<commands>

aaahrens:
Not the most complicated, would appreciate feedback on the printing sections of the code.

### DIFF
--- a/lib/kcerts/ssh.go
+++ b/lib/kcerts/ssh.go
@@ -120,7 +120,7 @@ func (a SSHAgent) Principals() ([]AgentCert, error) {
 	for _, key := range keys {
 		p, err := ssh.ParsePublicKey(key.Marshal())
 		if err != nil {
-			return nil, err
+			continue
 		}
 		if cert, ok := p.(*ssh.Certificate); ok {
 			toReturn = append(toReturn, AgentCert{
--- a/lib/kcerts/ssh.go
+++ b/lib/kcerts/ssh.go
@@ -99,6 +99,30 @@ func (a SSHAgent) Valid() bool {
 	return err == nil
 }
 
+// Principals returns a map where the keys are the CA's PKS and the certs identities are the values
+func (a SSHAgent) Principals() (map[string][]string, error) {
+	conn, err := net.Dial("unix", a.Socket)
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Close()
+	keys, err := agent.NewClient(conn).List()
+	if err != nil {
+		return nil, err
+	}
+	toReturn := map[string][]string{}
+	for _, key := range keys {
+		p, err := ssh.ParsePublicKey(key.Marshal())
+		if err != nil {
+			return nil, err
+		}
+		if cert, ok := p.(*ssh.Certificate); ok {
+			toReturn[ssh.FingerprintLegacyMD5(cert.SignatureKey)] = cert.ValidPrincipals
+		}
+	}
+	return toReturn, err
+}
+
 // AddCertificates loads an ssh certificate into the agent.
 // privateKey must be a key type accepted by the golang.org/x/ssh/agent AddedKey struct.
 // At time of writing, this can be: *rsa.PrivateKey, *dsa.PrivateKey, ed25519.PrivateKey or *ecdsa.PrivateKey.
--- a/lib/kcerts/ssh.go
+++ b/lib/kcerts/ssh.go
@@ -117,7 +117,8 @@ func (a SSHAgent) Principals() (map[string][]string, error) {
 			return nil, err
 		}
 		if cert, ok := p.(*ssh.Certificate); ok {
-			toReturn[ssh.FingerprintLegacyMD5(cert.SignatureKey)] = cert.ValidPrincipals
+			md := ssh.FingerprintLegacyMD5(cert.SignatureKey)
+			toReturn[md] = append(toReturn[md], cert.ValidPrincipals...)
 		}
 	}
 	return toReturn, err
--- a/lib/kcerts/ssh.go
+++ b/lib/kcerts/ssh.go
@@ -99,8 +99,14 @@ func (a SSHAgent) Valid() bool {
 	return err == nil
 }
 
+type AgentCert struct {
+	MD5       string
+	Principal []string
+	ValidFor  time.Duration
+}
+
 // Principals returns a map where the keys are the CA's PKS and the certs identities are the values
-func (a SSHAgent) Principals() (map[string][]string, error) {
+func (a SSHAgent) Principals() ([]AgentCert, error) {
 	conn, err := net.Dial("unix", a.Socket)
 	if err != nil {
 		return nil, err
@@ -110,15 +116,18 @@ func (a SSHAgent) Principals() (map[string][]string, error) {
 	if err != nil {
 		return nil, err
 	}
-	toReturn := map[string][]string{}
+	var toReturn []AgentCert
 	for _, key := range keys {
 		p, err := ssh.ParsePublicKey(key.Marshal())
 		if err != nil {
 			return nil, err
 		}
 		if cert, ok := p.(*ssh.Certificate); ok {
-			md := ssh.FingerprintLegacyMD5(cert.SignatureKey)
-			toReturn[md] = append(toReturn[md], cert.ValidPrincipals...)
+			toReturn = append(toReturn, AgentCert{
+				MD5:       ssh.FingerprintLegacyMD5(cert.SignatureKey),
+				Principal: cert.ValidPrincipals,
+				ValidFor:  time.Unix( int64(cert.ValidBefore), 0).Sub(time.Now()),
+			})
 		}
 	}
 	return toReturn, err
--- a/lib/kcerts/ssh.go
+++ b/lib/kcerts/ssh.go
@@ -100,9 +100,9 @@ func (a SSHAgent) Valid() bool {
 }
 
 type AgentCert struct {
-	MD5       string
-	Principal []string
-	ValidFor  time.Duration
+	MD5        string
+	Principals []string
+	ValidFor   time.Duration
 }
 
 // Principals returns a map where the keys are the CA's PKS and the certs identities are the values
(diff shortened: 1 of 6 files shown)


lib/kcerts/ssh.go
```diff
@@ -99,6 +99,31 @@ func (a SSHAgent) Valid() bool {
...
+	defer conn.Close()
+	keys, err := agent.NewClient(conn).List()
+	if err != nil {
+		return nil, err
+	}
+	toReturn := map[string][]string{}
+	for _, key := range keys {
+		p, err := ssh.ParsePublicKey(key.Marshal())
+		if err != nil {
+			return nil, err
```
ccontavalli, line 117: minor nit: the agent could be an arbitrary ssh-agent from 1972, or 2045, when our code will still be running. Rather than return error and fail if the agent handed us something we cannot parse, I would recommend continuing. Use mutlierror to return an error? Clarify that toReturn can contain some values even in case of error? Or return 3 values: valid keys, invalid keys, err

lib/kcerts/ssh.go
```diff
@@ -99,6 +99,31 @@ func (a SSHAgent) Valid() bool {
...
+	conn, err := net.Dial("unix", a.Socket)
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Close()
+	keys, err := agent.NewClient(conn).List()
+	if err != nil {
+		return nil, err
+	}
+	toReturn := map[string][]string{}
```
ccontavalli, line 113: so, maps in go are randomized (for good reasons). If I had to guess, if you are printing this out anywhere you're getting a different order every time. My recommendation would be to use a plain list.

I would also recommend a small struct to reduce risk of errors. Eg, something like:
```
type KnownCert struct {
  MD5 string
  Principal []string
}

toReturn := []KnownCert{}
```

ccontavalli: (this would also show the results in the same order as the ssh-agent, which is a bouns)

proxy/ptunnel/commands/agent.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return err
+	}
+	if config.Print {
+		fmt.Printf("The enkit agent is running at socket %s \n", agent.Socket)
+		fmt.Printf("The enkit agent's pid is %d \n", agent.PID)
+
+	}
+	if config.SetEnv && agent.Valid() {
+		if err := os.Setenv("SSH_AUTH_SOCK", agent.Socket); err != nil {
```
ccontavalli, line 45: The `SetEnv` here are a noop. In `unix` `linux`, there is no (reasonable) way for a process to change the environment variables of the calling process. Eg, `bash` started `enkit agent`, enkit agent changes its own environment, bash won't see those changes. Next command won't have the variables set.

What I would do:

- get rid of the current -e. Instead, pass `args` from the cobra `RunE` to your `RunAgentCommand`. If no option is passed (eg, no '-p', no '-l'), use `cmd/exec` to run the command specified with `args`. If `args` is empty, spawn `$SHELL`. Use `cmd.Cmd.Env` to set environment variables. Make sure to set `Cmd.Stdin` to `os.Stdin`, same for `Stdout` and `Stderr`. Try as much as possible to avoid poking at process environment variables, as those are global, affect all goroutines, and can have unpredictable effects. I've rarely seen a good reason to use Setenv in go programs. Bonus point if before running the command you print on `/dev/tty` (and it has to be `/dev/tty`) a message like 'Yoda: "Your wisdom must use, enfabrica credentials now using you are"'.
(side note: on posix systems really we should use `syscall.Exec` here, but we can leave that for later.
 
- rename `-p` to `-e`, and change the print statements above to print exactly as `ssh-agent -s` prints the output.
  (have print output the same format as ssh-agent, not a human readable string).

In terms of pseudo code, I'd reorder the operations:
- check for `-l` first. If provided, print list, return.
- check for `-e` next, if provided, print agent details, return.
- no option provided ... either run command, or shell.

proxy/ptunnel/commands/agent.go
```diff
@@ -0,0 +1,65 @@
...
+			bf.Log.Warnf("error setting SSH_AGENT_PID %v", err)
+		}
+	}
+	if config.ListIdentities {
+		principals, err := agent.Principals()
+		if err != nil {
+			return err
+		}
+		for k, v := range principals {
+			fmt.Printf("For Public Key Signature %s, you can SSH as: \n", k)
```
ccontavalli, line 58: I'd be less human here, something like:
"key:%s - users: %s"

(fingerprint is normally used to identify keys, I think it'd be clear, users on the same line to simplify scripting if we have to)

proxy/ptunnel/commands/ssh.go
```diff
@@ -159,6 +159,6 @@ func NewSSH(base *client.BaseFlags) *SSH {
 	root.Command.Flags().StringVar(&root.Subcommand, "tunnel-command", tcommand, "Subcommand to use with the tunnel command. Defaults to empty if the tunnel command does not end with enkit")
 	root.Command.Flags().StringVar(&root.Extra, "tunnel-extra", "", "Extra arguments to pass to the tunnel command")
 	root.Command.Flags().BoolVar(&root.UseInternalAgent, "use-internal-agent", true, "Use the builtin agent that enkit manages")
-
+	root.Command.AddCommand(NewAgentCommand(root.BaseFlags))
```
ccontavalli, line 162: meta: leave as is for now, but I'm not convinced that adding this to `ptunnel` as `agent` subcommand is the right thing to do. 2 concerns:
1) Agent management has to do with authentication, it's weird to see this command in a library that does tunneling.
2) Let's say in my network I have two hosts: one named "agent", the other named "plumber". I can `enkit ssh plumber` just fine (all works) while I'll get a fairly unexpected result if I run `enkit ssh agent`. It'd take me a few minutes to realize that `enkit ssh -- agent` is the correct command, != from `enkit ssh agent`, and different from `enkit ssh agent -- rsync -avz`.

Solutions that come to mind:
a) make agent a  top level command. We have `enkit ssh`, `enkit login`, we can have `enkit agent`. This solves both 1 and 2.
b) make agent entirely flags based. Eg, like you do with `-l` and `-e`, we can just have `-r` to run a command (as proposed above, runs all args with os/exec). So the ssh command itself is modified to have 3 extra flags.

ccontavalli (commented):
I really like this PR, in case it wasn't clear. Usability recommendation below.

lib/kcerts/ssh.go
aaahrens: So I toyed with the idea of ignoring errors here, but I decided that's why I silently ignore public keys that are not certificates. key.Marshall() is "supposed" to return the Public key always, so if it doesn't then there's something funky going on. 

Totally fine to ignore it too since there's no harm in it, but those are my thoughts. I'll put in a continue instead

proxy/ptunnel/commands/agent.go
aaahrens: Did this and then some

proxy/ptunnel/commands/agent.go
aaahrens: Ah see I thought it was bazel forking that was changing it, will remove

aaahrens: If people want to interact with it directly, I'm all for them just copy pasting from the print statement. And we already have a way in code with the GetEnv() function to add it to our in code commands.

lib/kcerts/ssh.go
aaahrens: Implemented, nice.

proxy/ptunnel/commands/ssh.go
aaahrens: So the reason for putting this here is that in my head I know its an ssh agent,  dumb logic of enkit => ssh =>  agent. I suppose that's not as transparent to some people. and you're right that the way the cli is currently designed does use kwargs, which means it should be a terminal command. 

Moving it to top level

proxy/ptunnel/commands/agent.go
ccontavalli: Totally OK doing this in a separate PR, but was personally very excited about supporting execing a command with the env variables all set. It would make debugging and running things like `rsync` much simpler, without overriding the configuration of the current shell.

proxy/ptunnel/commands/agent.go
ccontavalli: don't see it, forgot a git push?

proxy/ptunnel/commands/agent.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/enfabrica/enkit/lib/client"
+	"github.com/enfabrica/enkit/lib/kcerts"
+	"github.com/spf13/cobra"
+	"os"
+	"strconv"
+)
+
+type agentConfig struct {
+	Print          bool
+	SetEnv         bool
```
ccontavalli: set env no longer necessary? well, unless you implement exec.

proxy/ptunnel/commands/ssh.go
ccontavalli: FWIW, yes, putting it under ssh was the logical choice. Had I thought about this a while back, probably I would have made `ssh` have subcommands, like `enkit ssh run`.

lib/kcerts/ssh.go
ccontavalli: +1! You can just say in the doc of the function that Principals returns only certificates that enkit auth recognizes.
Adding a comment there.

lib/kcerts/ssh.go
```diff
@@ -99,6 +99,40 @@ func (a SSHAgent) Valid() bool {
 	return err == nil
 }
 
+type AgentCert struct {
+	MD5        string
+	Principals []string
+	ValidFor   time.Duration
+}
+
+// Principals returns a map where the keys are the CA's PKS and the certs identities are the values
```
ccontavalli, line 108: no map is returned here! `... returns the list of certificates loaded in the agent and supported by enkit`.

ccontavalli (approved):
One thing I'd like fixed/changed, then merge at will:

1) Print in `ssh-agent -s` format, rather than the current english sentences. Following the same format will make this more scriptable, and is equally well readable by humans that know what they're looking for. (your current format is certainly better for humans that don't know what they're reading - but it won't fix the underlying problem).

2) I'd really like the exec support, but can go in a separate PR.

PR #85, opened 2021-04-30
bugfix: allow the generated certs to allocate a pty.

ccontavalli:
Background:
OpenSSH server (and this is an OpenSSH feature - other servers
don't necessarily implement it) allows the certificate to specify
what the user can and cannot do on the remote host.
Eg, can agent forwarding be used? can a pty be allocated?

This PR enables all extensions in ssh.


ccontavalli:
@aaahrens I believe this is the PTY problem. If you run `ssh-keygen -L -f file.cert` on a file generated by ssh tools, you can see the extensions there. No such extension in our certs. See also the document linked in the comment in the PR itself.

However, we need to make this configurable: eg, the caller should tell the kcerts library which extensions to enable. Further, we should remove the hard coded 1 and 2 in the ssh-cert signing code.

Can you take this on? I'm ok leave those two AIs as TODOs.

aaahrens (approved):

PR #86, opened 2021-05-04
Fix SSHD Weirdness on HostKeys

aaahrens:
Right Now, if you ssh encode a ed25519 private key, it is invalid in sshd, but not ssh for some reason. 

Needs investigation, but this works with sshd.

### DIFF
--- a/lib/kcerts/ked25519/cert.go
+++ b/lib/kcerts/ked25519/cert.go
@@ -46,7 +46,6 @@ func MarshalED25519PrivateKey(key ed25519.PrivateKey) []byte {
 	// Add the pubkey to the optionally-encrypted block
 	pk, ok := key.Public().(ed25519.PublicKey)
 	if !ok {
-		//fmt.Fprintln(os.Stderr, "ed25519.PublicKey type assertion failed on an ed25519 public key. This should never ever happen.")
 		return nil
 	}
 	pubKey := []byte(pk)
--- a/lib/kcerts/ked25519/cert.go
+++ b/lib/kcerts/ked25519/cert.go
@@ -1,0 +1,87 @@
+package ked25519ackage
+
+import (
+	"golang.org/x/crypto/ed25519"
+	"golang.org/x/crypto/ssh"
+	"math/rand"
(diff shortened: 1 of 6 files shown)


lib/kcerts/ked25519/cert.go
```diff
@@ -0,0 +1,87 @@
+package ked25519ackage
+
+import (
+	"golang.org/x/crypto/ed25519"
+	"golang.org/x/crypto/ssh"
+	"math/rand"
+)
+// Copied from https://github.com/mikesmitty/edkey/blob/master/edkey.go
+/* Writes ed25519 private keys into the new OpenSSH private key format.
```
ccontavalli, line 9: nit: please format this comment as per normal golang conventions, so godoc generates the documentation properly.

Eg,
```
// MarshalED25519PrivateKey writes ed25519 private keys into the new OpenSSH private key format.
```

lib/kcerts/ked25519/cert.go
```diff
@@ -0,0 +1,87 @@
...
+	pk1.Check1 = ci
+	pk1.Check2 = ci
+
+	// Set our key type
+	pk1.Keytype = ssh.KeyAlgoED25519
+
+	// Add the pubkey to the optionally-encrypted block
+	pk, ok := key.Public().(ed25519.PublicKey)
+	if !ok {
+		//fmt.Fprintln(os.Stderr, "ed25519.PublicKey type assertion failed on an ed25519 public key. This should never ever happen.")
```
ccontavalli, line 47: this function should probably return `([]byte, error)`, this should be a `fmt.Errorf()`

ccontavalli: well, ok, nil is ok, as a single error is possible. We should add to the comment documenting the function that it will either return a byte array, or nil in case of error.

ccontavalli: the // fmt.Println can go away.

ccontavalli (approved):
Fix the comment, and merge.

PR #87, opened 2021-05-04
Adds Enkit  Agent --  <Command> Feature

aaahrens:
Adds in the ability arbitrarily exec command with the agent's environment variables set.

### DIFF
--- a/proxy/ptunnel/commands/agent.go
+++ b/proxy/ptunnel/commands/agent.go
@@ -80,8 +80,7 @@ func RunAgentCommand(command *cobra.Command, bf *client.BaseFlags, args []string
 	}
 	return nil
 }
-const PrintSSHTemplate = `
-SSH_AUTH_SOCK=%s; export SSH_AUTH_SOCK;
+const PrintSSHTemplate = `SSH_AUTH_SOCK=%s; export SSH_AUTH_SOCK;
 SSH_AGENT_PID=%d; export SSH_AGENT_PID;
 echo Agent pid %d;
 `
--- a/proxy/ptunnel/commands/agent.go
+++ b/proxy/ptunnel/commands/agent.go
@@ -80,13 +80,23 @@ func RunAgentCommand(command *cobra.Command, bf *client.BaseFlags, args []string
 	}
 	return nil
 }
-
+const PrintSSHTemplate = `
+SSH_AUTH_SOCK=%s; export SSH_AUTH_SOCK;
+SSH_AGENT_PID=%d; export SSH_AGENT_PID;
+echo Agent pid %d;
+`
 func NewPrintCommand(parent *cobra.Command, bf *client.BaseFlags) *cobra.Command {
 	c := &cobra.Command{
 		Use:   "print",
-		Short: "alias for agent run -- ssh-agent",
+		Short: "Prints out the enkit agent as if you ran ssh-agent -s, compatible with bourne shells",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return RunAgentCommand(parent, bf, []string{"ssh-agent"})
+			agent, err := kcerts.FindSSHAgent(bf.Local, bf.Log)
+			if err != nil {
+				return err
+			}
+			defer agent.Close()
+			fmt.Printf(PrintSSHTemplate, agent.Socket, agent.PID, agent.PID)
+			return nil
 		},
 	}
 	return c
--- a/proxy/ptunnel/commands/agent.go
+++ b/proxy/ptunnel/commands/agent.go
@@ -84,9 +84,9 @@ func RunAgentCommand(command *cobra.Command, bf *client.BaseFlags, args []string
 func NewPrintCommand(parent *cobra.Command, bf *client.BaseFlags) *cobra.Command {
 	c := &cobra.Command{
 		Use:   "print",
-		Short: "alias for agent run -- ssh-add -s",
+		Short: "alias for agent run -- ssh-agent",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return RunAgentCommand(parent, bf, []string{"ssh-agent", "-s"})
+			return RunAgentCommand(parent, bf, []string{"ssh-agent"})
 		},
 	}
 	return c
--- a/proxy/ptunnel/commands/agent.go
+++ b/proxy/ptunnel/commands/agent.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"github.com/enfabrica/enkit/lib/client"
 	"github.com/enfabrica/enkit/lib/kcerts"
+	"github.com/enfabrica/enkit/lib/kflags"
 	"github.com/spf13/cobra"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


proxy/ptunnel/commands/agent.go
```diff
@@ -5,37 +5,43 @@ import (
...
 		},
-		Use:   "agent",
-		Short: "commands for the enkit specific ssh-agent",
+		Use:   "agent -- [Command]",
+		Short: "commands for the enkit specific ssh-agent, anything passed in will execute with SSH_AUTH_SOCK and SSH_AGENT_PID set for the enkti agent.",
 	}
 	// Note the following is intended to be user friendly, identities here are cert principals
 	c.Flags().BoolVarP(&agentConfig.ListIdentities, "list-identities", "l", false, "list the identities loaded current in the agent")
 	c.Flags().BoolVarP(&agentConfig.Print, "print", "p", false, "print the socket and PID of the running agent")
+	c.Flags().BoolVarP(&agentConfig.UseHostEnv, "use-env", "e", false, "Use your terminals env when executing also")
```
ccontavalli, line 33: I would personally remove this option, and always do as if use-env was true. That is generally the expected behavior when forking a command. If a clean environment is needed, one can wrap the command in `/bin/env`

proxy/ptunnel/commands/agent.go
```diff
@@ -46,5 +52,18 @@ func RunAgentCommand(bf *client.BaseFlags, config *agentConfig) error {
 			fmt.Printf("PKS: %s Identities: %v ValidFor: %s \n", p.MD5, p.Principals, p.ValidFor.String())
 		}
 	}
+	if len(config.RemainingArgs) >= 1 {
+		cmd := exec.Command("sh", append([]string{"-c"}, strings.Join(config.RemainingArgs[:], " "))...)
+		cmd.Stdout = command.OutOrStdout()
+		cmd.Stderr = command.ErrOrStderr()
+		cmd.Stdin = command.InOrStdin()
+		cmd.Env = agent.GetEnv()
```
ccontavalli, line 60: as per comment above, would always have:
```
cmd.Env = append(os.Environ(), agent.GetEnv())
```

proxy/ptunnel/commands/agent.go
```diff
@@ -46,5 +52,18 @@ func RunAgentCommand(bf *client.BaseFlags, config *agentConfig) error {
 			fmt.Printf("PKS: %s Identities: %v ValidFor: %s \n", p.MD5, p.Principals, p.ValidFor.String())
 		}
 	}
+	if len(config.RemainingArgs) >= 1 {
```
ccontavalli, line 55: not sure it's worth the effort, but if we do it, better do it now. But if we make `enkit agent` take arbitrary argv to run in a shell, we won't be able to add subcommands, just like it happened with `enkit ssh`.

If we had:
`enkit agent list`
`enkit agent run`
`enkit agent eval`

corresponding to the options you have now, we could make `enkit agent run` automatically run a new shell, and `enkit agent run -- rsync -avz ...` run the specified command.

Thoughts?

ccontavalli (approved):
Fix the env stuff before merging. Up to you if you want to re-factor in subcommands.

proxy/ptunnel/commands/agent.go
```diff
@@ -46,5 +52,18 @@ func RunAgentCommand(bf *client.BaseFlags, config *agentConfig) error {
...
+		cmd := exec.Command("sh", append([]string{"-c"}, strings.Join(config.RemainingArgs[:], " "))...)
+		cmd.Stdout = command.OutOrStdout()
+		cmd.Stderr = command.ErrOrStderr()
+		cmd.Stdin = command.InOrStdin()
+		cmd.Env = agent.GetEnv()
+		if config.UseHostEnv {
+			cmd.Env = append(os.Environ(), cmd.Env...)
+		}
+		if err := cmd.Run(); err != nil {
+			return err
```
ccontavalli, line 65: ah! forgot. When you run a command, `enkit agent` should exit with the status of that command. Eg, if the command fails with exit status 5, `enkit agent command` should also exit with that status.

You can achieve this with something like:
```
    var exit *exec.ExitError
    if errors.As(err, &exit) {
        return kflags.NewStatusError(exit.ExitCode(), err)
    }
```
(code not checked, don't remember if it should be a pointer or not in the var)

aaahrens: Done, and nice catch.

proxy/ptunnel/commands/agent.go
aaahrens: Up to you, I don't have an opinion. Implemented

proxy/ptunnel/commands/agent.go
```diff
@@ -4,47 +4,81 @@ import (
...
-
 	}
-	if config.ListIdentities {
-		principals, err := agent.Principals()
-		if err != nil {
-			return err
-		}
-		for _, p := range principals {
-			fmt.Printf("PKS: %s Identities: %v ValidFor: %s \n", p.MD5, p.Principals, p.ValidFor.String())
+	cmd := exec.Command("sh", append([]string{"-c"}, strings.Join(args[:], " "))...)
```
ccontavalli, line 72:
```
shell := os.Getenv("SHELL")
if shell == "" {
  shell = "/bin/bash"
}

if len(args) > 0 {
  args = append([]string{"-c"}, strings.Join(args, " "))
}
cmd := exec.Command(shell, args...)
```

eg, give an interactive shell if no command is provided, use the same shell the user has.

proxy/ptunnel/commands/agent.go
```diff
@@ -4,47 +4,81 @@ import (
...
+func RunAgentCommand(command *cobra.Command, bf *client.BaseFlags, config *agentConfig, args []string) error {
 	agent, err := kcerts.FindSSHAgent(bf.Local, bf.Log)
 	if err != nil {
 		return err
 	}
 	if config.Print {
 		fmt.Printf("The enkit agent is running at socket %s \n", agent.Socket)
 		fmt.Printf("The enkit agent's pid is %d \n", agent.PID)
-
 	}
```
ccontavalli, line 71: I thought you had changed this to print in the format of `ssh-agent -s`?

```
var format = `SSH_AUTH_SOCK=%s; export SSH_AUTH_SOCK;
SSH_AGENT_PID=%d; export SSH_AGENT_PID;
echo Agent pid %d;
`

fmt.Printf(format, agent.Socket, agent.PID, agent.PID)
```

proxy/ptunnel/commands/agent.go
```diff
@@ -4,47 +4,81 @@ import (
...
 	return c
 }
 
-func RunAgentCommand(bf *client.BaseFlags, config *agentConfig) error {
+func RunAgentCommand(command *cobra.Command, bf *client.BaseFlags, config *agentConfig, args []string) error {
 	agent, err := kcerts.FindSSHAgent(bf.Local, bf.Log)
 	if err != nil {
 		return err
 	}
 	if config.Print {
```
ccontavalli, line 68: Would move this to a different command, as now it is impossible to just print the values and exit? as even an empty args is a valid command. eg, `print` or `config`, so eg one can manually do:
```
eval `enkit agent print`
```
to configure the same agent as enkit.

ccontavalli (approved):

proxy/ptunnel/commands/agent.go
aaahrens: Done

proxy/ptunnel/commands/agent.go
```diff
@@ -4,47 +4,90 @@ import (
...
 	}
 	return nil
 }
+
+func NewPrintCommand(parent *cobra.Command, bf *client.BaseFlags) *cobra.Command {
+	c := &cobra.Command{
+		Use:   "print",
+		Short: "alias for agent run -- ssh-add -s",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return RunAgentCommand(parent, bf, []string{"ssh-agent", "-s"})
```
ccontavalli, line 89: uhm .. this creates a NEW ssh-agent. It does not print out the details of the ssh-agent created by enkit?

aaahrens: Just using print now

proxy/ptunnel/commands/agent.go
```diff
@@ -4,47 +4,100 @@ import (
...
+	cmd.Env = append(os.Environ(), agent.GetEnv()...)
+	if err := cmd.Run(); err != nil {
+		if exit, ok := err.(*exec.ExitError); ok {
+			return kflags.NewStatusError(exit.ExitCode(), err)
 		}
+		return err
 	}
 	return nil
 }
+const PrintSSHTemplate = `
```
ccontavalli, line 83:
```
const PrintSSHTemplate = `SSH_AUTH_SOCK=...
```

no newline at the top? otherwise you'll print an extra empty line.

ccontavalli (approved):

PR #94, opened 2021-05-17
Add NVidia driver build rule

cbrune:
Add a build rule for building the NVidia driver family of kernel
modules.  The NVidia kernel module build does not follow the model of
a stock kernel module build, so create a custom rule for it.

This new rule is inspired by the regular kernel module build rule,
modified to deal with the peculiarities of the NVidia build:

- builds multiple kernel modules from one master Makefile

- provides custom arguments to the NVidia Makefile

- builds a configuration test, that generates headers used by all the
  kernel modules.

With this rule in place I can have a BUILD.bazel file for the NVidia driver
that looks like this, without any modifications to the original NVidia source:
```bazel
load("@enkit//bazel/linux:defs.bzl", "nv_driver")

# To build the nvidia driver modules bazel, you can run:
#
#   bazelisk build :nv_driver
#       Will build the driver modules for all the kernels listed.
#
#   bazelisk build :enf-5.11.16-latest-nv_driver
#       Will build the modules for the specified kernel.
#

nv_driver(
    name = "nv_driver",
    kernels = [
        "enf-5.11.16-latest",
        "ubuntu-default-kernel-tree",
    ],
    modules = [
        "nvidia",
        "nvidia-uvm",
        "nvidia-modeset",
        "nvidia-drm",
    ],
    # To see the default module build output set 'silent' to 'False'.
    # The default is True.
    # silent = False,
    extra = [
        # To see verbose module output pass "NV_VERBOSE=1" as an extra
        # argument.  'silent' must also be set to 'False' for this to
        # work.
        # "NV_VERBOSE=1",
    ],
)

```

### DIFF
--- a/bazel/linux/defs.bzl
+++ b/bazel/linux/defs.bzl
@@ -126,17 +126,17 @@ a kernel_tree on its own is not expected to be hermetic.
     },
 )
 
-KernelModuleInfo = provider(
-    doc = """Maintains the information necessary to represent a compiled kernel module.""",
+KernelModulesInfo = provider(
+    doc = """Maintains the information necessary to represent compiled kernel modules.""",
     fields = {
         "name": "Name of the rule that defined this kernel module.",
         "package": "A string indicating which package this kernel module has been built against. For example, 'centos-kernel-5.3.0-1'.",
-        "module": "File representing the compiled kernel module (.ko).",
+        "modules": "A list of files representing the compiled kernel modules (.ko).",
     },
 )
 
-def _kernel_module(ctx):
-    module = ctx.attr.module
+def _kernel_modules(ctx):
+    modules = ctx.attr.modules
     inputs = ctx.files.srcs + ctx.files.kernel
     srcdir = ctx.file.makefile.dirname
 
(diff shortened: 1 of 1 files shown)


bazel/linux/defs.bzl
```diff
@@ -655,3 +655,181 @@ The test will run locally inside a user-mode linux process.
...
+NvDriverInfo = provider(
+    doc = """Maintains the information necessary to represent the nvidia driver kernel modules.""",
+    fields = {
+        "name": "Name of the rule that defined this driver.",
+        "package": "A string indicating which package this driver has been built against. For example, 'centos-kernel-5.3.0-1'.",
+        "modules": "An array of files representing the compiled kernel modules (.ko).",
+    },
+)
+
+def _nv_driver(ctx):
```
ccontavalli, line 668: both `nv_driver` and `nv_driver_rule` look very similar to `kernel_module` and `kernel_module_rule`. Do we need a dedicated nvidia rule? or can we just generalize the `kernel_module_rule` to eg, take either a list or a single module name? take extra parameters for the makefile generation?

cbrune: I can give it a go

cbrune: I added a new commit that generalizes the kernel module rules.  These existing targets all build fine:

`bazel build //sdk/driver:enf`
`bazel build //sdk/driver:enf_queues_test`
`bazel build //poc/ibuverbs/driver:all`

Along with a new `//poc/nv-fpga/driver:all` target that is WIP.

However, the current patch breaks the kunit test.  I get this now:
```
(default-dev-env)curt@enfabrica:~/src/internal$ bazel build //sdk/driver:enf_queues_unit_test
INFO: Invocation ID: 311c40b1-f2fc-4868-bd17-76e82eea9122
ERROR: /home/curt/src/internal/sdk/driver/BUILD.bazel:42:12: in module attribute of kernel_test rule //sdk/driver:enf_queues_unit_test: '//sdk/driver:enf_queues_test' does not have mandatory providers: 'KernelModulesInfo'
ERROR: Analysis of target '//sdk/driver:enf_queues_unit_test' failed; build aborted: Analysis of target '//sdk/driver:enf_queues_unit_test' failed
INFO: Elapsed time: 2.903s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded, 0 targets configured)
```

cbrune: I understand why the kunit test broke.  I will add commit soon to address it.

bazel/linux/defs.bzl
```diff
@@ -261,37 +329,68 @@ def kernel_module(*args, **kwargs):
...
 
-    module = kwargs.get("module", kwargs["name"])
-    if not module.endswith(".ko"):
-        module = module + ".ko"
-    kwargs["module"] = module
+    if "module" in kwargs:
+        module = kwargs["module"]
+        kwargs.pop("module")
+    else:
+        module = kwargs["name"]
```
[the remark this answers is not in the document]
ccontavalli: or if you want to support both having a single module or a list of modules...
```
kwargs.setdefault("modules", [kwargs.pop("module", kwargs["name"])])
```
although this will silently override modules with module if both are specified.

bazel/linux/defs.bzl
```diff
@@ -261,37 +329,68 @@ def kernel_module(*args, **kwargs):
...
+      kernel: a label, indicating the kernel_tree to build the module against. nv_driver ensures
+            the label starts with an '@', as per bazel convention.
+      kernels: list of kernel (same as above). kernel_module will instantiate multiple
+            kernel_module_rule, one per kernel, and ensure they all build in parallel.
+    """
+
+    if "makefile" not in kwargs:
+        kwargs["makefile"] = "Makefile"
+
+    if "srcs" not in kwargs:
```
ccontavalli, line 385: I was about to suggest to move this "srcs" in a separate function, but then I realized that 1) the list is different from kernel_module, but the rest of the mechanic is pretty much the same as `kernel_module`?

Not sure this is overkill, feel free to do it or not. But suggestion:

a) add a `_kernel_default_srcs(extra_includes = [], extra_excludes = [])` function that pretty much `return native.glob ... include + extra_includes, exclude = BUILD_LEFTOVERS + extra_excludes, ...)`

b) use `_kernel_default_srcs()` both here and above.

ccontavalli: also nit: just remembered that skylark supports setdefault, so you could:

```
kwargs.setdefault("makefile", "Makefile")
kwargs.setdefault("srcs", _kernel_default_srcs(includes = kwargs["makefile"]))
kwargs.setdefault("make_format_str", ...)
```
at your taste.

bazel/linux/defs.bzl
```diff
@@ -239,11 +271,47 @@ def _normalize_kernel(kernel):
...
+    for m in kwargs.get("modules", kwargs["name"]):
+        if not m.endswith(".ko"):
+            m = m + ".ko"
+        modules += [m]
+    kwargs["modules"] = modules
+
+    kernels = []
+    if "kernel" in kwargs:
+        kernels += [kwargs["kernel"]]
+        kwargs.pop("kernel")
```
ccontavalli, line 287: nit: mostly a matter of taste, but [pop](https://docs.bazel.build/versions/master/skylark/lib/dict.html#pop) returns the value that was removed, and return a default. So this could be:

```
kernels = kwargs.pop("kernels", [])
if "kernel in kwargs:
  kernels += [kwargs.pop("kernel")]
```

ccontavalli (approved):
all comments are really stylistic nits below. It's entirely up to you if you want to address them, or keep things as they are.

Only important note: I assume you tested this with current bazel targets, and the new rule is backward compatible (it looks like it is, but I may have missed some details)?

One trick I use for testing is the bazel flag `--override_repository=enkit=/path/to/my/local/enkit`.

cbrune:
@ccontavalli -- ok, I fixed the kunit test problem and addressed some of your nits.  Can you merge it in? I don't have access.

PR #96, opened 2021-05-19
Add automation for enf and UML kernel releases.

rajachan:


rajachan:
Updated to push an image for VMs too.

ccontavalli (approved):
My main comment is that as is, this is very enfabrica specific, for the public enkit repo it'd probably be good to be more generic.

Still, this can serve as a good example, and as discussed, the enkit repo is probably the best place for the script, so merging. Thank you!

PR #99, opened 2021-05-25
Increase license manager timeout to 2 hours

ahungrynacho:
30 minute license queue timeout is way too short.

### DIFF
--- a/bazel/go_repositories.bzl
+++ b/bazel/go_repositories.bzl
@@ -2151,8 +2151,8 @@ def go_repositories():
     go_repository(
         name = "org_golang_google_grpc",
         importpath = "google.golang.org/grpc",
-        sum = "h1:o1bcQ6imQMIOpdrO3SWf2z5RV72WbDwdXuK0MDlc8As=",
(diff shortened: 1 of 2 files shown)


bazel/go_repositories.bzl
```diff
@@ -2151,8 +2151,8 @@ def go_repositories():
     go_repository(
         name = "org_golang_google_grpc",
         importpath = "google.golang.org/grpc",
-        sum = "h1:o1bcQ6imQMIOpdrO3SWf2z5RV72WbDwdXuK0MDlc8As=",
-        version = "v1.36.0",
+        sum = "h1:zWTV+LMdc3kaiJMSTOFz2UgSBgx8RNQoTGiZu3fR9S0=",
```
ccontavalli, line 2154: this seems a spurious change? why do we need to change grpc version to increase the timeout to two hours?

ahungrynacho: arg, let me create a new branch after rebasing from master to clean this up

PR #100, opened 2021-05-25
Improve license manager timeout

ahungrynacho:
- Make timeout default to 7200 seconds on the client side wrapper script in this PR but also as a configurable parameter
https://github.com/enfabrica/internal/pull/692
- Read total license count as an environment variable from within the container so that a new version of enkit does not have to be released.

### DIFF
--- a/manager/client/main.go
+++ b/manager/client/main.go
@@ -26,9 +26,9 @@ func run(timeout time.Duration, cmd string, args ...string) {
 		log.Fatalf("Job failed to complete after %s: %s %s\n", timeout, cmd, strings.Join(args, " "))
 	}
 	if err != nil {
-		log.Fatalf("Job \"%s %v\" failed with error %s", cmd, strings.Join(args, " "), err)
+		log.Fatalf("Job \"%s %s\" failed with error %s", cmd, strings.Join(args, " "), err)
 	}
-	log.Printf("Job completed successfully: %s %v\n", cmd, strings.Join(args, " "))
+	log.Printf("Job completed successfully: %s %s\n", cmd, strings.Join(args, " "))
 }
 
 func polling(client rpc_license.LicenseClient, username string, quantity int32, vendor string, feature string,
--- a/manager/client/main.go
+++ b/manager/client/main.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/exec"
 	"os/user"
+	"strconv"
 	"strings"
 	"time"
 )
@@ -22,7 +23,7 @@ func run(timeout time.Duration, cmd string, args ...string) {
 	job := exec.CommandContext(ctx, cmd, args...)
 	err := job.Run()
 	if ctx.Err() == context.DeadlineExceeded {
-		log.Fatalf("Job failed to complete after %d seconds: %s %s \n", timeout, cmd, strings.Join(args, " "))
+		log.Fatalf("Job failed to complete after %s: %s %s \n", timeout, cmd, strings.Join(args, " "))
 	}
 	if err != nil {
 		log.Fatalf("Job \"%s %s\" failed with error %s", cmd, strings.Join(args, " "), err)
@@ -31,8 +32,8 @@ func run(timeout time.Duration, cmd string, args ...string) {
 }
 
 func polling(client rpc_license.LicenseClient, username string, quantity int32, vendor string, feature string,
-	cmd string, args ...string) {
-	timeout := 1800 * time.Second
+	timeout int, cmd string, args ...string) {
+	maxTimeout := time.Duration(timeout) * time.Second
 	waiting := 0 * time.Second
 	interval := 5 * time.Second
 	stream, err := client.Polling(context.Background())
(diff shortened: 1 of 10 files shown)


manager/client/main.go
```diff
@@ -74,8 +75,8 @@ func polling(client rpc_license.LicenseClient, username string, quantity int32,
 		time.Sleep(interval)
 		waiting += interval
 	}
-	if waiting >= timeout {
-		log.Fatalf("Failed to acquire %d %s feature %s license after %d seconds \n", quantity, vendor, feature, waiting)
+	if waiting >= maxTimeout {
+		log.Fatalf("Failed to acquire %d %s feature %s license after %s \n", quantity, vendor, feature, waiting)
```
ccontavalli, line 79: nit: whitespace before \n

manager/client/main.go
```diff
@@ -87,7 +88,11 @@ func polling(client rpc_license.LicenseClient, username string, quantity int32,
 func main() {
 	var quantity int32 = 1
 	host, port := os.Args[1], os.Args[2]
-	vendor, feature := os.Args[3], os.Args[4]
+	timeout, err := strconv.Atoi(os.Args[3])
```
ccontavalli, line 91: if you do it like this, we will be forced to update enkit and the eda rules in lockstep, as the new client is backward incompatible with the previous one.

Add flag parsing, really. All you need is something like:

```
timeout = flag.Duration("timeout", 20 * time.Minute,  "How long to hold the license for at most")
```
and then add:
```
flag.Parse()
```
of course with an:
```
import "flag"
```
statement.

manager/client/main.go
```diff
@@ -31,8 +32,8 @@ func run(timeout time.Duration, cmd string, args ...string) {
 }
 
 func polling(client rpc_license.LicenseClient, username string, quantity int32, vendor string, feature string,
-	cmd string, args ...string) {
-	timeout := 1800 * time.Second
+	timeout int, cmd string, args ...string) {
```
ccontavalli, line 35: pass timeout as a `time.Duration`, not as an int, and get rid of the conversion below.

manager/common/license.go
```diff
@@ -120,8 +122,17 @@ type License struct {
 	Used  int
 }
 
-var totalCadenceLic = 3
-var totalXilinxLic = 3
+func strToInt(num string) int {
```
ccontavalli, line 125: You have two entirely orthogonal features in one PR: 1) to make the client timeout configurable, 2) to change the server. Create two separate PRs.

manager/client/main.go
```diff
@@ -22,7 +23,7 @@ func run(timeout time.Duration, cmd string, args ...string) {
 	job := exec.CommandContext(ctx, cmd, args...)
 	err := job.Run()
 	if ctx.Err() == context.DeadlineExceeded {
-		log.Fatalf("Job failed to complete after %d seconds: %s %s \n", timeout, cmd, strings.Join(args, " "))
+		log.Fatalf("Job failed to complete after %s: %s %s \n", timeout, cmd, strings.Join(args, " "))
```
ccontavalli, line 26: nit: if you use `%v`, you can just leave `args` without the strings.Join()

manager/common/license.go
```diff
@@ -120,8 +122,17 @@ type License struct {
...
+	result, err := strconv.Atoi(num)
+	if err != nil {
+		log.Fatalf("Failed to convert %s to int", num)
+	}
+	return result
+}
+
+// Get total number of licenses from container env var
+var totalCadenceLic = strToInt(os.Getenv("CADENCE_LICENSES"))
+var totalXilinxLic = strToInt(os.Getenv("XILINX_LICENSES"))
```
ccontavalli, line 135: so... to fix the immediate issue, we just need to increase `totalXilinxLic` and `totalCadenceLic`, and restart the server. I recommend you do that, so we can move forward.

In terms of merging this code in the repository: I would strongly prefer not to have environment variables as the default config mechanism. Either command line flags, or a simple json / yaml config file.

The json/yaml file could load pretty much the entire content of LicenseCounter in one go. It would be pretty simple to do:
```
type LicenseConfig struct {
  Licenses map[string]*License
}
```
and then:
```
var config LicenseConfig
err := marshal.UnmarshalFile("path/to/config/file", &config)
```
with marshal library being:
`import "github.com/enfabrica/enkit/lib/config/marshal"`

ccontavalli (changes requested):

manager/client/main.go
ahungrynacho: It still prints with brackets 
```
vivado [-mode batch -source fpga/projects/lupa/ip/generate_ip_bazel.tcl -tclargs xcvu9p-fsgd2104-2L-e lupa bazel-out/k8-fastbuild/bin/fpga/projects/lupa/ip/vcu1525/ip_axis_switch_S8xM1 fpga/projects/lupa/ip fpga/projects/lupa/ip/common/ip_axis_switch_S8xM1.tcl -notrace]
```
which I do not want.

ccontavalli (approved):

PR #107, opened 2021-05-28
Adds the ability to support hostnames in the proxy

aaahrens:
Right now in /proxy, the host query parameter can only support ips. This adds the ability to use hostnames. 

Also establishes a way to test with httptest package.

### DIFF
--- a/proxy/nasshp/nassh.go
+++ b/proxy/nasshp/nassh.go
@@ -128,17 +128,6 @@ func WithRelayHost(relayHost string) Modifier {
 	}
 }
 
-type Verdict int
-
-const (
-	// Can't decide either way. This is useful for chaning filters.
-	VerdictUnknown Verdict = iota
-	// Let the request in.
-	VerdictAllow
-	// Block the request.
-	VerdictDrop
-)
-
 type Filter func(proto string, hostport string, creds *oauth.CredentialsCookie) Verdict
 
 func WithFilter(filter Filter) Modifier {
@@ -267,14 +256,12 @@ func (np *NasshProxy) ServeProxy(w http.ResponseWriter, r *http.Request) {
 	params := r.URL.Query()
 	host := params.Get("host")
 	port := params.Get("port")
-
 	origin := r.Header.Get("Origin")
 	if origin != "" && OriginMatcher.MatchString(origin) {
 		w.Header().Add("Vary", "Origin")
 		w.Header().Add("Access-Control-Allow-Credentials", "true")
 		w.Header().Add("Access-Control-Allow-Origin", origin)
 	}
-
 	if np.authenticator != nil {
 		creds, err := np.authenticator(w, r, nil)
 		if err != nil {
@@ -338,9 +325,14 @@ func (np *NasshProxy) allow(r *http.Request, w http.ResponseWriter, sid, hostpor
 		}
 		logid = LogId(sid, r, hostport, creds)
 	}
-
 	if np.filter != nil {
 		verdict := np.filter("tcp", hostport, creds)
+		if HostIp(hostport).IsUrl() {
+			for _, u := range HostIp(hostport).Resolve() {
+				// TODO(adam): make verdict merging configurable from ACL list
+				verdict = verdict.MergePreferAllow(np.filter("tcp", u, creds))
+			}
+		}
 		if verdict != VerdictAllow {
 			np.log.Infof("%s was rejected by filter", logid)
 			http.Error(w, fmt.Sprintf("Go somewhere else, you are not allowed to connect here."), http.StatusUnauthorized)
--- a/proxy/nasshp/types.go
+++ b/proxy/nasshp/types.go
@@ -1,0 +1,51 @@
+package nasshp
+
+import (
+	"net"
+	"strings"
+)
+
+type HostIp string
+
+func (hi HostIp) IsUrl() bool {
+	res := strings.Split(string(hi), ":")
+	if len(res) != 2 {
+		return false
+	}
+	return net.ParseIP(res[0]) == nil
+}
+
+func (hi HostIp) Resolve() []string {
+	res := strings.Split(string(hi), ":")
+	if len(res) != 2 {
+		return []string{}
+	}
+	p := res[1]
+	ips, err := net.LookupHost(res[0])
+	if err != nil {
+		return []string{}
+	}
+	var toReturn []string
+	for _, ip := range ips {
+		toReturn = append(toReturn, strings.Join([]string{ip, ":", p}, ""))
+	}
+	return toReturn
+}
+
+type Verdict int
+
+const (
+	// Can't decide either way. This is useful for chaning filters.
+	VerdictUnknown Verdict = iota
+	// Let the request in.
+	VerdictAllow
+	// Block the request.
+	VerdictDrop
+)
+
+func (v Verdict) MergePreferAllow(vv Verdict) Verdict {
+	if v == VerdictAllow || vv == VerdictAllow {
+		return VerdictAllow
+	}
+	return VerdictDrop
+}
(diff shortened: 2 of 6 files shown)


proxy/nasshp/types.go
```diff
@@ -0,0 +1,51 @@
...
+const (
+	// Can't decide either way. This is useful for chaning filters.
+	VerdictUnknown Verdict = iota
+	// Let the request in.
+	VerdictAllow
+	// Block the request.
+	VerdictDrop
+)
+
+func (v Verdict) MergePreferAllow(vv Verdict) Verdict {
```
ccontavalli, line 46: So, there are different ways to Merge multiple verdicts? This will return allow as long as one verdict is allow, correct? eg, if either verdict is allow, then allow. 

Was that the intent? Or was the intent to allow only if all are Allow or Unknown?

I would rename the function.

Also.. there is no need to test all resolutions, if at the first one we find allowed, it's guaranteed to be allowed?

proxy/ptunnel/tunnel_test.go
```diff
@@ -129,3 +135,73 @@ func TestBasic(t *testing.T) {
...
+		u.RawQuery = url.Values{"host": {host}, "port": {fmt.Sprintf("%d", port)}}.Encode()
+		fmt.Println("uri is ", u.String())
+		responseString := ""
+		err = protocol.Get(u.String(), protocol.Read(protocol.String(&responseString)))
+		if shouldFail {
+			assert.NotNil(t, err)
+			return
+		}
+		assert.Nil(t, err)
+		//// TODO(adam): make some nice SID utility functions, right now this just checks that the sid is sent
```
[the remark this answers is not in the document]
aaahrens: Yup

proxy/nasshp/types.go
aaahrens: I removed this, it was going to be a precursor to OPA, but it isn't needed

proxy/nasshp/types.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"net"
+	"strings"
+)
+
+type HostIp string
+
+func (hi HostIp) IsUrl() bool {
+	res := strings.Split(string(hi), ":")
```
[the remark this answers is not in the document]
aaahrens: Fixed, used a mediocre label jump

proxy/nasshp/nassh.go
```diff
@@ -326,18 +326,26 @@ func (np *NasshProxy) allow(r *http.Request, w http.ResponseWriter, sid, hostpor
 		logid = LogId(sid, r, hostport, creds)
 	}
 	if np.filter != nil {
-		verdict := np.filter("tcp", hostport, creds)
-		if HostIp(hostport).IsUrl() {
-			for _, u := range HostIp(hostport).Resolve() {
+		host, port, err := net.SplitHostPort(hostport)
```
ccontavalli, line 329: Can you change my Sprintf() above (the one that creates `hostport`) to use `net.JoinHostPort`? So this is guaranteed to work correctly, even in the presence of IPv6.

proxy/nasshp/nassh.go
```diff
@@ -326,18 +326,26 @@ func (np *NasshProxy) allow(r *http.Request, w http.ResponseWriter, sid, hostpor
...
+		if err != nil {
+			goto DENY
```
ccontavalli, line 331: I would avoid the goto, and differentiate the error messages? So if we ever land here, we can tell what the problem is? Also, we avoid the goto, considered harmful and evil (except in C code for cleanup).

```
if err != nil {
  np.log.Infof("%s could not split %s in filter validation", logid, hostport)
  http.Error(w, fmt.Sprintf("Something fishy in your request, cannot apply ACLs, rejected)
}

File: proxy/nasshp/nassh.go [hunk 6]
```diff
@@ -326,18 +326,26 @@ func (np *NasshProxy) allow(r *http.Request, w http.ResponseWriter, sid, hostpor
...
+		}
+		{
```
ccontavalli, line 333: why this extra block? unnecessary?

proxy/nasshp/nassh.go
```diff
@@ -326,18 +326,26 @@ func (np *NasshProxy) allow(r *http.Request, w http.ResponseWriter, sid, hostpor
...
+			res, err := net.LookupHost(host)
+			if err != nil {
+				goto DENY
```
ccontavalli, line 336: same comment as above. I would avoid the goto, in favor of a message that indicates what the problem is.
So eg, if DNS is broken, we notice.

(and it's possible for it to be broken: I had to ensure the proxy container used our hacked dns server for resolution, it was pointing to 8.8.8.8 originally)

proxy/nasshp/nassh.go
```diff
@@ -326,18 +326,26 @@ func (np *NasshProxy) allow(r *http.Request, w http.ResponseWriter, sid, hostpor
...
+			}
+			for _, u := range res {
 				// TODO(adam): make verdict merging configurable from ACL list
-				verdict = verdict.MergePreferAllow(np.filter("tcp", u, creds))
+				if verdict := np.filter("tcp", net.JoinHostPort(u, port), creds); verdict == VerdictAllow {
```
ccontavalli, line 340: problem: when the proxy connects below, DNS will pick one of the hosts at random. So... our ACLs should really check that all are allowed. Otherwise, one may create a DNS record that resolves to 4 names, 3 of which are ok, 1 of which is evil, and have 75% probability of getting through the ACLs, and 25% past that to actually be able to connect to the evil host, so... if he makes ~6 attempts, he'll have a reasonable probability one of them will succeed?

aaahrens: Changed the ACL to only accept if all are allowed

proxy/nasshp/nassh.go
aaahrens: Done

proxy/nasshp/nassh.go
```diff
@@ -338,14 +324,31 @@ func (np *NasshProxy) allow(r *http.Request, w http.ResponseWriter, sid, hostpor
...
 			http.Error(w, fmt.Sprintf("Go somewhere else, you are not allowed to connect here."), http.StatusUnauthorized)
 			return logid, false
 		}
+		res, err := net.LookupHost(host)
+		if err != nil {
+			np.log.Warnf("%v err looking up host %s", err, host)
+			http.Error(w, fmt.Sprintf("Go somewhere else, you are not allowed to connect here."), http.StatusUnauthorized)
+			return logid, false
+		}
+		verdict := VerdictAllow
```
ccontavalli, line 340: initialize verdict to `VerdictUnknown`, see comment below.

proxy/nasshp/types.go
```diff
@@ -0,0 +1,19 @@
...
+	// Can't decide either way. This is useful for chaning filters.
+	VerdictUnknown Verdict = iota
+	// Let the request in.
+	VerdictAllow
+	// Block the request.
+	VerdictDrop
+)
+
+func (v Verdict) MergeOnlyAcceptAllow(vv Verdict) Verdict {
+	if v != VerdictAllow || vv != VerdictAllow {
```
ccontavalli, line 15: `VerdictUnknown` should be "non voting".

Given the set of checks to perform, I would succeed only if there is at least one VerdictAllow, and no VerdictDrop.

In terms of logic, that would mean

1)  initializing verdict to Unknown (see comment above)
2) changing MergeOnlyAcceptAllow like below
3) checking that the result is VerdictAllow, like you are doing already

MergeOnlyAcceptAllow would look something like:
```
switch vv {
case VerdictDrop: return VerdictDrop // no doubt, we have to drop.
case VerdictUnknown: return v // whatever we decided before, is still valid.
case VerdictAllow:
    if v == VerdictUnknown {
      return VerdictAllow
    }
    return v // if there was a previous Drop determination, let's stick with that. If it's not Drop, it's Allow already.
}
```

ccontavalli (approved):

PR #117, opened 2021-06-04
Basic Working Setup for React projects

aaahrens:
This does the following:
- adds in rules to merge package.jsons. The base is ui/package.json
- adds in rules for copying files and cutting short directories
- start, build and tests work. Code coverage also works, but we need to figure out where to upload it before i enable it. 
- Upgrades rules_nodejs to the latest version. 
- IBazel works. 

TODOS:
- needs work for incorporating proto files. Should be in the next PR
- SOP in ui/README.md for getting started
- Vendor jq for merging

Suggestions:
- Performance suggestions would be nice, not sure how efficient my rules are. 
- Code organization


ui/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
...
+# Note this patch (May 25th,
+filegroup(
+    name = "git-patches",
+    srcs = glob(["patches/*"]),
+    visibility = ["//visibility:public"],
+)
+
+react_project(
+    name = "ptunnel",
+    srcs = "//ui/ptunnel:srcs",
```
ccontavalli, line 17: for `srcs`, `public`, `git-patches` and .. `ui-extras`, I would recommend not hard coding the copy file, rather have something like:

```
react_project(
  name = "ptunnel",
  elements = {
    "patches": [":git-patches"],
    "public": ["//ui/ptunnel:public"],
    "src": ["//ui/putnnel:srcs"],
  }
)
```

so in the macro implementation you can use a for loop, while allowing arbitrary paths to be brought into the rule?
Or you may just want to structure the code this way in the macro: first build the dict, and then do a loop.

ccontavalli (changes requested):
wow, I'm certainly impressed. A few comments below, mostly on 1) docstrings, 2) make the attributes stricter, 3) turn the copies into loop if we can, ...

ui/patches/jest-haste-map+26.6.2.patch
```diff
@@ -0,0 +1,21 @@
+diff --git a/node_modules/jest-haste-map/build/crawlers/node.js b/node_modules/jest-haste-map/build/crawlers/node.js
```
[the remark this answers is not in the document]
aaahrens: The issue is that react-scripts hasn't been updated to use the newer version. Once they update react script to use jest 27.0.0 this shouldn't be an issue.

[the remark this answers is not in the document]
aaahrens: React scripts vendors its jest library (which is why you can see in the package it only uses @types for jest)

ui/BUILD.bazel
aaahrens: I agree with making it self contained, but I disagree with the dictionary. I would prefer for named inputs, considering there's no way to run a modern project without them, and for easy of editor use. For optional configs sure, but I view the current inputs as required.

ui/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
+load("//bazel/ui:react.bzl", "react_project")
+
+exports_files([
+    "package.json",
+    "merge.js",
+])
+
+# Note this patch (May 25th,
```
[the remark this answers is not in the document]
aaahrens: completed

bazel/ui/react.bzl
```diff
@@ -0,0 +1,203 @@
...
+            out_path = paths.join(ctx.attr.base_dir, f.basename)
+        else:
+            out_path = trim_path(f.short_path, ctx.attr.prefix, ctx.attr.base_dir)
+        out = ctx.actions.declare_file(out_path)
+        all_outputs += [out]
+        ctx.actions.run_shell(
+            outputs = [out],
+            inputs = depset([f]),
+            arguments = [f.path, out.path],
+            command = "cp $1 $2",
```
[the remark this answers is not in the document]
aaahrens: symlink seems to have no issue

bazel/utils/files.bzl
```diff
@@ -0,0 +1,90 @@
+load("@bazel_skylib//lib:paths.bzl", "paths")
+
+""" Changes a path at a specified keyword to another path, appending the remaining onto it
+Examples:
+  >>> trim_path("/tree/foo/bar", "foo", "/etc")
+  /etc/bar
+"""
```
ccontavalli, line 7: this docstring is documenting a trim_path, that does not exist in this file? Did you intend to rename `rebase_path`?

If this is the docstring for rebase_path, it should be tied to the `rebase_path` function, not at file level.

bazel/utils/files.bzl
```diff
@@ -0,0 +1,90 @@
...
+        )
+
+    return [
+        DefaultInfo(
+            files = depset(all_outputs),
+            runfiles = ctx.runfiles(files = all_outputs),
+        ),
+    ]
+
+"""Copies all files in a directory to another directory, rebasing to a prefix as it goes along. If you wish to copy straight
```
ccontavalli, line 46:
```
"""One line description (period).

Longer description longer longer longer.
Longer longer.

Examples:
    >>> code
    ["/base/src/bar.out", "/base/src/curr/bar.out"]
"""
```

Note the period, newline, example has `>>>` in the code, nothing (just alignment) in the result, which must be a valid python type (" and [ use)

ui/BUILD.bazel
```diff
@@ -0,0 +1,36 @@
+load("//bazel/ui:react.bzl", "react_project")
+
+exports_files([
+    "package.json",
+    "merge.js",
+])
+
+# Note this patch (May 25th, 2021) has landed in jest 27. React scripts vendors jest, so at a later release of react-scripts
+# the patches won't be necessary.
```
ccontavalli, line 9: good comment!

bazel/utils/files.bzl
```diff
@@ -0,0 +1,90 @@
...
+            flag = True
+        if flag:
+            tret.append(m)
+    return paths.join(base, *tret)
+
+def _rebase_and_copy_files_impl(ctx):
+    all_input_files = [
+        f
+        for t in ctx.attr.source_files
+        for f in t.files.to_list()
```
ccontavalli, line 24: ping? `ctx.files.source_files`?

ccontavalli (approved):
All nits left, please address, and feel free to merge.

PR #120, opened 2021-06-07
[Proposal] Add OPA as RBAC Engine

aaahrens:
Reasons: 

- We need an RBAC system to deal with groups, ABAC and possibly offline embedded permissions with certificates.
- We can either embed the OPA PAM library in machinist or write it ourselves, but this lets us use offine RBAC bare metal as well. 
- Homegrowing RBAC is bad. 

Pros: 
- The ability to change RBAC rules without writing code or dealing with regressions. Also the ability to update over the air at a   later date. 
- Hugely Battle tested and e2e tested. 
- Fully embeddable
- OPA is much faster and more declarative then if we homegrow it ourselves. It also doesn't lock us into a set of rules or automagical configs.

Cons:
- You have to learn rego. And its OPA's huge API.
- Moderately overkill for now, but not something I want to skimp out on and build huge technical debt on.
- Our design philosophy is a little overkill (referring to bazel and ptunnel) so it's inline with enkit's thinking. 

What this PR is:

A simple dummy RBAC service, that uses custom plugins, rules and runs fully embedded. It should be pretty easy to see how it would grow in production.

I would recommend looking at the service_test first.


lib/authz/plugins/resources.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	"fmt"
+	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/open-policy-agent/opa/types"
+)
+
+type Role string
+
+var (
```
ccontavalli, line 12: those are constants, right?

```
const (
  RoleUser Role = "user"
  RoleAdmin = "admin"
)
```

lib/authz/types.go
```diff
@@ -0,0 +1,18 @@
+package authz
+
+// In this example, a resource is either an ip address or it is a uuid
+type Resource string
+
+type Action string
+
+// Basic Crud
+var (
```
ccontavalli, line 9: same as other comment? can you use const?

lib/authz/plugins/resources.go
```diff
@@ -0,0 +1,103 @@
...
+func (s DummyService) RoleOnResource(username, resource string) (string, error) {
+	if s.findUser(username).Owns == resource {
+		return string(RoleAdmin), nil
+	}
+	return string(RoleUser), nil
+}
+
+func (s DummyService) findUser(username string) DummyUser {
+	for _, v := range s.Users {
+		if v.Name == username {
```
ccontavalli, line 44: use a map instead of a vector for s.Users?

lib/authz/plugins/resources.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+}
+
+func ResourceRole(d *DummyService) (*rego.Function, rego.Builtin2) {
+	return &rego.Function{
+			Name:    "resource.role",
+			Memoize: true,
+			Decl:    types.NewFunction(types.Args(types.S, types.S), types.S),
+		}, func(bctx rego.BuiltinContext, op1 *ast.Term, op2 *ast.Term) (*ast.Term, error) {
+			fmt.Println("here")
```
ccontavalli, line 76: here?

lib/authz/service.go
```diff
@@ -0,0 +1,79 @@
...
+	return errors.New("undefined response given")
+}
+
+//this is a compile time assertion.
+
+func NewService(d *plugins.DummyService) (*AuthService, error) {
+	rego.RegisterBuiltin2(plugins.ResourceOwn(d))
+	rego.RegisterBuiltin1(plugins.UserRole(d))
+	rego.RegisterBuiltin2(plugins.ResourceRole(d))
+	c := ast.MustCompileModules(opa_files.OpaFiles)
```
ccontavalli, line 68: Use `CompileModules`, return error if compiling fails? library.

lib/authz/service.go
```diff
@@ -0,0 +1,79 @@
...
+		Resource: req.Resource,
+		Action:   req.Action,
+	}
+	res, err := as.q.Eval(ctx, rego.EvalInput(i))
+	if err != nil {
+		return err
+	}
+	fmt.Println(res)
+	if len(res) == 0 {
+		return ErrorOnlyAccessSelfOwned
```
ccontavalli, line 45: pretty cryptic? If it wasn't for the "permission denied to perform this action" string I would not know what this was. Better name for the error?

lib/authz/service.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	if m, ok := res[0].Bindings["x"].(map[string]interface{}); ok {
+		result := m["allow"]
+		if canDo, ok := result.(bool); ok {
+			if canDo {
+				return nil
+			} else {
+				fmt.Println(i.JSON(), res)
```
ccontavalli, line 54: println?

lib/authz/service.go
```diff
@@ -0,0 +1,79 @@
...
+	res, err := as.q.Eval(ctx, rego.EvalInput(i))
+	if err != nil {
+		return err
+	}
+	fmt.Println(res)
+	if len(res) == 0 {
+		return ErrorOnlyAccessSelfOwned
+	}
+
+	if m, ok := res[0].Bindings["x"].(map[string]interface{}); ok {
```
ccontavalli, line 48: I would invert this if for readability:
```
if m, ok := res[0].Bindings["x"].(map[string]interface{}); !ok {
  return errors.New("wrong response type given")
}

canDo, ok := m["allow"].(bool)
if !ok {
  return errors.new("no allow statement was provided")
}

if canDo {
  return nil
}

return ErrorOnlyAccessSelfOwned
```

ccontavalli: BTW, this is pretty obscure... what is "x"? why res[0]? It seems to be very error prone fragile, which is a bit concerning for security code.

ccontavalli: is there a res[1] and res[2] to parse? do we ignore them? Can you add a comment explaining the logic/response expected? are there other keys but x we need to care about?

lib/authz/rbac/hello.rego
```diff
@@ -0,0 +1,29 @@
...
+    resource.role(input.username, input.resource) == admin_roles[i]
+}
+
+allow {
+    resource.own(input.username, input.resource)
+}
+
+allow {
+    resource.own(input.username, input.resource)
+}
```
ccontavalli, line 21: why do we need two allow statements with exactly the same content?

lib/authz/rbac/hello.rego
```diff
@@ -0,0 +1,29 @@
...
+
+default allow = false
+
+allow {
+    user.root(input.username)
+}
+# Admins can performa
+allow {
+    admin_roles := ["admin"]
+    some i
```
ccontavalli, line 11: what is `some i`?

ccontavalli: why not just:

```
resource.role(input.username, input.resource) == "admin"
```

here?

lib/authz/plugins/resources.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+type Role string
+
+var (
+	RoleUser  = Role("user")
+	RoleAdmin = Role("admin")
+)
+
+type DummyUser struct {
```
ccontavalli, line 17: nit: rather than Dummy, name it Simple? or ... not sure?

What does this service do compared to others? comment?

lib/authz/plugins/resources.go
```diff
@@ -0,0 +1,103 @@
...
+	Name string
+	Owns string
```
ccontavalli, line 19: So, the model that I had in mind is that:

a) A user is part of a set of groups. Example: carlo is part of astore-users, astore-admins, docker-pushers, system-builders, ...
b) Groups give various privileges on services. Eg, astore-user allows to read/write/list/... artifacts in astore, astore-admins can access the astore secret key to build it, or gcloud role to take job up/down, ...
c) groups (or names) can then be used by individual services. Eg, directory '/eda/' in astore is writeable by members of the group build-tool-maintainers, readable by everyone.

Given what I just typed above...
1) Shouldn't user have multiple groups? Eg, for each user, we have a set of groups he's part of.
2) Shouldn't the rego file be built around those groups?

PR #122, opened 2021-06-09
proxy/ptunnel: fix race condition in showing error messages.

ccontavalli:
Background:
With v17 of enkit tunnel, we noticed that the login error instructing
users to re-login in case of authentication error was not always showing.
Some times it was showing, some times it was not - for the same error.

We tracked it down to a race condition on shutdown, introduced in v17:
- v17 added code to properly shutdown the proxied connections.
  This is key for using tunnels in listening mode.
- as a result of that change, a normal tunnel used by ssh also started
  propagating connection closures correctly.

Now:
- when the ssh client uses a ProxyCommand, if the command closes
  the file descriptor, ssh will kill the command.
- this introduced the problem:
  1) Proxy closes connection.
  2) Tunnel command notices, immediately closes file descriptor for ssh.
  RACE CONDITION: ssh may kill the tunnel before or after it is able
  to print the error message. If ssh is fast, no error message is printed.
  If ssh is slow, then the error message is printed.

In this PR:
- Changed the code so when running in normal tunnel mode, we don't close
  the write descriptor if the proxy closes the connection. The error
  is still detected by the tunnel command, which will print the correct
  message and exit, but won't give reasons to ssh to kill the command
  before then.

### DIFF
--- a/lib/knetwork/nopclose.go
+++ b/lib/knetwork/nopclose.go
@@ -1,0 +1,35 @@
+package knetwork
+
+import (
+	"io"
+)
+
+type nopReadCloser struct {
+	io.Reader
+}
+
+func (n *nopReadCloser) Close() error {
+	return nil
+}
+
+// NopReadCloser turns any io.Reader into a io.ReadCloser with a nop closer.
+//
+// This can also be used to ignore Close() events on another ReadCloser.
+func NopReadCloser(reader io.Reader) io.ReadCloser {
+	return &nopReadCloser{reader}
+}
+
+type nopWriteCloser struct {
+	io.Writer
+}
+
+func (n *nopWriteCloser) Close() error {
+	return nil
+}
+
+// NopWriteCloser turns any io.Writer into a io.WriteCloser with a nop closer.
+//
+// This can also be used to ignore Close() events on another WriteCloser.
+func NopWriteCloser(writer io.Writer) io.WriteCloser {
+	return &nopWriteCloser{writer}
+}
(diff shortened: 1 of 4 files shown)


lib/knetwork/nopclose.go
```diff
@@ -0,0 +1,35 @@
+package knetwork
+
+import (
+	"io"
+)
+
+type nopReadCloser struct {
```
aaahrens, line 7: This already exists in the io package? https://golang.org/src/io/io.go?s=6022:6080#L612

lib/knetwork/nopclose.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// NopReadCloser turns any io.Reader into a io.ReadCloser with a nop closer.
+//
+// This can also be used to ignore Close() events on another ReadCloser.
+func NopReadCloser(reader io.Reader) io.ReadCloser {
+	return &nopReadCloser{reader}
+}
+
+type nopWriteCloser struct {
```
aaahrens, line 22: So I don't like this. It scares me and NoOp writers scare me. It has it's valid uses, but I would rather see the workarounds in the proposal thread that wanted this feature.

https://github.com/golang/go/issues/22823

aaahrens (changes requested):

lib/knetwork/nopclose.go
aaahrens: Otherwise, yolo.

lib/knetwork/nopclose.go
ccontavalli: removed, good find.

PR #126, opened 2021-06-28
Update go and rules_go

aaahrens:
In this PR
- Updates rules_go to the latest version
- Updates the go version to 1.15


ccontavalli (commented):
Also need to update: https://github.com/enfabrica/enkit/blob/master/astore/server/deploy/app.yaml?

(and any other app.yaml that may be in the repository? Otherwise we'll be building and testing on one golang version, and pushing to the cloud on another)

ccontavalli (approved):

PR #136, opened 2021-07-13
bazel/init.bzl: initialize golang sdk in one place only.

ccontavalli:
Before this change:
- WORKSPACE initializes the go toolchain
- bazel/init.bzl initializes the go toolchain
- the first initialization wins (WORKSPACE), so all works in enkit/

- another repository (internal) uses @enkit/baze/init.bzl
- there is no WORKSPACE initialization

---> CRASH, as the initialization in bazel/init.bzl is invalid with
the newer versions of the toolchain.

In this PR:
- initialize the golang toolchain from one place only,
  bazel/init.bzl, with the correct version.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -9,11 +9,6 @@ load("//bazel:deps.bzl", "enkit_deps")
 
 enkit_deps()
 
-load("@io_bazel_rules_go//go:deps.bzl", "go_download_sdk")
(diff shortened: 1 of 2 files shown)


ccontavalli:
@rajachan this fixes the problem you were seeing.

rajachan (approved):
Tested the fix, works for me 👍

PR #139, opened 2021-07-28
[BugFix] Fixes SSH Agent Private key and Cert TTL

aaahrens:
- When ssh keys are added to the agent, the time added to the agent matches the ttl of the cert
- Fixed return type of ed25519.Raw() provide


lib/kcerts/certs.go
```diff
@@ -201,3 +202,12 @@ func NewOptions(mods ...Modifier) (*certOptions, error) {
 	}
 	return co, nil
 }
+
+// SSHCertTTL returns the remaining ttl of a cert, either when compared to current time, or it's total expiry.
+func SSHCertTTL(cert *ssh.Certificate, useCurrentTime bool) time.Duration {
```
ccontavalli, line 207: this function was very confusing for me to read :/ two notes:
- you assign `before` from Valid*After*
- and then subtract `before` from Valid*Before*

1) use different names for the variables?

2) Also, there's really two TTLs:
- total length of certificate life
- how longer the certificate has left to live.

Rather than a flag useCurrentTime, I would personally have two functions. Eg:

```
func SSHCertTotalLife(cert *ssh.Certificate) {
}

func SSHCertRemainingLife(cert *ssh.Certificate) {
}
```
3) Looking at: https://cs.opensource.google/go/x/crypto/+/a769d52b:ssh/certs.go;l=404

The ValidAfter and ValidBefore could also be negative. Given impact of computing those values and TTLs incorrectly, can you add more defense in depth?

Eg, what if ValidBefore is 0 or negative? What if ValidAfter is 0 or negative?

lib/kauth/save.go
```diff
@@ -34,7 +33,8 @@ func SaveCredentials(credentials *EnkitCredentials, store cache.Store, l logger.
 		return err
 	}
 	defer agent.Close()
-	if err = agent.AddCertificates(credentials.PrivateKey, credentials.SSHCertificate, uint32((time.Hour * 48).Seconds())); err != nil {
+
+	if err = agent.AddCertificates(credentials.PrivateKey, credentials.SSHCertificate); err != nil {
```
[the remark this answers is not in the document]
aaahrens: sure

lib/kcerts/certs.go
aaahrens: I agree about the two different functions, implemented.

I would add the extra layer of defense in agent.AddCertificate (since detecting TTL is fundamentally an error less function).

Refusing if it's negative is ok, adding it in.

lib/kcerts/certs.go
```diff
@@ -201,3 +202,13 @@ func NewOptions(mods ...Modifier) (*certOptions, error) {
 	}
 	return co, nil
 }
+
+// SSHCertTTL returns the remaining ttl of a cert, either when compared to current time, or it's total expiry.
```
ccontavalli, line 206: comment is now wrong, update?

lib/kcerts/ssh.go
```diff
@@ -148,10 +149,14 @@ func (a SSHAgent) AddCertificates(privateKey PrivateKey, publicKey ssh.PublicKey
 		return fmt.Errorf("public key is not a valid ssh certificate")
 	}
 	agentClient := agent.NewClient(conn)
+	ttl := SSHCertRemainingTTL(cert).Seconds()
+	if math.Signbit(ttl) {
```
ccontavalli, line 153: uhm .. let's say the ValidBefore is -2**63, if I subtract the current time, it'll underflow and SSHCertRemainigTTL may return a positive value?

I'd personally be much more comfortable with SSHCertRemainingTTL and friend just erroring out directly if the value in the cert is invalid.

Eg,
```
func SSHCertRemainingTTL(...) (time.Time, error) {
  if ...
}
```

aaahrens: So it cant underflow, the Wall of time is casted to a uint64.

`	
ts := time.Unix(int64(math.Pow(2, 64)), 0 )
fmt.Println("printing seconds")
fmt.Println("ts is ", ts.String())
fmt.Println(ts.Sub(time.Unix(0, `0)).Seconds())
`
something like this still returns negative
and arbitrarily adding errors to something like ttl I don't like, because the caller decides whether or not it likes the TTL

lib/kcerts/certs.go
aaahrens: updated doc

lib/kcerts/ssh.go
ccontavalli: Let me put it a different way: we have a client, that is receiving a certificate from a server.
There's two fields that by spec should only have positive values. Negative values are invalid?
But the code uses those values anyway, doing some math on them.
It's hard for me to tell that the caller can / will detect all bad combinations.
(example: I don't see the uint64 conversion, but do see the result you describe if I run this on play.google.com)

The code in the golang library either rejects or handles correctly invalid values, with very simple code:
https://cs.opensource.google/go/x/crypto/+/a769d52b:ssh/certs.go;l=404

I think we should do the same? If you don't want to return error, return some value indicating that it has expired?

There's also a value that represents infinity: CertTimeInfinity, which sould be handled differently?

aaahrens: Added in more checks and fixed some broken tests with auth_server

ccontavalli (approved):

PR #142, opened 2021-07-28
Vendor jq and Update tests

aaahrens:
- Adds in the download_binary and declare_binary rules, which are simple wrappers
- Uses the wrapper to download binaries and executes them inside the ui tests.

### DIFF
--- a/bazel/utils/binary.bzl
+++ b/bazel/utils/binary.bzl
@@ -1,15 +1,6 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_file")
 
 def _download_binary(ctx):
-    """Downloads a single binary that is not tarballed.
-
-    Example:
-          download_binary(
-                name = "jq_macos_amd64",
-                binary_name = "jq",
-                uri = "https://github.com/stedolan/jq/releases/download/jq-1.6/jq-osx-amd64",
-            )
-    """
     ctx.download(
         url = [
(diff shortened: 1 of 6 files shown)


bazel/utils/binary.bzl
```diff
@@ -0,0 +1,72 @@
...
+            mandatory = True,
+        ),
+        "uri": attr.string(
+            mandatory = True,
+        ),
+    },
+)
+
+def _declare_binary(ctx):
+    """Declares a single binary, used as a wrapper around a select() statement
```
ccontavalli, line 38: move this docstring to be the `doc = """` parameter of `rule(...` (see kernel/defs.bzl for example).

Tl;Dr: documentation generally goes on the rule definition, the function _declare_binary generally gets no docstring, or a simple docstring like """Implements the declare_binary rule."""

bazel/utils/binary.bzl
```diff
@@ -0,0 +1,72 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_file")
+
+def _download_binary(ctx):
+    """Downloads a single binary that is not tarballed.
```
ccontavalli, line 4: some comment as below.

ccontavalli (approved):
this is great, move the docstrings as recommended, and merge at will.

ccontavalli (approved):

PR #143, opened 2021-07-30
bazel/linux/defs.bzl: get kernel dependencies to work correctly

ccontavalli:
Background:
Before this PR, a module depending on another did not
compile correctly. There were 2 separate issues:

1) Module.symvers file, was not exported/imported correctly.
   Beside the mechanics to import the file, the -C parameter
   passed to make causes it to change directory, and breaks
   bazel generated relative paths.

2) Most modules are compiled for a set of kernels. When
   one module depends on another, it must be linked against
   the same kernel version being built.

This PR:
- adds the correct logic to export and bring in Module.symvers
  files (thanks Raghu!)
- adds logic so that when a module depends on another, it selects
  the variant built with the same kernel version.

### DIFF
--- a/bazel/linux/defs.bzl
+++ b/bazel/linux/defs.bzl
@@ -135,23 +135,74 @@ KernelModulesInfo = provider(
     },
 )
 
+def _import_kernel_bundle_dep(name, ki, d, inputs, extra_symbols):
+    """Extracts information necessary to import a kernel bundle.
+
+    A kernel bundle represents a single kernel module compiled for
+    multiple kernels. Given a bundle d, a rule named name, trying
+    to compile a module for kernel ki, updates inputs and extra_symbols
+    with the correct list of dependencies.
+
+    Args:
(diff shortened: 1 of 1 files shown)


bazel/linux/defs.bzl
```diff
@@ -135,17 +135,74 @@ KernelModulesInfo = provider(
...
+    """
+
+    matched = 0
+    for sub in d[KernelBundleInfo].modules:
+        matched += int(_import_kernel_modules_dep(name, ki, sub, inputs, extra_symbols))
+
+    if matched <= 0:
+        fail("%s is being built for kernel %s, depends on target %s, but this " +
+             "target is NOT being built for the specified kernel", name, ki.package, d.label.name)
+
```
rajachan, line 168: Will this fail early, or will it fail only when building `name`, after a dep target has been built for all the other kernels?

rajachan (approved):

PR #152, opened 2021-08-10
Add OAuth2 Methods (stateless now)

aaahrens:
Adds in the ability to specify a single other oauth2 method (multiple are trivial to add) that must execute before the "required" flow occurs.

It accomplishes this by keeping track of which oauth2.Configs have been redeemed, and resets the flow until all have been finished.
I still need to figure out a way to e2e test the auth server against real OIDC.

    add in the --add-oauth CLI flag, which enables additional oauth methods
    adds in a flow controller to the Authenticator struct in lib/oauth

Now stateless deriving from the URL (no longer stateful)


lib/oauth/multi.go
```diff
@@ -0,0 +1,182 @@
...
+	"math/rand"
+	"net/http"
+	"reflect"
+)
+
+const requiredKey = "required"
+
+type MultiOauth struct {
+	RequiredAuth *Authenticator
+	OptAuth      map[string]*Authenticator
```
ccontavalli, line 21: make OptAuth a simple array: `OptAuth []*Authenticator`. So we have extra authenticator 0, extra authenticator 1, ...

Then, change `CompletedFlows`. Instead of having a `[]string`, make it a simple `int`. Which basically gets incremented, 0, 1, 2, 3, ...

Reasons:
1. Less state to move around (literally, fewer bytes).
2. Authentications all happen in the same order. a) Imagine users having to some times to authenticate with github first, and google oauth next, and some times in the reverse order. b) Imagine having to debug an issue related to how oauth providers work, and only happening in some cases, depending on order.
3. Simpler code - you can remove the uuid generator?

lib/oauth/factory.go
```diff
@@ -131,7 +135,7 @@ func (f *Flags) Register(set kflags.FlagSet, prefix string) *Flags {
 		"Prefer using the --"+prefix+"secret-file option - as it hides the secret from 'ps'. Secret key of the client to use with the oauth provider")
 	set.DurationVar(&f.AuthTime, prefix+"auth-time", f.AuthTime,
 		"How long should the token forwarded to the remote oauth server be valid for. This bounds how long the oauth authentication process can take at most")
-
+	set.ByteFileVar(&f.AdditionalOAuth, prefix+"oauth-file", "", "file for additional oauth flow to occur before required oauth")
```
ccontavalli, line 138: is this flag still needed?

lib/oauth/factory.go
```diff
@@ -359,14 +363,18 @@ func WithCookiePrefix(prefix string) Modifier {
 
 func WithFlags(fl *Flags) Modifier {
 	return func(o *Options) error {
-		if len(fl.OauthSecretJSON) == 0 && (fl.OauthSecretID == "" || fl.OauthSecretKey == "") {
-			return fmt.Errorf("you must specify the secret-file or (secret-key and secret-id) options")
+		var mods []Modifier
+		if len(fl.AdditionalOAuth) == 0 {
```
ccontavalli, line 367: is this logic still needed?

lib/oauth/factory.go
```diff
@@ -389,19 +397,60 @@ func WithFlags(fl *Flags) Modifier {
...
+			WithAuthTime(fl.AuthTime),
+			WithSecrets(fl.OauthSecretID, fl.OauthSecretKey),
+			WithSigningExtractorFlags(fl.SigningExtractorFlags),
+			WithAdditionalFlow(fl.AdditionalOAuth),
+		)
 		return Modifiers(mods).Apply(o)
 	}
 }
 
+func WithAdditionalFlow(fileContent []byte) Modifier {
```
ccontavalli, line 410: this function can be turned into an extra flag in `oauth/factory.go` to create a provider from a single json file? It would be a good improvement?

But not needed strictly in this PR now?

astore/server/main.go
```diff
@@ -135,13 +136,19 @@ func Start(targetURL, cookieDomain, oAuthType string, astoreFlags *astore.Flags,
 		return fmt.Errorf("could not initialize auth server - %s", err)
 	}
 
-	authWeb, err := oauth.New(rng, oauth.WithFlags(oauthFlags), auth.FetchCredentialOpts(oAuthType))
+	reqAuth, err := oauth.New(rng, oauth.WithFlags(oauthFlags), auth.FetchCredentialOpts(oAuthType))
+	if err != nil {
+		return fmt.Errorf("could not initialize oauth authenticator - %s", err)
+	}
+	optAuth, err := oauth.New(rng, oauth.WithFlags(optAuthFlags), auth.FetchCredentialOpts(oAuthType))
```
ccontavalli, line 143: Ok, one problem here. The end goal is for the user to authenticate once with our systems, and be good for the rest of the day.

At the end of the login process, 2 things happen:

a) An authentication token is returned via gRPC to the CLI interface, for use by enkit.

b) A cookie is set in the browser, containing an encrypted token. For use in web interfaces.

Web interfaces that care about b) are the enkit proxy (so user does not have to authenticate separately with it), and b) some of the astore URLs (I believe authenticated users can download stuff via browser for some artifacts - if not implemented yet, I'd sure like it). Pretty much, anything we put on GAE should check the credentials cookie, just like astore or the proxy.

Now... I'm confident that the code for the CLI works properly here. I'm not confident about the code for the browser.

I suspect the last `PerformAuth` you invoke overwrites the cookie for the browser. Further, I fear that if the user only half completes the authentication (example: only authenticates with google, and not github), the user will still get a valid authentication cookie!

Two easy alternative:
1) The oauth library supports a different cookie name per instance of the object. We could use it? I would prefer solution 2, for safety.

2) In the oauth library, break the `PerformAuth` method in two: 1) That checks and extracts the state, 2) That issues an authentication cookie. To not break other users of the API, `PerformAuth` is then changed to simply invoke function 1 and function 2 in order.

From MultiOauth, then, you only call function 1 from the oauth providers. And call function 2 only at the end of the process (eg, when all authentications have completed).

ccontavalli (commented):
Thank you! I really like this change. There's only ~3 important comments here, the rest are nits/cleanups.

lib/oauth/multi.go
```diff
@@ -0,0 +1,182 @@
...
+func NewMultiOAuth(rng *rand.Rand, required *Authenticator, opts ...*Authenticator) *MultiOauth {
+	var encs []*token.TypeEncoder
+	m := map[string]*Authenticator{}
+	for _, v := range opts {
+		encs = append(encs, v.authEncoder)
+		m[randSeq(rng, 25)] = v
+	}
+
+	m[requiredKey] = required
+	encs = append(encs, required.authEncoder)
```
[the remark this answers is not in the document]
aaahrens: while in this case it would work, generally it wouldn't because they can have multiple encoders. Ideally in the future they are different encoders (that can be part of the refactor)

lib/oauth/factory.go
aaahrens: yes, it will still error out due to ordering if this isnt there

lib/oauth/factory.go
aaahrens: It's already used this way, and registered as a flag <.<

lib/oauth/multi.go
```diff
@@ -0,0 +1,182 @@
...
+	flowState.CompletedFlows = append(flowState.CompletedFlows, key)
+	flowState.CurrentFlow = ""
+	data.State = flowState
+	return data, false, err
+}
+
+func (mo *MultiOauth) PerformLogin(w http.ResponseWriter, r *http.Request, lm ...LoginModifier) error {
+	options := LoginModifiers(lm).Apply(&LoginOptions{})
+	// if the passed in state does not match what we want, pack in the previous state to unpack later.
+	if _, ok := options.State.(*MultiOAuthState); !ok {
```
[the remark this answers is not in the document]
aaahrens: This issue is that decodeRaw handles options.State which is an interface{} and url query state which is a string.

So It's used in two places that have two different types. I prefer just having the decode raw because a lot of wire deserialization structs have ti and it's pretty intuitive.

astore/server/main.go
aaahrens: The solution for this really doesn't require any change in code yeah?

Just add `--opt-base-cookie=external` or something

astore/server/main.go
```diff
@@ -210,34 +217,47 @@ func Start(targetURL, cookieDomain, oAuthType string, astoreFlags *astore.Flags,
...
-
-		data, handled, err := authWeb.PerformAuth(w, r, copts...)
+		data, terminal, err := authWeb.PerformAuth(w, r, copts...)
 		if err != nil {
 			ShowResult(w, r, "angry", "Not Authorized", messageFail, http.StatusUnauthorized)
-			log.Printf("ERROR - could not perform login - %s", err)
+			log.Printf("ERROR - could not perform token exchange - %s", err)
+			return
+		}
+		if !terminal {
```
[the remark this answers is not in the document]
aaahrens: Done

lib/oauth/multi.go
```diff
@@ -0,0 +1,182 @@
...
+		options.State = mo.NewState(options.State)
+	}
+	// if the passed in options arent already a flow state, pack them up to unpack later.
+	state, err := mo.decodeRaw(options.State)
```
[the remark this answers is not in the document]
aaahrens: I posted a comment about decodeRaw

astore/server/main.go
```diff
@@ -210,34 +217,36 @@ func Start(targetURL, cookieDomain, oAuthType string, astoreFlags *astore.Flags,
...
+	// it will redirect to /a with additional logins.
 	mux.HandleFunc("/e/", func(w http.ResponseWriter, r *http.Request) {
 		copts := []kcookie.Modifier{kcookie.WithPath("/")}
 		if cookieDomain != "" {
 			// WithSecure and WithSameSite are required to get the cookie forwarded via the NASSH plugin in chrome (for SSH).
 			copts = append(copts, kcookie.WithDomain(cookieDomain), kcookie.WithSecure(true), kcookie.WithSameSite(http.SameSiteNoneMode))
 		}
-
-		data, handled, err := authWeb.PerformAuth(w, r, copts...)
+		data, _, err := authWeb.PerformAuth(w, r, copts...)
```
ccontavalli: why did you remove handled? The goal here is that if PerformAuth does a redirect, we don't call ShowResult below that can overwrite the http.Status and response returned.

(without checking handled here, ShowResult is always invoked, even after a redirect)

lib/oauth/factory.go
ccontavalli: WithAdditionalFlows should be renamed as WithJsonConfig?

lib/oauth/factory.go
ccontavalli: AdditionalOauth -> should be renamed JsonConfig.

lib/oauth/multi.go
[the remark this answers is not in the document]
ccontavalli: as per chat over VC, the if here is decoding the state already, so no reason to call decodeRaw anymore. At that point, decodeRaw is only ever called with a string as parameter, and can be simplified.

lib/oauth/multi.go
```diff
@@ -0,0 +1,153 @@
...
+
+func (mo *MultiOauth) NewState(Extra interface{}) *MultiOAuthState {
+	return &MultiOAuthState{
+		Extra: Extra,
+	}
+}
+
+var errStateNotType = errors.New("state did not match")
+
+func (mo *MultiOauth) decodeRaw(state interface{}) (*MultiOAuthState, error) {
```
ccontavalli, line 38: see comment below, change to ever only take a string only as parameter, after the change below decodeRaw is only used in one place.

lib/oauth/multi.go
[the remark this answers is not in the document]
ccontavalli: makese sense!

astore/server/main.go
```diff
@@ -210,34 +217,37 @@ func Start(targetURL, cookieDomain, oAuthType string, astoreFlags *astore.Flags,
...
-			log.Printf("ERROR - could not perform login - %s", err)
+			log.Printf("ERROR - could not perform token exchange - %s", err)
 			return
 		}
-
-		if key, ok := data.State.(common.Key); ok {
-			authServer.FeedToken(key, data)
-		}
-		if !handled {
+		if terminal {
```
ccontavalli, line 244: I like terminal better, but I'm concerned that `!has_redirected`

lib/oauth/multi.go
aaahrens: changed to decodeString

ccontavalli (approved):
PerformAuth is now broken: it cannot set a cookie after http.Redirect() has been called.

Please fix it, test it, and then feel free to merge the code, no further review, approving.

Suggestions (alternatives - ugly, no matter what):
- break PerformAuth in 3 functions instead of 2.
- just break out SetCookie, and then your multi code invokes ExtractAuth, and http.Redirect directly.

PR #155, opened 2021-08-16
Changes to Adam's PR

ccontavalli:


lib/oauth/oauth.go
```diff
@@ -644,27 +644,19 @@ func (a *Authenticator) CredentialsCookie(value string, co ...kcookie.Modifier)
...
 	}
 
 	auth, err = a.SetAuthCookie(auth, w, co...)
 	if err != nil {
-		return AuthData{}, false, err
+		return AuthData{}, true, err
 	}
 
-	return auth, CheckRedirect(w, r, auth), nil
+	return auth, true, nil
```
ccontavalli, line 659: This is changing the contract for PerformAuth, you need to update other users of PerformAuth, otherwise they'll be broken by this change.

PR #156, opened 2021-08-16
gee: improve ssh connectivity

jonathan-enf:
Added some better checks.  At one point, I did away with our generation
of an ssh key in favor of using enkit's certicate to authenticate to github.

Unfortunately, this didn't work as expected (the enkit cert only grants
access to enfabrica/internal, but not $GHUSER/internal).  So, for the moment,
it seems we still need gee to generate ssh keys for the user, so I rolled
back that part of the change but kept other improvements.

- Check enkit agent.
- move to new org-64667743@github.com remotes
- Use enkit cert exclusively.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -687,6 +687,19 @@ function _get_branch_rootdir() {
   echo "${BRDIR}"
 }
 
+function _confirm_default_yes() {
+  local _PROMPT RESP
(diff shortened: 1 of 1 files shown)


jonathan-enf:
re-opening after force-push accidentally closed it.

ccontavalli (approved):

PR #161, opened 2021-09-01
Adds in state to the DNS server 

aaahrens:
Does two things:
- refactors state to be a little more idiomatic
- adds in serialization and de serialization of state


machinist/mserver/command.go
```diff
@@ -61,5 +62,6 @@ func NewCommand(bf *client.BaseFlags) *cobra.Command {
 	c.PersistentFlags().IntVar(&cpf.DnsPort, "dns-port", 5353, "the udp port that the dns will be served on, also note it will also allocate the tcp socket on it as well")
 	c.PersistentFlags().StringSliceVar(&cpf.Domains, "domains", []string{}, "domains that the master ControlPlane will be serving")
 	c.PersistentFlags().StringVar(&cpf.BindNet, "bind-net", "127.0.0.1", "the address to bind the grpc listener to")
+	c.PersistentFlags().StringVar(&cpf.StateFile, "state", "./state.json", "file to write and load state to")
```
aaahrens, line 65: Genuinely not sure if I should use os.Getcwd here

machinist/mserver/controller.go
```diff
@@ -163,3 +166,18 @@ func (en *Controller) ServeAllRecords() {
...
+
+// WriteState writes state to the specified state file every 2 seconds. Will not exit or error out unless no statefile is
+// provided.
+func (en *Controller) WriteState() {
+	if en.stateFile == "" {
+		en.Log.Warnf("No path to state provided, state is fully in memory")
+	}
+	for {
+		<-time.After(en.stateWriteTTL)
+		if err := state.WriteController(en.State, en.stateFile); err != nil {
```
ccontavalli, line 178: when and how are entries expired? Eg, let's say a machine changes name. When does the old name go away? Let's say a machine is dead, when does the name go away?

machinist/mserver/command.go
ccontavalli: I think this is fine. A system daemon will likely need to store the state file in `/var/lib/$nameofdaemon`, and if run in a docker container, be persisted through a volume.

The only thing I would do is use lib/config/ and the store interface here, but it can come in a separate PR.
(so, good as is, no action from me)

machinist/state/controlplane_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+
+
+func TestReadInController(t *testing.T) {
+	rng := rand.New(srand.Source)
+	rngName := func() string {
+		return strconv.Itoa(rng.Int())
+	}
+	t.Run("Test consecutive read ins", func(t *testing.T) {
+		rname := rngName() + ".json"
```
ccontavalli, line 22: why not just use `TempFile`? and in the returned object, use the `Name()` method to discover the path? It'll help prevent spurious failures.

ccontavalli: If you do use TempFile, there's very little reason left to use the nested t.Run? so it could just be multiple top level `func Test...`?

machinist/mserver/controller.go
```diff
@@ -163,3 +166,18 @@ func (en *Controller) ServeAllRecords() {
 		_ = <-time.After(5 * time.Second)
 	}
 }
+
+// WriteState writes state to the specified state file every 2 seconds. Will not exit or error out unless no statefile is
+// provided.
+func (en *Controller) WriteState() {
+	if en.stateFile == "" {
+		en.Log.Warnf("No path to state provided, state is fully in memory")
```
ccontavalli, line 174: important: don't you need a return here? Otherwise the loop below will keep running, and fail forever?

machinist/mserver/factory.go
```diff
@@ -35,11 +33,36 @@ func DnsPort(dnsPort int) ControllerModifier {
...
+			controller.dnsServer = dnsServ
+		}
+		return nil
+	}
+}
+
+func WithStateFile(filepath string) ControllerModifier {
+	return func(controller *Controller) error {
+		s, err := state.ReadInController(filepath)
+		if err != nil {
```
ccontavalli, line 50: let's say the file on disk is corrupted (eg, SIGTERM, CTL+C, crash, ... in the middle of a write).
Should we log the error, and continue? Rather than keep the DNS down until an operator manually fixes the file?

(eg, if this happened at 4am, ignoring and continuing would basically allow the system to recover on its own, rather than require someone to wake up and fix it)

ccontavalli (commented):
a couple minor nits, change looks good!

machinist/mserver/controller.go
aaahrens: Expiry is going to be in a separate PR.

machinist/state/controlplane_test.go
aaahrens: The issue is that os.TempFile actually creates the file. So if I use that, ReadInController always fails and I can't test auto state file creation.

machinist/mserver/controller.go
aaahrens: good catch, it passed the test because you cant write to ""

machinist/mserver/factory.go
aaahrens: yeah seems fine to me.

machinist/state/controlplane_test.go
ccontavalli: ack, makes sense!

machinist/state/controlplane_test.go
```diff
@@ -0,0 +1,60 @@
...
+		for range make([]int, 10) {
```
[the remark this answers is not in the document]
ccontavalli: I meant everywhere in the file :) you have a few make([]int, 10 left.

ccontavalli (approved):
just get rid of the remainint makes, and you're good to merge!

PR #167, opened 2021-09-05
bats: bash automated test system

jonathan-enf:
I want to be able to write unit tests for bkup and gee, and "bats" seems to be
the richest way to create unit tests for bash.  Also, "bats" might be a better
alternative to `sh_test` in general, enabling easier testing of things other
than bash scripts.

bats is broken up into 3 separate github projects (bats-core, bats-support,
bats-assert), but really all 3 are needed to do anything useful.  The
`bats_test` rule created by this PR ensures that all 3 are available for all
tests.

This PR also adds a regression test to demonstrate use of bats, and validate
that the bzl rules work.


bazel/shellutils.bzl
```diff
@@ -0,0 +1,57 @@
+# Bazel rules for helping with shell scripts.
+
+def sh_escape(x):
```
ccontavalli, line 3: use skylib, shell, _quote? (we use skylib in other places already), remove this function?

bazel/shellutils.bzl
```diff
@@ -0,0 +1,57 @@
...
+  ctx.actions.write(
+      output = ctx.outputs.executable,
+      is_executable = True,
+      content = script,
+  )
+  return DefaultInfo(
+      runfiles = runfiles,
+  )
+
+bats_test = rule(
```
ccontavalli, line 37: add a `doc` attribute with an explanation of the rule, adjust the comment on top of the function accordingly. This will allow automated doc generation to work correctly.

bazel/shellutils.bzl
jonathan-enf: Done.

bazel/shellutils.bzl
jonathan-enf: Done.

WORKSPACE
```diff
@@ -67,6 +67,32 @@ yarn_install(
     yarn_lock = "//ui:yarn.lock",
 )
 
+# bats: Bash Automated Testing System
+http_archive(
```
ccontavalli, line 71: if you want to be able to re-use the `bats_test` rule from other repositories using enkit, the `http_archive` deps need to go in the `enkit_deps` macro, in `bazel/deps.bzl`

ccontavalli (approved):

WORKSPACE
jonathan-enf: Done.

PR #172, opened 2021-09-07
gee: new conflict resolution flow

jonathan-enf:
Instead of dropping the user into an interactive shell, the
new conflict resolution flow prompts the user for each
conflicting file in each commit, and asks the user to
choose between:
   - accepting "your" (child) version,
   - accepting "their" (parent) version,
   - running a merge tool
   - skiping a commit
   - aborting the rebase operation altogether.


ccontavalli (approved):
This looks like it would not generally be necessary with better handling of squash commits?

PR #176, opened 2021-09-09
gee: add pyformat to fix

jonathan-enf:


scripts/gee
```diff
@@ -2650,16 +2650,29 @@ function gee__fix() {
...
+  if [[ -x /usr/bin/clang-format-10 ]]; then
+    local -a CLANG_FILES=()
+    for FNAME in "${FILES[@]}"; do
+      if [[ "${FNAME}" =~ \.h$|\.c$|\.cc$|\.cpp ]]; then
+        CLANG_FILES+=( "${FNAME}" )
+      fi
+    done
+    if (( "${#CLANG_FILES[@]}" )); then
+      _info "clang-format: Fixing ${CLANG_FILES[*]}"
+      _cmd /usr/bin/clang-format-10 --verbose --style=Google -i "${CLANG_FILES[@]}"
```
ccontavalli, line 2663: no, we have bazel rules to format according to our style guide (from Google, but with minor corrections)

scripts/gee
```diff
@@ -2650,16 +2650,29 @@ function gee__fix() {
...
+
+  # pyformat, if available:
+  if [[ -x "${BDIR}/pyformat.sh" ]]; then
+    local -a PY_FILES=()
+    for FNAME in "${FILES[@]}"; do
+      if [[ "${FNAME}" =~ \.py$ ]]; then
+        PY_FILES+=("${FNAME}")
+      fi
+    done
+    _cmd "${BDIR}/pyformat.sh" "${PY_FILES[@]}"
```
ccontavalli, line 2675: fwiw, would be good to unify under a single way to run formatters. buildifier, clang-format, gazelle are currently run via bazel rules, we'll probably want to do something similar for pyformat.

scripts/gee
jonathan-enf: Unfortunately, bazel run :cc_format doesn't take any arguments, and always iterates over all files.  I'll fix that, and switch to //:cc_format.

jonathan-enf: Fixed, done.

Because @clang-format is part of internal's workspace, and clang-format.bzl is part of enkit's repo, it's impossible to test this change without first checking the code in, but it probably works.

scripts/gee
jonathan-enf: Agree -- some toplevel "fixit" command that just runs everything on all edited files in the branch.  "gee fix" was meant to be that, but it should work without gee.  Another day.

ccontavalli (approved):

PR #180, opened 2021-09-09
Adds in useMultiFlag

aaahrens:
- Creates an interface for authenticators. This should be nice for testing later on, 
- Lets you choose on oauth2  whether or not to use the multi oauth2 flow. It defaults to not using it.

### DIFF
--- a/lib/oauth/types.go
+++ b/lib/oauth/types.go
@@ -4,7 +4,10 @@ import (
 	"github.com/enfabrica/enkit/lib/khttp/kcookie"
 	"net/http"
 )
-
+// An IAuthenticator is any object capable of performing authentication for a web server.
+// PerformLogin initiates the login process.
+// PerformAuth is invoked at the end, to verify that the login was successful.
+// Complete is used to verify that the returned AuthData indicates that the process is now complete.
 type IAuthenticator interface {
 	PerformLogin(w http.ResponseWriter, r *http.Request, lm ...LoginModifier) error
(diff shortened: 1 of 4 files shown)


scripts/BUILD.bazel
```diff
@@ -2,16 +2,16 @@ load("@//bazel/astore:defs.bzl", "astore_upload")
 
 filegroup(
     name = "gee-script",
-    srcs = [ "gee" ],
+    srcs = ["gee"],
```
ccontavalli, line 5: revert this file from this PR? unrelated to the change being introduced?

lib/oauth/types.go
```diff
@@ -0,0 +1,12 @@
+package oauth
+
+import (
+	"github.com/enfabrica/enkit/lib/khttp/kcookie"
+	"net/http"
+)
+
+type IAuthenticator interface {
```
ccontavalli, line 8: Can you add a coment like:

```
// An IAuthenticator is any object capable of performing authentication for a web server.
// PerformLogin initiates the login process.
// PerformAuth is invoked at the end, to verify that the login was successful.
// Complete is used to verify that the returned AuthData indicates that the process is now complete.
```

Would be good to add an example/move the example on top of oauth.go to be here. Another day.

ccontavalli (approved):

lib/oauth/types.go
aaahrens: I also thing it would be a good idea to change this to authenticator at some point, and have single.go and multi.go

PR #187, opened 2021-09-13
add gee-shellcheck_test

jonathan-enf:
gee: use shelltest as a regression test for gee.

This should help me avoid embarrassing mistakes in the future, though this tool really needs to be rewritten in a real language.

### DIFF
--- a/scripts/BUILD.bazel
+++ b/scripts/BUILD.bazel
@@ -1,5 +1,5 @@
 load("@//bazel/astore:defs.bzl", "astore_upload")
-load("@//bazel:shellutils.bzl", "bats_test")
+load("@//bazel:shellutils.bzl", "bats_test", "shellcheck_test")
 
 sh_library(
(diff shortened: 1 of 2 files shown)


scripts/BUILD.bazel
```diff
@@ -12,6 +12,11 @@ bats_test(
     deps = [":gee-lib"],
 )
 
+shellcheck_test(
```
ccontavalli, line 15: shellcheck in the end is a linter, it can catch bugs before the code is even run.

I think we should have a `en_sh_library` or `en_sh_binary` target that as part of the build, runs linters.

Eg, as a pattern/policy, we should not learn about automatically detectable problems _after_ someone adds a test target, and actually runs it, we should learn about them at build time, as we're writing the code.

This is a very good change, though, approving!

ccontavalli (approved):

PR #192, opened 2021-09-15
lib/kflags: add ability to provide command line flags from a map.

ccontavalli:
This change introduces a new Augmenter, the MapAugmenter, which allows
to provide flag defaults from a simple map, provided at the agumenter
creation time.

The MapAgumenter can be configured to use one or more EnvManglers
(now renamed VarMangler) to allow to compute a different key from
a flag name. For example, by using a Camel mangler, for a flag named
"network-timeout", the key "NetworkTimeout" is looked up in the flag.

Additionally, this PR restructures manglers to be more genereically
useful, and introduces a couple new ones.

No functional change to the codebase, this just adds new library
functions that will be used in an upcoming PR.

### DIFF
--- a/lib/kflags/map.go
+++ b/lib/kflags/map.go
@@ -1,0 +1,73 @@
+package kflags
+
+type MapAugmenter struct {
+	args map[string]string
+	manglers []VarMangler
+}
+
+type MapModifier func(e *MapAugmenter)
+
+type MapModifiers []MapModifier
+
+func (ems MapModifiers) Apply(e *MapAugmenter) {
+	for _, em := range ems {
+		em(e)
+	}
+}
+
+// WithMapMangler sets the list of manglers to use to lookup the parameters in the map.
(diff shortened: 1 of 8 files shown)


lib/kflags/map.go
```diff
@@ -0,0 +1,73 @@
...
+	augmenter := &MapAugmenter{
+		args: args,
+		manglers: DefaultVarMangler,
+	}
+
+	MapModifiers(mods).Apply(augmenter)
+	return augmenter
+}
+
+// VisitCommand implements the VisitCommand interface of Augmenter. In AssetAugmenter, it is a noop.
```
aaahrens, line 46: For readability and compile time checks, could you add a type assertion?
`var _ Augmenter = &MapAugmenter{}`

aaahrens (changes requested):

aaahrens (approved):

PR #229, opened 2021-10-01
license_manager: Add proto definition file

minor-fixes:
This change adds a gRPC definition for the license_manager service,
intended to supplant the existing license service under `//manager` with
a protocol that uses multiple unary RPCs instead of a bidirectional
streaming RPC, in an effort to clarify the implementation of both the
client and the server.

Requirements and more info can be found in: https://docs.google.com/document/d/1TNqbBprpcNU9tTHVCFzRwaQoHlGFdjkw221C5p9UsAw/edit

Tested: All targets build:

```
bazel build //license_manager/...
```

Jira: INFRA-90

### DIFF
--- a/license_manager/proto/license_manager.proto
+++ b/license_manager/proto/license_manager.proto
@@ -1,0 +1,175 @@
+syntax = "proto3";
+
+package license_manager.proto;
+
+import "google/protobuf/timestamp.proto";
+
+// Describes a vendor-agnostic frontend for allocating licenses to hardware tool
+// command invocations.
+// Terminology:
+//
+//   License: A tuple of (vendor, license_type) strings that identifies a
+//     resource that can only be used by one process at a time. These resources
+//     are ultimately fetched from and returned to a vendor-specific license
+//     server.
+//
+//     Licenses are "allocated" when the LicenseManager frontend associates it
+//     with an invocation.
+//
+//     Licenses are "in use" when the invocation actually uses a license from
+//     the underlying license server. Licenses should not be "in use" unless
+//     they are also "allocated", but "allocated" licenses are not always "in
+//     use".
+//
+//   Invocation: A single execution of a particular hardware toolchain binary.
+//     Invocations may use a license, and to do so they must be allocated one
+//     before the underlying hardware toolchain process is allowed to execute.
+
+service LicenseManager {
+  // Allocate attempts to allocate a specific license type for this invocation.
+  // If successful, invocations will keep the allocation alive by calling
+  // Refresh(). If the invocation is queued for the license instead, invocations
+  // will continue polling with Allocate() until allocation is successful.
+  rpc Allocate(AllocateRequest) returns (AllocateResponse) {}
+
+  // Refresh refreshes a license allocation while an invocation is still
+  // potentially using the license. If the invocation fails to refresh the
+  // allocation, the underlying license may be allocated to another invocation.
+  rpc Refresh(RefreshRequest) returns (RefreshResponse) {}
+
+  // LicensesStatus returns the status of all license types, as reported by both
+  // the LicenseManager and the underlying license servers.
+  rpc LicensesStatus(LicensesStatusRequest) returns (LicensesStatusResponse) {}
+}
+
+message AllocateRequest {
+  // Licenses to acquire for this invocation.
+  //
+  // Invocations that need multiple of the same license type should insert
+  // duplicate entries in this field.
+  //
+  // Implementations currently only support reserving once license per
+  // invocation, though future implementations may allow for allocations of
+  // multiple and of different types.
+  repeated License license = 1; // required
+
+  // Owning entity issuing the allocation request. Used for logging purposes
+  // only. This could be the name of the user or system issuing the request.
+  string owner = 2; // required
+
+  // Build tag of the allocation request. This tag does not need to be unique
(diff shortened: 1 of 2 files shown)


license_manager/proto/license_manager.proto
```diff
@@ -0,0 +1,171 @@
...
+
+message RefreshRequest {
+  // Opaque identifier for this invocation, as returned by the server in the
+  // AllocateResponse that allocated this license.
+  string invocation_id = 1; // required
+
+  // If true, the invocation is indicating that it no longer requires the
+  // allocated license, and the server should deallocate it. The resulting
+  // RefreshResponse can be expected to have `allocated` == false.
+  bool release = 2;
```
ccontavalli, line 119: I would probably recommend a different call, Release() with RefreshRequest as parameter, if only for clarity of the API.

license_manager/proto/license_manager.proto
```diff
@@ -0,0 +1,171 @@
...
+  rpc Allocate(AllocateRequest) returns (AllocateResponse) {}
+
+  // Refresh refreshes a license allocation while an invocation is still
+  // potentially using the license. If the invocation fails to refresh the
+  // allocation, the underlying license may be allocated to another invocation.
+  rpc Refresh(RefreshRequest) returns (RefreshResponse) {}
+
+  // LicensesStatus returns the status of all license types, as reported by both
+  // the LicenseManager and the underlying license servers.
+  rpc LicensesStatus(LicensesStatusRequest) returns (LicensesStatusResponse) {}
```
ccontavalli, line 42: this is good for debugging and CLI, although as discussed the primary monitoring will be via prometheus.

ccontavalli (approved):
This is good, approving, all minor nits, no need for further review from me. Merge at will, iterate.

Only thing that is not covered are error conditions: eg, what if the license the client is attempting to allocate does not exist? (eg, the license kind of type is not known, not all are out), or if a client refreshes with an invalid invocation id?
How are refresh timers handled?

Those can all be addressed as the code is written, I'd just recommend updating the comments accordingly.

license_manager/proto/license_manager.proto
minor-fixes: My thoughts exactly.

license_manager/proto/license_manager.proto
minor-fixes: > what if the license the client is attempting to allocate does not exist?
Probably will return `NOT_FOUND` as opposed to `RESOURCE_EXHAUSTED` in the all-out-of-licenses case, to indicate that this case is not retryable. This does require docs, and I can add them as the implementation comes in.

> or if a client refreshes with an invalid invocation id
This probably requires more thought; I'm thinking to support server restarts, the server has a startup period (related to the queue/allocation refresh times) where new and existing queue requests with unknown `invocation_id`s are re-queued, refresh requests for allocations are "adopted" (by decrementing the corresponding counter, in an effort to not have to cancel running invocations) and no new allocations are issued from the queue until the end of the startup period. After the startup period, allocation refresh requests with an unknown ID could either:

* return a successful response with `allocated == false` telling the client to immediately kill the invocation (it doesn't have a license)
* return an error response (maybe `FAILED_PRECONDITION`) and clients should immediately kill the invocation.

I'm tempted to lean towards the latter because returning an error has a higher chance of being logged and therefore alert us more readily to this state that shouldn't happen.

Similarly, for queue refreshes with an unknown ID, we could:

* re-queue the request at the end, successfully
* return an error response (`FAILED_PRECONDITION`)

Leaning towards the former in this case to simplify the client execution - if the client is waiting in the queue and then suddenly its place in the queue is "forgotten", the client would need logic to detect this error and re-queue, vs. the server auto-requeueing the client.

WDYT?

minor-fixes: > Release() with RefreshRequest as parameter

Are you suggesting to rename `Refresh` to `Release`, or add a separate call named `Release`, or something else?

aaahrens (approved):
Don't really see any issue with this that hasn't been raised

license_manager/proto/license_manager.proto
ccontavalli: WDYT, first message: +100, well thought, having support for rolling restarts while builds are in progress sounds great. Minor additions:
- client should be written so that if the server goes down, builds in progress keep moving forward.
- I would argue that allocated = true/false is no longer necessary? in the RefreshResponse message? We can just use error values? Refresh successful -> all good. Refresh returning error -> either server is down, or license is no longer allocated, or ... it's part of the error handling code, checking different error status?
- client ... we need to decide what to do if server is down at startup. Assume there are licenses, and move forward? or assume there are no licenses, and wait? probably a flag, but let's discuss this later?

Re Release: I was suggesting having a different Release method, rather than Refresh with release = true. Don't see any big advantage in having a bool flag, separate Release method likely makes for a more straightforward API?

minor-fixes: > I would argue that allocated = true/false is no longer necessary?

Removed.

>  Assume there are licenses, and move forward? or assume there are no licenses, and wait? probably a flag, but let's discuss this later

Agreed - probably the latter by default, with an override for emergencies.

minor-fixes: > Re Release: I was suggesting having a different Release method

Done. Thanks!

PR #232, opened 2021-10-03
gee affected: report/run affected targets

jonathan-enf:
Add "affected" command to gee.

PR #230 sounded like a good idea, so I threw a prototype implementation into gee.
This builds on PR #228.

- gee whatsout: show differences from parent rather than main.
- gee__whatsout: read parents file
- fix polarity of git diff --quiet exit code
- Added "affected" command.


scripts/gee
```diff
@@ -2235,19 +2245,82 @@ _register_help "whatsout" \
...
+  _info "Reading list of changed files."
+  local -a FILES=()
+  _read_cmd FILES "${GIT}" diff --name-only "${PARENT}"
+  if _branch_has_unstaged_changes; then
     _info "Note: This branch contains uncommitted changes."
   fi
+
+  _info "Translating changed files into targets."
+  local -a TARGETS=()
+  NOFAIL=1 _read_cmd TARGETS bazelisk query --keep_going --noshow_progress \
```
minor-fixes, line 2303: I've written a changed targets detector twice over the past few years based on VCS-reported files changed + `bazel query`. The main drawback is that while it works for source files:

```
[~/dev/enkit] (bazel_diff_impl)
bminor-> bazel query --keep_going --noshow_progress 'set(manager/rpc/license.proto)' 2>/dev/null
//manager/rpc:license.proto
[=== 2021-10-04 16:52:32 === SUCCESS ===]

[~/dev/enkit] (bazel_diff_impl)
bminor-> bazel query --keep_going --noshow_progress 'rdeps(//..., set(//manager/rpc:license.proto))' 2>/dev/null
//manager/server:upload-server
//manager/server:server
//manager/server:license-manager
//manager/server:license-manager.binary
//manager/server:go_default_library
//manager/common:go_default_library
//manager/client:client
//manager/client:go_default_library
//manager/rpc:go_default_library
//manager/rpc:license_go_proto
//manager/rpc:license_proto
//manager/rpc:license.proto
[=== 2021-10-04 16:52:48 === FAIL: 3 ===]
```

, it doesn't work for any bazel files:

```
[~/dev/enkit] (bazel_diff_impl)
bminor-> bazel query --keep_going --noshow_progress 'set(manager/rpc/BUILD.bazel)' 2>/dev/null
//manager/rpc:BUILD.bazel
[=== 2021-10-04 16:53:01 === SUCCESS ===]

[~/dev/enkit] (bazel_diff_impl)
bminor-> bazel query --keep_going --noshow_progress 'rdeps(//..., set(//manager/rpc:BUILD.bazel))' 2>/dev/null
# bazel prints `INFO: Empty results` to stderr
[=== 2021-10-04 16:53:11 === FAIL: 3 ===]
```

Specifically, `BUILD`, `WORKSPACE`, and `*.bzl` changes don't result in changed targets as conveniently as one might like.

The implementation going into `enkit` will take a different approach: at two points in the source ("before" and "after") dump the dependency graph and process a hash for each target such that any change to a target's attributes, input files, or dependencies' hash will cause a hash change. Then diff the target -> hash mapping between "before" and "after" to determine changed targets, ignoring targets only present in "before" and including targets only present in "after".

This implementation will be in Go to do this more involved processing (though the initial version may shell out to https://github.com/Tinder/bazel-diff for a faster implementation to unblock a better presubmit for enkit and internal)

---

If we want this subcommand in `gee` I can see a few ways forward:

* Output a conservative estimate on changed targets based purely on source files - something like "based on files XXX, the following targets have definitely changed: YYY; changes in the following files may have caused targets not listed here to change: ZZZ"
* Output a generous estimate on changed targets by:
  * assuming any changes to BUILD files affects all targets in that file (`foo/bar/BUILD.bazel` -> `//foo/bar:all-targets`)
  * assuming any changes to `WORKSPACE` or `*.bzl` files affects all targets
* Shell out to the enkit implementation after it is done

What do you think is the most useful here?

PR #234, opened 2021-10-05
testing: Add errdiff lib

minor-fixes:
This change adds a testonly library called "errdiff" that allows unit
tests to easily see if errors match an expected substring.

This allows for a bit more fidelity than checks by the assert package,
as the latter only has functions for checking error presence (though
oftentimes the error expected in tests is not the one actually returned)
and exact error equivalence (though it may be difficult and/or brittle
to formulate an expectation on an exact error value).

### DIFF
--- a/lib/errdiff/errdiff.go
+++ b/lib/errdiff/errdiff.go
@@ -1,0 +1,34 @@
+package errdiff
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Substring checks that an error `got` matches the expectation `want`.
+// Empty `want` string means no error is expected; if not empty, `got` must be
+// non-nil and contain `want` as a substring to meet the expectation.
+// Returns an empty string if `got` meets the `want` expectation, or a string
+// containing an explanation of the discrepancy otherwise.
+//
+// Example use:
+//
+//   if diff := errdiff.Substring(gotErr, "some substring"); diff != "" {
+//     t.Error(diff)
+//   }
+func Substring(got error, want string) string {
+	switch {
+	case got == nil && want == "":
+		return ""
+	case got == nil && want != "":
+		return fmt.Sprintf("got no error; want error containing %q", want)
+	case got != nil && want == "":
+		return fmt.Sprintf("got error: '%v'; want no error", got)
+	case got != nil && want != "":
+		if strings.Contains(got.Error(), want) {
+			return ""
+		}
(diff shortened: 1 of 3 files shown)


lib/errdiff/errdiff.go
```diff
@@ -0,0 +1,34 @@
+package errdiff
```
aaahrens, line 1: So if you ever want to make a library that's importable by tests only, you can give it a no release annotation at the top.
```
//+build !release
```

You can technically import it in a library, but it would fail on building.

lib/errdiff/errdiff.go
```diff
@@ -0,0 +1,34 @@
...
+// non-nil and contain `want` as a substring to meet the expectation.
+// Returns an empty string if `got` meets the `want` expectation, or a string
+// containing an explanation of the discrepancy otherwise.
+//
+// Example use:
+//
+//   if diff := errdiff.Substring(gotErr, "some substring"); diff != "" {
+//     t.Error(diff)
+//   }
+func Substring(got error, want string) string {
```
aaahrens, line 19: So honestly I would nix this whole PR, this already exists within the errors package by default. 

https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/errors/wrap.go;l=39

Also generally, don't be afraid of passing in a *testing.T to a helper function for testing. It's a good way of soft locking a function to just test.

aaahrens: Just some example code on how to use it with fmt.Errorf
```
	somethingErr := errors.New("library: something error")
	myErr := fmt.Errorf("err: %w %s", somethingErr, "my path")
	fmt.Println(errors.Is(myErr, somethingErr)) //true
```

aaahrens (changes requested):

lib/errdiff/errdiff.go
minor-fixes: Since we use bazel, I've marked it as `testonly` there; this also fails the build if the library makes it into a non-testonly target. Is it necessary to do both?

aaahrens (approved):

minor-fixes:
As discussed in our meet chat:
* moving forward with this to unblock future PRs
* if we find a more robust style that we like for checking error context in unit tests, we can replace uses of this
* there should be a function that takes a *testing.T and does its own assertion rather than making the tests do it

I've implemented the last point by making `Substring()` unexported, but using it as the main logic for `Check()`, which takes a `*testing.T`.

PR #235, opened 2021-10-05
git: Add Go lib with temp worktree generation

minor-fixes:
This change adds a git library, with initial support for generating
worktrees from a provided repo path.

The bazel changed target detector will need the ability to run at two
different points in the source history locally. One way to do this is to
issue checkout commands in the user's repository; however, it is
possible that a crash/error will leave a user's repository in an
unexpected/unknown state.

Instead, we can shell out to `git worktree` to create essentially a
shallow clone at a particular committish in a different directory (in
this case, a temporary directory) and perform bazel queries on these
temporary worktrees. Here the only risk is that we fail to delete the
temp worktrees upon crash/error, which are trivially cleaned up later
manually or via reboot.

go-git does have the ability to traverse a worktree, but it does not
appear to be trivial to spill a particular commit to disk so that we can
invoke `bazel` on it.

This change adds a small library that shells out to `git worktree` to
add/remove the worktree instead.

Tested: Unit tests in this commit

Jira: INFRA-52


minor-fixes:
Draft FYI - This includes contents of parent PRs

TODO before this moves out of draft:

- [ ] Submit parent PR and rebase

lib/git/worktree.go
```diff
@@ -0,0 +1,51 @@
...
+	return w.worktreePath
+}
+
+func (w *TempWorktree) Close() error {
+	cmd := exec.Command("git", "worktree", "remove", w.worktreePath)
+	cmd.Dir = w.repoPath
+	_, err := runCommand(cmd)
+	if err != nil {
+		err = fmt.Errorf("failed to delete temp git worktree at %q: %v", w.worktreePath, err)
+		glog.Warning(err)
```
ccontavalli, line 47: returning error should be enough? The caller can interpret the meaning of an error here? would remove glog.

lib/git/worktree_test.go
```diff
@@ -0,0 +1,60 @@
...
+					Dir:  "/foo/bar",
+				},
+				&exec.Cmd{
+					Path: "git",
+					Args: []string{"worktree", "remove", "foo"},
+					Dir:  "/foo/bar",
+				},
+			},
+		},
+	}
```
ccontavalli, line 37: not objecting, but there is a single testcase here. Is it worth the complexity of having this struct/array? makes the test arguably much harder to read and understand.

lib/git/worktree_test.go
```diff
@@ -0,0 +1,60 @@
...
+	for _, tc := range testCases {
```
ccontavalli, line 38: minor: the test mocks out git (good!) and checks that the generated commands are as expected (also good!).

But does not test that the actual commands are valid git commands. Is it worth adding a simple test that 1) opens a temp directory (maybe just some testdata/ moved over), and does something like 2) git init, git add . git commit -m "blah", and then 3) tries the worktree commands?

(it's more of an integration test, but should take a few seconds to run)

lib/git/worktree_test.go
minor-fixes: Yeah, this is nonsensical; removed the test table.

lib/git/worktree.go
```diff
@@ -0,0 +1,51 @@
+package git
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os/exec"
+
+	"github.com/golang/glog"
+)
+
```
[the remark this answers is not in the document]
minor-fixes: Oops I have no docstrings! Fixed.

PR #236, opened 2021-10-05
bazel: Implement checkout, query in `enkit bazel affected-targets`

minor-fixes:
This change partially implements `enkit bazel affected-targets` by doing
the following:

* creating temporary git worktrees for the specified VCS points
* performing a query to list all targets
* parsing the response and listing the total number of targets

Tested: Unit tests in this commit

Benchmarks:

`enkit bazel affected-targets list` currently finishes in ~103s on my
machine on the first run, and ~73s on subsequent runs (which save some
overhead on bazel fetching, since cache directories are reused). We can
expect presubmit machines to always do the former, and it's possible
that this can time can be reduced by half in a future PR by issuing
bazel queries on both worktrees in parallel.

Jira: INFRA-52


minor-fixes:
Draft FYI - This includes contents of parent PRs

TODO before this moves out of draft:

- [ ] Submit parent PR and rebase

lib/bazel/commands/commands.go
```diff
@@ -87,6 +91,84 @@ func NewAffectedTargetsList(parent *AffectedTargets) *AffectedTargetsList {
...
+	}
+	return filepath.Join(d, "enkit/bazel", subDir), nil
+}
+
+func findGitRoot() (string, error) {
+	bazelWorkspace := os.Getenv("BUILD_WORKSPACE_DIRECTORY")
+	if bazelWorkspace != "" {
+		return bazelWorkspace, nil
+	}
+	return os.Getwd()
```
ccontavalli, line 173: this is unlikely to be the git root? We have the code below in `SuggestGitName` (astore/client/astore/astore.go), basically: git.open current directory, and then get the worktree.

```
        repo, err := git.PlainOpenWithOptions(filepath.Dir(name), &git.PlainOpenOptions{
                DetectDotGit: true,
        })
        if err != nil {
                return "", err
        }

        tree, err := repo.Worktree()
        if err != nil {
                return "", err
        }
        root := tree.Filesystem.Root()

        absRoot, err := filepath.Abs(root)
        if err != nil {
                return "", err
        }

        absName, err := filepath.Abs(name)
        if err != nil {
                return "", err
        }

        destName, err := filepath.Rel(absRoot, absName)
        if err != nil {
                return "", err
        }

        remotes, err := repo.Remotes()
        if err != nil {
                return "", err
        }
```

lib/bazel/commands/commands.go
```diff
@@ -87,6 +91,84 @@ func NewAffectedTargetsList(parent *AffectedTargets) *AffectedTargetsList {
...
+	if err != nil {
+		return fmt.Errorf("failed to open bazel workspace: %v", err)
+	}
+
+	// Get all target info for both VCS time points.
+	targets, err := startWorkspace.Query("deps(//...)", bazel.WithKeepGoing(), bazel.WithUnorderedOutput())
+	if err != nil {
+		return fmt.Errorf("failed to query deps for start point: %v", err)
+	}
+	// TODO(scott): Replace with logging
```
ccontavalli, line 147: yeah, I'd pass in a logger object to the function. Also consider that ... do we send the log to stderr? syslog? debug as well? a file on disk? ...

You can use: https://github.com/enfabrica/enkit/blob/master/lib/logger/klog/factory.go#L48 and friends to create / populate the object.

(in the main/app - in the main/app, though, it's fine to initialize and pass a glog object into the command initializer)

lib/bazel/commands/commands.go
```diff
@@ -87,6 +91,84 @@ func NewAffectedTargetsList(parent *AffectedTargets) *AffectedTargetsList {
...
+func cacheDir(subDir string) (string, error) {
+	d, err := os.UserCacheDir()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(d, "enkit/bazel", subDir), nil
+}
+
+func findGitRoot() (string, error) {
+	bazelWorkspace := os.Getenv("BUILD_WORKSPACE_DIRECTORY")
```
ccontavalli, line 169: Note that the BUILD_WORKSPACE_DIRECTORY may not match the git root. WORKSPACE may start in a subdirectory, or even have multiple WORKSPACEs in the same repo.

Let's not worry about handling this complexity in code, but I'd rename the function to match the intent.

Intent is to find workspace root? call it findWorkspaceRoot, use code like I have at the bottom of this code.

Intent is to find the git root? remove the check for BUILD_WORKSPACE_DIRECTORY,  invoke the comment in my next comment.

Here's a snippet of code I wrote in the past to find the WORKSPACE root:
```
import (
        "github.com/bazelbuild/buildtools/wspace"
)

func WorkspaceRoot() string {
        root := os.Getenv("BUILD_WORKSPACE_DIRECTORY")
        if root != "" {
                return root
        }

        root, _ = wspace.FindWorkspaceRoot("")
        return root
}
```

lib/bazel/options.go
```diff
@@ -0,0 +1,111 @@
...
+// flags returns the startup flags as passed to bazel.
+func (o *baseOptions) flags() []string {
+	var f []string
+	if o.OutputBase != "" {
+		f = append(f, "--output_base", o.OutputBase)
+	}
+	return f
+}
+
+// apply applies all the options to this option struct.
```
ccontavalli, line 39: lol. Generally in the repository we've used the opposite convention, see example below.

I would stick to what we did in the rest of the repository, not because I have any opinion on one being worse/better than the other, but for consistency. It took me 30 seconds to realize what was different here.

```
type Options []Option

func (opts Options) apply(o *baseOptions) {
  for _, opt := range opts {
    opt(o)
  }
}
```

lib/bazel/options.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// filterError filters out expected error codes based on the provided query
+// arguments.
+func (o *queryOptions) filterError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	if o.keepGoing {
```
ccontavalli, line 75: nit: would invert the if, and merge with above.

```
if err == nil || !o.keepgoing {
  return err
}

if err, ok := err.(*exec.ExitError); ok {
...
}
..
```

lib/bazel/query.go
```diff
@@ -0,0 +1,113 @@
...
+		return nil, nil, fmt.Errorf("can't get stdout for bazel query: %v", err)
+	}
+	err = cmd.Start()
+	if err != nil {
+		return nil, nil, fmt.Errorf("can't start bazel command: %v", err)
+	}
+
+	pipeReader, pipeWriter := io.Pipe()
+	errChan := make(chan error)
+	go func() {
```
ccontavalli, line 34: don't fully understand why you use this function. Why not just return stdout, and let the caller consume that pipe directly? stdout itself is io.ReadCloser, why is pipeReader preferred? the error will be delivered on cmd.Wait called by the caller.

Even if we wanted errChan, we could limit go func() to calling cmd.Wait()? and let the reader use stdout directly?

lib/bazel/query.go
```diff
@@ -0,0 +1,113 @@
...
+			if err := proto.Unmarshal(buf, &target); err != nil {
+				readErr = err
+				return
+			}
+			targets[targetName(&target)] = &target
+		}
+		if err != io.EOF {
+			readErr = err
+		}
+	}()
```
ccontavalli, line 82: also not sure why you need this goroutine. why not have the code directly in the function? it loops, and then closes readDoneChan.

Immediately after the goroutine, you wait for it to complete. Sounds like you could just run the code in the function directly?

BUILD.bazel
```diff
@@ -23,6 +23,13 @@ gazelle(
     prefix = "github.com/enfabrica/enkit",
 )
 
+# Custom gazelle resolve mappings
+# Some third-party dependencies that are bazel-aware bamboozle gazelle running
+# in our repository, which tries to generate dependency labels that don't exist.
+# These mappings keep gazelle from generating bogus dependency labels.
```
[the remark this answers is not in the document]
minor-fixes: Ack; though it sounds like we would also have to confirm that none of our dependencies use the old naming convention? IOW, what happens when we mix and match deps that use both naming conventions?

lib/bazel/commands/commands.go
```diff
@@ -87,6 +91,84 @@ func NewAffectedTargetsList(parent *AffectedTargets) *AffectedTargetsList {
...
+
 	return fmt.Errorf("not yet implemented")
 }
 
+func cacheDir(subDir string) (string, error) {
+	d, err := os.UserCacheDir()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(d, "enkit/bazel", subDir), nil
```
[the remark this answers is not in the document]
ccontavalli: So, we don't override the default basel `--output_base`, so it's [computed based on a hash of the path of the repository](https://docs.bazel.build/versions/main/command-line-reference.html#flag--output_base).

Given that some hardware builds take a long time, it's not uncommon for hardware folks to launch a command from a workdir, and then create a different workdir where they code code code bazel bazel bazel.

How about using the same approach as bazel here? hash(realpath(CWD)) as the root of the cache file?

Another option is to use COMMIT HASH of the checked out tree? so we can cache past tree invocations easily without a server?

This works for past tree, current tree may have local changes... ugh.

Something like:
`hash(realpath(cwd) + magicSeparator + commitHash)` with commitHash = 0 for top of tree with local changes may work?

BUILD.bazel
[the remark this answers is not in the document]
minor-fixes: Makes sense.

lib/bazel/commands/commands.go
minor-fixes: Hmm the intent is complected.

There is an assumption that the git root and the workspace root are one and the same for repos in which this will be run (enkit and internal, do you foresee a scenario where this assumption breaks down?)

I guess the "correct" method is both: we need to find the git root to execute the worktree commands, and then find the workspace root of the resulting worktrees.

`BUILD_WORKSPACE_DIRECTORY` is still needed if we expect people to invoke `enkit` via `bazel run` (do we?) because otherwise PWD will be completely outside the git tree.

lib/bazel/options.go
minor-fixes: Fixed

lib/bazel/commands/commands.go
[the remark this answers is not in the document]
ccontavalli: uhm .. up to you. I'm ok iterating later, but strongly prefer not having code that can cause overwriting/corrupting files by another runner - so temp directory as initial solution sounds good.

Re expected use: I think it depends entirely on how fast your implementation will be.

My typical usage pattern is that 1) I work in a directory, 2) I run tests in that directory. But here we have wild dependencies: I change a CSR file, a bunch of hardware or software stuff breaks.

If we had a simple "run all relevant tests" command that is FAST, I would strongly advertise it for everyone to use as the default "use this command to run tests".

ccontavalli (commented):
only comments left open are the goroutine ones?

lib/bazel/query.go
minor-fixes: Inlined this into the function.

lib/bazel/query.go
```diff
@@ -0,0 +1,102 @@
...
+
+// streamedBazelCommand exec's out to bazel in the specified workspace with the
+// specified arguments. It is a variable that allows for stubbing during tests.
+//
+// In production, the implementation will return an io.Reader containing stdout,
+// an error channel that will emit any errors (if present) during execution, and
+// an error if any occur while starting the command. errChan is closed after the
+// command completes, but the caller should read all of the returned io.Reader
+// before checking the error channel.
+var streamedBazelCommand = func(cmd *exec.Cmd) (io.Reader, chan error, error) {
```
aaahrens, line 22: Just adding in a comment that it would probably be best just to write into a buffer file. Would also address Carlos comments.

PR #239, opened 2021-10-05
license_manager: Add server boilerplate

minor-fixes:
This change adds:
* A LicenseManager service with all handlers returning unimplemented
* Tests for said service expecting unimplemented for all methods
* A server binary that starts the LicenseManager service on a port
  specified by flag.

Design and more info: https://docs.google.com/document/d/1TNqbBprpcNU9tTHVCFzRwaQoHlGFdjkw221C5p9UsAw/edit#

Tested: Builds and all unit tests pass:

```
bazel test //license_manager/...
```

Server starts:

```
[~/dev/enkit] (license_manager_server_boilerplate)
sminor-> bazel run //license_manager/server:license_manager -- --port=8081 --alsologtostderr &
[1] 9262
[=== 2021-10-01 05:21:39 === SUCCESS ===]

[~/dev/enkit] (license_manager_server_boilerplate)
sminor-> INFO: Invocation ID: 7742f725-9df7-43e2-96a4-49bbe64969b0
INFO: Analyzed target //license_manager/server:license_manager (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //license_manager/server:license_manager up-to-date:
  bazel-bin/license_manager/server/license_manager_/license_manager
INFO: Elapsed time: 0.254s, Critical Path: 0.06s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
I1001 05:21:39.325719    9270 main.go:32] Listening for gRPC requests on :8081...

[~/dev/enkit] (license_manager_server_boilerplate)
sminor-> grpc_cli ls -l localhost:8081
filename: reflection/grpc_reflection_v1alpha/reflection.proto
package: grpc.reflection.v1alpha;
service ServerReflection {
  rpc ServerReflectionInfo(stream grpc.reflection.v1alpha.ServerReflectionRequest) returns (stream grpc.reflection.v1alpha.ServerReflectionResponse) {}
}

filename: license_manager/proto/license_manager.proto
package: license_manager.proto;
service LicenseManager {
  rpc Allocate(license_manager.proto.AllocateRequest) returns (license_manager.proto.AllocateResponse) {}
  rpc Refresh(license_manager.proto.RefreshRequest) returns (license_manager.proto.RefreshResponse) {}
  rpc Release(license_manager.proto.ReleaseRequest) returns (license_manager.proto.ReleaseResponse) {}
  rpc LicensesStatus(license_manager.proto.LicensesStatusRequest) returns (license_manager.proto.LicensesStatusResponse) {}
}

[=== 2021-10-05 10:52:02 === SUCCESS ===]
```

Jira: INFRA-90

### DIFF
--- a/license_manager/server/main.go
+++ b/license_manager/server/main.go
@@ -1,0 +1,37 @@
+package main
+
+import (
+	"flag"
+	"net"
+	"strconv"
+
+	lmpb "github.com/enfabrica/enkit/license_manager/proto"
+	"github.com/enfabrica/enkit/license_manager/service"
+
+	"github.com/golang/glog"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+)
+
+var (
+	port = flag.Int("port", 8080, "Port for gRPC services")
+)
+
+func main (){
+	flag.Parse()
+
+	addr := net.JoinHostPort("", strconv.FormatInt(int64(*port), 10))
+	listen, err := net.Listen("tcp", addr)
+	if err != nil {
+		glog.Fatalf("Failed to listen on %q: %v", addr, err)
+	}
+	server := grpc.NewServer()
+	lmpb.RegisterLicenseManagerServer(server, &service.Service{})
+	reflection.Register(server)
+
+	glog.Infof("Listening for gRPC requests on %s...", addr)
+	err = server.Serve(listen)
+	if err != nil {
+		glog.Fatalf("Failed to serve: %v", err)
+	}
+}
--- a/license_manager/server/main.go
+++ b/license_manager/server/main.go
@@ -1,37 +1,16 @@
 package main
 
 import (
-	"flag"
-	"net"
-	"strconv"
-
+	"github.com/enfabrica/enkit/lib/server"
 	lmpb "github.com/enfabrica/enkit/license_manager/proto"
 	"github.com/enfabrica/enkit/license_manager/service"
 
-	"github.com/golang/glog"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/reflection"
 )
 
-var (
-	port = flag.Int("port", 8080, "Port for gRPC services")
-)
+func main() {
+	grpcs := grpc.NewServer()
+	lmpb.RegisterLicenseManagerServer(grpcs, &service.Service{})
 
-func main (){
-	flag.Parse()
-
-	addr := net.JoinHostPort("", strconv.FormatInt(int64(*port), 10))
-	listen, err := net.Listen("tcp", addr)
-	if err != nil {
-		glog.Fatalf("Failed to listen on %q: %v", addr, err)
-	}
-	server := grpc.NewServer()
-	lmpb.RegisterLicenseManagerServer(server, &service.Service{})
-	reflection.Register(server)
-
-	glog.Infof("Listening for gRPC requests on %s...", addr)
-	err = server.Serve(listen)
-	if err != nil {
-		glog.Fatalf("Failed to serve: %v", err)
-	}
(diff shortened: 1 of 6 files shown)


license_manager/service/service_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+func TestAllocateUnimplemented(t *testing.T) {
+	ctx := context.Background()
+	s := &Service{}
+	wantCode := codes.Unimplemented
+	req := &lmpb.AllocateRequest{}
+
+	_, err := s.Allocate(ctx, req)
+	if gotCode := status.Code(err); gotCode != wantCode {
+		t.Errorf("got code %v; want code %v", gotCode, wantCode)
```
ccontavalli, line 21: nit: don't feel strongly, but we've been using the assert library within the repo. I know there are differing opinions related to its use, but recommend using it for simple tests even if just for consistency.

license_manager/server/main.go
```diff
@@ -0,0 +1,37 @@
...
+
+func main (){
+	flag.Parse()
+
+	addr := net.JoinHostPort("", strconv.FormatInt(int64(*port), 10))
+	listen, err := net.Listen("tcp", addr)
+	if err != nil {
+		glog.Fatalf("Failed to listen on %q: %v", addr, err)
+	}
+	server := grpc.NewServer()
```
ccontavalli, line 28: FYI (don't change anything): if we were planning to run this on google app engine or outside our network, I'd recommend using our ugly lib/server lib/client, and grow them. Reason: the libraries wrap grpc around auth and grpcweb, so we can actually run apps behind non-http2 capable proxies, and can create UIs via piggybacking on grpcweb.

Why I'd leave this as is: we have no reasonable way to run the real license manager in the cloud at this point in time. The protocol is extremely fragile at best, has no authentication, the server is proprietary code that is supposed to require baremetal machines, and hard codes the MAC address in the licenses supplied by the vendor.

So, there is no need for our internal wrappers, and by using the native libraries, your code brings in fewer dependencies that can break.

If we were planning to implement a statusz, or were planning to run this in cloud run or app engine, I'd recommend using the wrappers instead.

license_manager/server/main.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"flag"
+	"net"
+	"strconv"
+
+	lmpb "github.com/enfabrica/enkit/license_manager/proto"
+	"github.com/enfabrica/enkit/license_manager/service"
+
+	"github.com/golang/glog"
```
ccontavalli, line 11: FYI: the repo is currently using the plain log library from golang, OR uber zap library. Given that libraries in the repo are supposed to be usable from any app, most often they rely on the simple lib/logger/Logger interface (but lose on semantic/field based logging, but we could add interfaces).

Here, I'd recommend using either plain log from golang, or our internal wrapper around zap library (lib/logger/klog), I would not bring in a 3rd logging library, unless it has some significant benefit.

ccontavalli (approved):

license_manager/service/service_test.go
minor-fixes: Done! It is much cleaner; thanks :)

license_manager/server/main.go
minor-fixes: No more logger since `server.Run` is used, which emits the log message.

license_manager/server/main.go
minor-fixes: Thanks for the suggestion; using the wrapper does lighten this main up quite a bit, eliminates the flag and solves the logging problem. I'll go with that until a change to our wrapper breaks this.

PR #244, opened 2021-10-08
license_manager: Implement server handlers

minor-fixes:
This change contains the bulk of the server handler implementation, along with associated unit tests.

The service proto is tweaked:

* More fields are added to `RefreshRequest` to better handle the case where the server is adopting existing allocations after a restart. The client must pass all the information (license type, owner, build_tag, etc.) during RefreshRequests in case the server has restarted; otherwise, the server won't know which license to allocate, who it is allocated to, etc.
* Common fields between RefreshRequest and AllocateRequest are deduplicated to an Invocation message.
* Repeated fields are pluralized
* Queued license count added to LicensesStats response

Jira: INFRA-90

### DIFF
--- a/license_manager/proto/license_manager.proto
+++ b/license_manager/proto/license_manager.proto
@@ -61,31 +61,8 @@ service LicenseManager {
 }
 
 message AllocateRequest {
-  // Licenses to acquire for this invocation.
-  //
-  // Invocations that need multiple of the same license type should insert
-  // duplicate entries in this field.
-  //
-  // Implementations currently only support reserving once license per
-  // invocation, though future implementations may allow for allocations of
-  // multiple and of different types.
-  repeated License licenses = 1; // required
-
-  // Owning entity issuing the allocation request. Used for logging purposes
-  // only. This could be the name of the user or system issuing the request.
-  string owner = 2; // required
-
-  // Build tag of the allocation request. This tag does not need to be unique
-  // across multiple AllocateRequests, and may be used to associate multiple
-  // such requests with one higher-level task. Typically the Bazel build
-  // invocation ID is used here.
-  string build_tag = 3; // required
-
-  // If this is not the first AllocateRequest message sent for this
-  // invocation, then this is the invocation_id in a previous AllocateResponse
-  // sent, so that this invocation can queue properly. Otherwise, this field
-  // should be left blank.
-  string invocation_id = 4;
+  // Invocation details for this allocation
+  Invocation invocation = 1; // required
 }
 
 message AllocateResponse {
@@ -126,26 +103,10 @@ message LicenseAllocated {
 }
 
 message RefreshRequest {
-  // Licenses to refresh.
-  //
-  // Implementations currently only support reserving once license per
-  // invocation, though future implementations may allow for allocations of
-  // multiple and of different types.
-  repeated License licenses = 1; // required
-
-  // Owning entity issuing the allocation request. Used for logging purposes
-  // only. This could be the name of the user or system issuing the request.
-  string owner = 2; // required
-
-  // Build tag of the allocation request. This tag does not need to be unique
-  // across multiple AllocateRequests, and may be used to associate multiple
-  // such requests with one higher-level task. Typically the Bazel build
-  // invocation ID is used here.
-  string build_tag = 3; // required
-
-  // Invocation ID returned in the Allocate() call that resulted in an
-  // allocation.
-  string invocation_id = 4; // required
+  // Existing invocation to refresh. Must have been allocated by a call to
+  // Allocate() successfully (not queued).
+  // Must have invocation_id set.
+  Invocation invocation = 1; // required
 }
 
 message RefreshResponse {
@@ -197,6 +158,37 @@ message LicenseStats {
   // LicenseManager. If this number is >0, then allocated_count should equal
   // total_license_count.
   uint32 queued_count = 6;
+}
+
+message Invocation {
+  // Licenses to acquire for this invocation.
+  //
+  // Invocations that need multiple of the same license type should insert
+  // duplicate entries in this field.
+  //
+  // Implementations currently only support reserving once license per
+  // invocation, though future implementations may allow for allocations of
+  // multiple and of different types.
+  repeated License licenses = 1; // required
+
+  // Owning entity issuing the allocation request. Used for logging purposes
+  // only. This could be the name of the user or system issuing the request.
+  // This must be sent on every Allocate() and Refresh() call in case the
+  // server is restarted.
+  string owner = 2; // required
+
+  // Build tag of the allocation request. This tag does not need to be unique
+  // across multiple AllocateRequests, and may be used to associate multiple
+  // such requests with one higher-level task. Typically the Bazel build
+  // invocation ID is used here. This must be sent on every Allocate() and
+  // Refresh() call in case the server is restarted.
+  string build_tag = 3; // required
+
+  // The server-generated ID for this invocation. In the initial Allocate()
+  // call, this can be blank for the initial request (to receive a generated
+  // one in the response) but subsequent Allocate() calls to refresh a queue
+  // position or Refresh() calls should have this field set.
+  string id = 4;
 }
 
 message License {
--- a/license_manager/proto/license_manager.proto
+++ b/license_manager/proto/license_manager.proto
@@ -126,9 +126,26 @@ message LicenseAllocated {
 }
 
 message RefreshRequest {
-  // Opaque identifier for this invocation, as returned by the server in the
-  // AllocateResponse that allocated this license.
-  string invocation_id = 1; // required
+  // Licenses to refresh.
+  //
+  // Implementations currently only support reserving once license per
+  // invocation, though future implementations may allow for allocations of
+  // multiple and of different types.
+  repeated License licenses = 1; // required
+
+  // Owning entity issuing the allocation request. Used for logging purposes
+  // only. This could be the name of the user or system issuing the request.
+  string owner = 2; // required
+
+  // Build tag of the allocation request. This tag does not need to be unique
+  // across multiple AllocateRequests, and may be used to associate multiple
+  // such requests with one higher-level task. Typically the Bazel build
+  // invocation ID is used here.
+  string build_tag = 3; // required
+
+  // Invocation ID returned in the Allocate() call that resulted in an
+  // allocation.
+  string invocation_id = 4; // required
 }
 
 message RefreshResponse {
@@ -175,6 +192,11 @@ message LicenseStats {
   // Number of licenses currently allocated, as reported by the
   // LicenseManager. This should be less than or equal to in_use_count.
   uint32 allocated_count = 5;
+
+  // Number of invocations queued for a license, as reported by the
+  // LicenseManager. If this number is >0, then allocated_count should equal
+  // total_license_count.
+  uint32 queued_count = 6;
 }
 
 message License {
--- a/license_manager/proto/license_manager.proto
+++ b/license_manager/proto/license_manager.proto
@@ -69,7 +69,7 @@ message AllocateRequest {
   // Implementations currently only support reserving once license per
   // invocation, though future implementations may allow for allocations of
   // multiple and of different types.
-  repeated License license = 1; // required
+  repeated License licenses = 1; // required
 
   // Owning entity issuing the allocation request. Used for logging purposes
   // only. This could be the name of the user or system issuing the request.
--- a/license_manager/service/license.go
+++ b/license_manager/service/license.go
@@ -53,11 +53,7 @@ func (l *license) Promote() {
 // GetAllocated returns an invocation by ID if the invocation is allocated a
 // license, or nil otherwise.
 func (l *license) GetAllocated(invID string) *invocation {
-	inv, ok := l.allocations[invID]
-	if !ok {
-		return nil
-	}
-	return inv
+	return l.allocations[invID]
 }
 
 // ExpireAllocations removes all allocations for invocations that have not
@@ -118,7 +114,6 @@ func (l *license) GetStats(name string) *lmpb.LicenseStats {
 func (l *license) Forget(invID string) int {
 	count := 0
 	newAllocations := map[string]*invocation{}
-	newQueue := []*invocation{}
 	for k, v := range l.allocations {
 		if k != invID {
 			newAllocations[k] = v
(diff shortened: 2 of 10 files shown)


license_manager/proto/license_manager.proto
```diff
@@ -126,9 +126,26 @@ message LicenseAllocated {
 }
 
 message RefreshRequest {
-  // Opaque identifier for this invocation, as returned by the server in the
-  // AllocateResponse that allocated this license.
-  string invocation_id = 1; // required
+  // Licenses to refresh.
```
ccontavalli, line 129: uhm ... RefreshRequest and AllocateRequest now are equal? Why not have a single message, then? Semantics also seem pretty much the same?

license_manager/proto/license_manager.proto
```diff
@@ -175,6 +192,11 @@ message LicenseStats {
   // Number of licenses currently allocated, as reported by the
   // LicenseManager. This should be less than or equal to in_use_count.
   uint32 allocated_count = 5;
+
+  // Number of invocations queued for a license, as reported by the
+  // LicenseManager. If this number is >0, then allocated_count should equal
+  // total_license_count.
+  uint32 queued_count = 6;
```
ccontavalli, line 199: I would consider returning the queue itself, <invocation id, owner, build tag> ... so the tool can print the queue itself. If this queue ever grows to more than O(10)s of items, we are in trouble, so I suspect returning the array to be just fine?

license_manager/service/license.go
```diff
@@ -0,0 +1,141 @@
...
+		l.allocations[l.queue[i].ID] = l.queue[i]
+		numAllocated++
+	}
+	l.queue = l.queue[numAllocated:]
+}
+
+// GetAllocated returns an invocation by ID if the invocation is allocated a
+// license, or nil otherwise.
+func (l *license) GetAllocated(invID string) *invocation {
+	inv, ok := l.allocations[invID]
```
ccontavalli, line 56: I thought that go maps, if used in the one return value form, returned the zero value for the type by default in case it was not found? Nil for pointers.  Can this be simplified with:
```
return l.allocations[invID]
```
?

license_manager/service/license.go
```diff
@@ -0,0 +1,141 @@
...
+		QueuedCount:       uint32(len(l.queue)),
+	}
+}
+
+// Forget removes invocations matching the specified ID from allocations and
+// the queue.
+func (l *license) Forget(invID string) int {
+	count := 0
+	newAllocations := map[string]*invocation{}
+	newQueue := []*invocation{}
```
ccontavalli, line 121: nit: I'd move newQueue next to the following for loop.

license_manager/service/license.go
```diff
@@ -0,0 +1,141 @@
...
+
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// license manages allocations and queued invocations for a single license
+// type.
+type license struct {
+	totalAvailable int                    // Constant total number of licenses available for invocations.
+	queue          []*invocation          // List of invocations waiting for a license, in FIFO order.
+	allocations    map[string]*invocation // Map of invocation ID to invocation data for an allocated license.
```
ccontavalli, line 17: code looks good, so no required change: but while reviewing the code, I was wondering if it made sense to have a single `map[string]*invocation` containing both allocated and non allocated invocations.

The invocation struct then gains a `prev, next *invocation` pointer, and the `license` struct gains a `first, last *invocation`. If the invocation is not allocated, then prev and next link it into a linked list starting from first. If the invocation is allocated, then either those two pointers are nil, or there's some other bit indicating it is allocated.

license_manager/service/service.go
```diff
@@ -2,36 +2,295 @@ package service
...
+			"xilinx::foo": &license{
+				totalAvailable: 2,
+				allocations:    map[string]*invocation{},
+			},
+			"xilinx::bar": &license{
+				totalAvailable: 2,
+				allocations:    map[string]*invocation{},
+			},
+		},
+		queueRefreshDuration:      5 * time.Second,
```
ccontavalli, line 40: // TODO: those should come from flags.

license_manager/service/service.go
```diff
@@ -2,36 +2,295 @@ package service
...
+		allocationRefreshDuration: 5 * time.Second,
+	}
+
+	go func(s *Service) {
+		t := time.NewTicker(1 * time.Second)
```
ccontavalli, line 45: `1 * time.Second` should probably also come from flag.

license_manager/service/service.go
```diff
@@ -2,36 +2,295 @@ package service
...
+		defer t.Stop()
+		for {
+			<-t.C
+			s.janitor()
+		}
+	}(service)
+
+	go func(s *Service) {
+		<-time.After(10 * time.Second)
```
ccontavalli, line 54: nit: same for the 10 seconds.

license_manager/service/service.go
```diff
@@ -2,36 +2,295 @@ package service
...
+		}
+		inv := &invocation{
+			ID:          invocationID,
+			Owner:       req.GetOwner(),
+			BuildTag:    req.GetBuildTag(),
+			LastCheckin: timeNow(),
+		}
+		lic.Enqueue(inv)
+
+		switch s.currentState {
```
ccontavalli, line 150: nit: here, I'd personally just use:
```
if s.currentState == stateRunning {
  lic.Promote()
}
```

license_manager/service/service.go
```diff
@@ -2,36 +2,295 @@ package service
...
+		case stateRunning:
+			lic.Promote()
+		default:
+			return nil, status.Errorf(codes.Internal, "unhandled state: %v", s.currentState)
+		}
+	}
+
+	// Invocation ID should now be known and either queued (due to the above
+	// insert, or from a previous request) or allocated (promoted from the queue
+	// by above, or asynchronously by the janitor).
```
ccontavalli, line 161: "promoted from the queue by above" ? where? Looks like GetAllocated won't succeed until the next janitor run? Unless it's a new allocation?

So here it will queue, and then 1second + refresh interval later it will get going? Probably ok if it only happens at restart, but from comment, it does not sound like the desired behavior?

license_manager/service/service.go
```diff
@@ -2,36 +2,295 @@ package service
...
+			ResponseType: &lmpb.AllocateResponse_Queued{
+				Queued: &lmpb.Queued{
+					InvocationId: invocationID,
+					NextPollTime: timestamppb.New(timeNow().Add(s.queueRefreshDuration)),
+				},
+			},
+		}, nil
+	}
+	// Invocation is not allocated or queued
+	switch s.currentState {
```
ccontavalli, line 188: nit: I'd personally also use an if for readability:

```
if s.currentState != stateStarting {
  return nil, status.Errorf(...)
}

inv := &invocation{
...
}
lic.Enqueue(ivn)
...
```

license_manager/service/service.go
```diff
@@ -2,36 +2,295 @@ package service
...
+
+	if len(req.GetLicenses()) != 1 {
+		return nil, status.Errorf(codes.InvalidArgument, "licenses must have exactly one license spec")
+	}
+	licenseType := formatLicenseType(req.GetLicenses()[0])
+	lic, ok := s.licenses[licenseType]
+	if !ok {
+		return nil, status.Errorf(codes.NotFound, "unknown license type: %q", licenseType)
+	}
+	invID := req.GetInvocationId()
```
ccontavalli, line 229: apart from locking, first ~10ish lines are common between Allocate and Refresh? Would it make sense to have a small helper? not sure myself, but maybe some of the "request validation" should be the same?

license_manager/service/service.go
```diff
@@ -2,36 +2,295 @@ package service
...
+	if invID == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "invocation_id must be set")
+	}
+	inv, ok := lic.allocations[invID]
```
ccontavalli, line 233: GetAllocated?

license_manager/service/license.go
```diff
@@ -0,0 +1,141 @@
...
+
+	lmpb "github.com/enfabrica/enkit/license_manager/proto"
+
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// license manages allocations and queued invocations for a single license
+// type.
+type license struct {
+	totalAvailable int                    // Constant total number of licenses available for invocations.
```
ahungrynacho, line 15: These are the license counts:

1. Xilinx: 3 Vivado tool licenses
2. Cadence: 6 Xcelium tool licenses, 5 Jasper tool licenses
3. Synopsys: 1 Fusion Compiler tool license, 2 VC static tool licenses, 2 Spyglass tool licenses, 1 VCS tool license, 1 Verdi tool license

Different EDA tool vendors contain a different number of license counts per tool.

license_manager/proto/license_manager.proto
minor-fixes: I realized the same when I went to implement it - didn't want to change the proto any more than necessary to implement what was already approved. Future PR ok?

license_manager/service/license.go
minor-fixes: Sounds like that can work as well - I'd have to think more about which is better. This is more straightforward to me (obviously because it's the first thing my mind jumped to) mainly because the data structures are more concrete (instead of storing them all in a map and then constructing a virtual organization using pointers); I don't think our numbers are high enough to worry about e.g. queue resizing over simplicity of implementation.

license_manager/service/service.go
minor-fixes: Added TODO

license_manager/service/service.go
minor-fixes: Added TODO

license_manager/service/service.go
minor-fixes: Added TODO

license_manager/service/service.go
```diff
@@ -2,36 +2,295 @@ package service
...
+		defer s.mu.Unlock()
+		s.currentState = stateRunning
+	}(service)
+
+	return service
+}
+
+// invocation maps to a particular command invocation that has requested a
+// license, and its associated metadata.
+type invocation struct {
```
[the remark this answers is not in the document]
minor-fixes: Licenses hold invocations (in the allocations and queue data structures) so presumably we can log this information from that level without duplicating data too much; admittedly, there are no logging statements in the current code.

license_manager/service/service.go
minor-fixes: > "promoted from the queue by above" ? where?

Line 153 (approximately) above: `lic.Promote()`

> Looks like GetAllocated won't succeed until the next janitor run?

If this was a new invocation, the above enqueued it and then synchronously called Promote on just that license to move everything it can from the head of the queue to allocated. So it is now either successfully allocated or still queued (because there were insufficient free slots to move it from the queue)

license_manager/service/service.go
minor-fixes: Done

license_manager/service/service.go
```diff
@@ -2,36 +2,295 @@ package service
...
+	if !ok {
+		return nil, status.Errorf(codes.NotFound, "unknown license type: %q", licenseType)
+	}
+	invID := req.GetInvocationId()
+	if invID == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "invocation_id must be set")
+	}
+	inv, ok := lic.allocations[invID]
+	if !ok {
+		switch s.currentState {
```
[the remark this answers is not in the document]
minor-fixes: Done

license_manager/service/service.go
minor-fixes: Done; removed all the switches.

license_manager/proto/license_manager.proto
minor-fixes: Exploring this in a future PR

license_manager/service/service.go
minor-fixes: This will disappear if we consolidate Refresh and Allocate.

license_manager/proto/license_manager.proto
ccontavalli: Should we have 2 different RPCs using the same message, though?

license_manager/proto/license_manager.proto
ccontavalli: SGTM

license_manager/service/license.go
ccontavalli: Sorry for the confusion. Let me clarify my comment.

There are two bits of information:
1) How many licenses we have available in total, and which kind.
2) How many licenses are allocated.

I still stand that by my original statement that allocation/deallocation should be out of band. Monitoring will tell us if there are discrepancies, and we'll react to that.

Several reasons for this, but the main one is that through a single build, some builds allocate/release the license in the middle of the build step. But if a license is not available in the middle of the step, the step can crash/fail. 

So, ... let's say you query the license server in the middle of such a build, it will likely say "this client is using 0 licenses", but in practice, 1 license MUST be reserved for the whole duration of the step. Our license server does the reservation.

That said... the TOTAL number of license AVAILABLE, the licenses we paid for, should be a pretty constant number, and both our "reservation system" and "license manager" need to somewhat agree on that number.

So far, the # of TOTAL licenses was hardcoded in the binary. Having a config file (or a built in config file, like we do for other binaries), is a good plan.

The annoying part is that whenever licenses expire, or we need to buy new ones, we need to update two places: the real license manager, and our own. Which is error prone.

So my thought was really to get the TOTAL # of licenses (at startup, or periodically) from the license server, so we don't have to maintain this second configuration file, and we can no longer make mistakes here.

license_manager/service/license.go
ccontavalli: SGTM

license_manager/service/service.go
ccontavalli: SGTM

ccontavalli (approved):

license_manager/service/license.go
minor-fixes: Thanks for the clarification - that makes sense.

I think overall that would be a good quality-of-life improvement - we can add that as that becomes the biggest issue? I think right now it's still an open question whether the current approach can result in something stable given significant build load as-is.

ccontavalli: sure, sgtm, yes. - future PR makes sense.

license_manager/proto/license_manager.proto
minor-fixes: Changed to 2 RPCs using the same message

PR #249, opened 2021-10-13
Adds in Machinist Changes for _all record refresh duration

aaahrens:
- Adds in the ability to specify how often to update the record _all

### DIFF
--- a/machinist/mserver/controller.go
+++ b/machinist/mserver/controller.go
@@ -163,7 +163,7 @@ func (en *Controller) ServeAllRecords() {
 			}
 			en.dnsServer.SetEntry(dnsName, rs)
 		}
-		_ = <-time.After(5 * time.Second)
+		_ = <-time.After(en.allRecordsRefreshRate)
 	}
 }
 
--- a/machinist/mserver/controller.go
+++ b/machinist/mserver/controller.go
@@ -18,17 +18,18 @@ type Controller struct {
 
 	startUpFunc []func()
 
+	allRecordsRefreshRate time.Duration
 
 	State         *state.MachineController
 	stateFile     string
 	stateWriteTTL time.Duration
 
 	dnsServer *kdns.DnsServer
-	dnsPort   int
 	domains   []string
 }
+
 // Init is designed to run after all components have been started up before running itself as a server
-func (en *Controller) Init()  {
+func (en *Controller) Init() {
 	for _, m := range en.State.Machines {
 		en.addNodeToDns(m.Name, m.Ips, m.Tags)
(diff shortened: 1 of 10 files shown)


machinist/mserver/mserver.go
```diff
@@ -45,12 +47,14 @@ func (s *ControlPlane) Run() error {
...
 	s.Controller.Init()
-	go s.Controller.ServeAllRecords()
+	go s.Controller.ServeAllRecords(s.allRecordsKillChannel)
 	go s.Controller.WriteState()
 	return grpcs.Serve(s.Listener)
 }
 
 func (s *ControlPlane) Stop() error {
+	s.allRecordsKillChannel <- struct{}{}
+	<-s.allRecordsKillChannel
```
minor-fixes, line 57: I think this should be two separate channels unfortunately; I don't think there's anything that stops the receive from picking up the kill signal inserted on the previous line.

machinist/mserver/controller.go
```diff
@@ -146,24 +147,29 @@ func (en *Controller) addNodeToDns(name string, ips []net.IP, tags []string) {
...
-		for _, d := range en.dnsServer.Domains {
-			dnsName := dns.CanonicalName(fmt.Sprintf("%s.%s", "_all", d))
-			var rs []dns.RR
-			for _, v := range ns {
-				for _, i := range v.Ips {
-					rr, err := dns.NewRR(fmt.Sprintf("%s %s %s", dnsName, "A", i.String()))
-					if err != nil {
-						en.Log.Errorf("err: %v", err)
+		select {
+		case <-time.After(en.allRecordsRefreshRate):
```
minor-fixes, line 153: This works, but consider instead using a time.Ticker for two reasons:

* the interval time is even if the loop body takes a long time to run relative to the refresh rate (currently the actual interval is `allRecordsRefreshRate` + the execution time of the loop body)
* bigger reason: the ticker can be (probably wrapped with some sort of interface, and then) passed in so that this can be tested without relying on timing. In prod, pass an actual time.Ticker, but in tests, pass a ticker that requires a call to emit each tick, so you can tick, assert state, tick, assert new state, etc.

minor-fixes:
Can you add some information on how this was tested? I'm curious to know repro steps for the current code (do we get a failure by running the current tests with the race detector, for instance?) as well as how we know the changes fix the failure

aaahrens:
It wasn't tested before (well it is in the next PR) because this was a day 0 request add when the dns launched. I added a test in this PR.

minor-fixes (approved):

PR #251, opened 2021-10-13
Adds in a stress test for the machinist dns. Tests TCP activation as well as large record locking, 

aaahrens:
- Adds in a large test for stressing dns server
- Mainly to check for large records and that there are no deadlocks in the controller.
- Moves AuthClient connect to entroll() directly


lib/knetwork/kdns/dns.go
```diff
@@ -119,6 +120,9 @@ func (s *DnsServer) HandleControllers() {
 	controllerMap := map[string]*RecordController{}
 	defer close(s.readOnlyChan)
 	defer close(s.newOrExistingChan)
+	defer func() {
+		fmt.Println("closing controllers??", s.Domains)
```
minor-fixes, line 124: Maybe log instead of printing directly to stdout? `DnsServer` has a logger

Also, defers run after the function returns. If I'm reading this right - the controllers are closed on line 137, no? Maybe this should log that the controllers have already been closed

machinist/testing/stress_e2e_test.go
```diff
@@ -0,0 +1,103 @@
...
+	assert.Equal(t, NumMachines, len(r.Answer))
+	time.Sleep(50 * time.Millisecond)
+}
+
+// just makes a bunch of ips that arent the same from each other
+func createIps(max int) []net.IP {
+	toRet := make([]net.IP, max)
+	index := 0
+	for first := 0; first < 256; first++ {
+		for second := 0; second < 256; second++ {
```
minor-fixes, line 89: How about fixing the first byte at 10? This ensures that none of the IPs generated are externally routable (in case something seriously weird happens with the test) and still gives you some 16M IPs to play with

machinist/testing/stress_e2e_test.go
```diff
@@ -0,0 +1,103 @@
...
+			machine.WithTags([]string{fmt.Sprintf("number%d", i)}),
+			machine.WithMachinistFlags(
+				config.WithEnableMetrics(false),
+				config.WithControlPlaneHost("127.0.0.1"),
+				config.WithControlPlanePort(grpcAddr.Port),
+			),
+		})
+	}
+	time.Sleep(5 * time.Second)
+	assert.NoError(t, err)
```
minor-fixes, line 70: Which `err` does this assertion test?

machinist/testing/stress_e2e_test.go
```diff
@@ -0,0 +1,103 @@
...
+	c := dns.Client{
```
minor-fixes, line 71: gofmt?

PR #255, opened 2021-10-20
bazel: Parse workspace_log during query

minor-fixes:
This change modifies Query to return a struct of query results. These
results include targets that match the query, as well as other optional
information. The first useful such piece of information is the workspace
log, which details when:

* WORKSPACE caused a download to happen, including the SHA256 of the
  download
* WORKSPACE caused a command execution to happen

among other events.

These events will be useful in reducing external dependencies down to
the SHA256s of their corresponding downloads, rather than needing to
resort to hashing individual files of the external dependencies.

Depends-On: #252


lib/bazel/affected_targets.go
```diff
@@ -53,19 +53,32 @@ func GetAffectedTargets(start string, end string) ([]string, error) {
...
 		return nil, fmt.Errorf("failed to query deps for start point: %w", err)
 	}
 
-	targets, err = endWorkspace.Query("deps(//...)", WithKeepGoing(), WithUnorderedOutput())
+	endResults, err := endWorkspace.Query("deps(//...)", WithKeepGoing(), WithUnorderedOutput(), workspaceLogEnd)
 	if err != nil {
 		return nil, fmt.Errorf("failed to query deps for end point: %w", err)
 	}
 
+	fmt.Fprintf(os.Stderr, "Before results:\tTargets: %d\tWorkspace events: %d\n", len(startResults.Targets), len(startResults.WorkspaceEvents))
```
aaahrens, line 76: Why is this printing to stderr? is the command expected to be run with &2>1?

lib/bazel/affected_targets.go
```diff
@@ -53,19 +53,32 @@ func GetAffectedTargets(start string, end string) ([]string, error) {
...
+	fmt.Fprintf(os.Stderr, "After  results:\tTargets: %d\tWorkspace events: %d\n", len(endResults.Targets), len(endResults.WorkspaceEvents))
```
aaahrens, line 77: Same as above

lib/bazel/options.go
```diff
@@ -62,6 +69,11 @@ func (o *queryOptions) Args() []string {
 	if o.unorderedOutput {
 		f = append(f, "--order_output=no")
 	}
+	if o.workspaceLog != nil {
```
aaahrens, line 72: it looks like the workspace log file is required, maybe just error out if it wasn't passed?

lib/bazel/options.go
```diff
@@ -73,17 +85,27 @@ func (o *queryOptions) filterError(err error) error {
 		return nil
 	}
 
-	if err, ok := err.(*exec.ExitError); ok {
+	var execErr *exec.ExitError
+	if errors.As(err, &execErr) {
 		// PARTIAL_ANALYSIS_FAILURE is expected when --keep_going is passed
 		// https://github.com/bazelbuild/bazel/blob/86409b7a248d1cb966268451f9aa4db0763c3eb2/src/main/java/com/google/devtools/build/lib/util/ExitCode.java#L38
-		if err.ExitCode() == 3 {
+		if execErr.ExitCode() == 3 {
```
aaahrens, line 92: Why is only error code 3 expected here? Does bazel return other error codes while querying that mean other things?

lib/bazel/options.go
minor-fixes: Shouldn't be - the workspace log is only parsed if the flag is passed?

lib/bazel/options.go
minor-fixes: Yes; see the URL in the comment for all of bazel's return codes and their various meanings.

lib/bazel/affected_targets.go
minor-fixes: Done

lib/bazel/query.go
```diff
@@ -40,19 +41,25 @@ var streamedBazelCommand = func(cmd *exec.Cmd) (io.Reader, chan error, error) {
...
 		if err != nil {
-			errChan <- err // Don't wrap, so raw ExitError can be picked up by the caller
+			errChan <- fmt.Errorf("command failed: `%s`: %w", strings.Join(cmd.Args, " "), err)
 		}
 	}()
 
 	return pipeReader, errChan, nil
 }
 
+type QueryResult struct {
```
aaahrens, line 51: So what I had mean by the log file being required: if the workspace log file is not passed in, then the workevents property is an empty list even if the targets property is not.  I think that's counter intuitive?

Maybe either document that for right now, or change the behaviour.

lib/bazel/options.go
aaahrens: See below review

PR #257, opened 2021-10-20
go: Update toolchain, dependencies

minor-fixes:
This change updates:

* the Go toolchain, to version 1.17.1
* rules_go, to the latest released version
* gazelle, to the latest released version
* rules_proto, to the latest released version, to fix compile issues
  caused by updating rules_go
* required bazel version/version downloaded by bazelisk to 4.2.0, the
  minimum required for the rules_go update

Tested: `bazel build //...` and `bazel test //...` both pass


bazel/init.bzl
```diff
@@ -9,7 +9,7 @@ load("//bazel/ui:deps.bzl", "install_ui_deps")
 
 def enkit_init_go():
     go_rules_dependencies()
-    go_register_toolchains(version = "1.16")
+    go_register_toolchains(version = "1.17.1")
```
aaahrens, line 12: So I specifically made it 1.16 because that's the most supported version of app engine. It should probably remain there until app engine supports it.

minor-fixes: Do you have a source? The docs I found suggest not even 1.16 is supported: https://cloud.google.com/appengine/docs/flexible/go/runtime#go_versions

minor-fixes: Reverted the go toolchain update - it's not necessary; I was just following the rules_go docs for their release.

PR #292, opened 2021-10-29
INFRA-121: Fix bazel graph parsing for directory labels

minor-fixes:
Some labels in `internal` are actually directories. When this is the
case, these directory labels are used as commandline flags rather than
file contents. It is debatable if this should just be fixed in
`internal`, but in the meantime dependency detection can tolerate this
case.

When a "source file" label that actually represents a directory is
encountered, the hash of the "source file" should correspond simply to
the hash of the label itself. This resolves a "can't read: is a
directory"-type error when attempting to hash the underlying "file" for
a directory source file target.

Jira: INFRA-121

### DIFF
--- a/lib/bazel/affected_targets_test.go
+++ b/lib/bazel/affected_targets_test.go
@@ -16,6 +16,7 @@ func testWorkspace(t *testing.T) *Workspace {
 	files := map[string][]byte{
 		"test/target/foo.txt":          []byte("Hello, world"),
 		"test/target/foo_modified.txt": []byte("Goodbye, world"),
+		"test/anotherdir/file.txt":     []byte("Another dir"),
 	}
 	sourceFS := testutil.NewFS(t, files)
 	genFS := testutil.NewFS(t, nil)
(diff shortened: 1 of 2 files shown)


minor-fixes:
Draft until:

- [ ] #292 is merged
- [ ] this is rebased onto master

aaahrens (approved):

ccontavalli:
> When a "source file" label that actually represents a directory is encountered, the hash of the "source file" should correspond simply to the hash of the label itself. This resolves a "can't read: is a directory"-type error when attempting to hash the underlying "file" for a directory source file target.

I should have commented on this earlier, but I don't think using the hash of the label itself is correct. Eg, let's say a file in the directory changed: the hash of the label will be the same. While the hash should have changed.

I was under the impression that when facing directories with eg, `ctx.actions.declare_directory()` or as input dependencies, bazel would use some form of "hash the content of the directory" to determine if it's the same or has changed? eg, hash it all?

(how do you otherwise detect changes in content of directories so they trigger another run of corresponding tests?)

PR #299, opened 2021-11-05
bazel presubmit: Set CARGO_HOME on all invocations

minor-fixes:
Packages fetched via Cargo by bazel race on the $CARGO_HOME directory,
which has a known global default, when multiple bazel invocations are
in-flight at once.

This change sets $CARGO_HOME to a tempdir on each invocation, so that
invocations never collide here.

Tested: Fixes flakiness issue in internal presubmit - goes from failing
50% to 0% out of 8 runs.

Jira: INFRA-140

### DIFF
--- a/lib/bazel/exec.go
+++ b/lib/bazel/exec.go
@@ -39,9 +39,10 @@ type Command interface {
 // fileCommand buffers stdout and stderr from the underlying command to a
 // temporary file.
 type fileCommand struct {
-	cmd        *exec.Cmd
-	stdoutPath string
-	stderrPath string
+	cmd           *exec.Cmd
+	stdoutPath    string
+	stderrPath    string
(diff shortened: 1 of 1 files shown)


aaahrens (approved):

lib/bazel/exec.go
```diff
@@ -57,10 +58,25 @@ var NewCommand = func(cmd *exec.Cmd) (Command, error) {
 		return nil, fmt.Errorf("failed to create stderr file: %w", err)
 	}
 	stderr.Close()
+
+	// BUG(INFRA-140) - By default, cargo will download packages to a well-known
```
ccontavalli, line 62: does this affect local builds as well? Should we change it in the bazel rules themselves rather than in the runner?

PR #302, opened 2021-11-09
flextape: Add metrics

minor-fixes:
This change adds metrics that allow us to tell:

* Number of invocations in-progress (allocated a license) per license
  type
* Number of invocations queued per license type
* Server-side execution time per RPC per returned gRPC status code
* Count of RPC executions per RPC per status code
* Janitor execution time

This change massages Go dependency loading order to ensure that our
go.mod repositories get loaded first, to avoid surprises where
rules_go/gazelle and friends load an older version overriding our own.

Tested:

Started flextape server locally, ran 10 clients in a loop with:

```
for i in $(seq 1 10); do (LICENSE_SERVER_IMPL=flextape bazel-bin/manager/client/client_/client localhost 6433 xilinx bar sleep 60 &); sleep 1; done
```

Saw metrics in local Prometheus instance:

flextape_metrics

flextape_metrics_2
9

Jira: INFRA-10


ahungrynacho (approved):

minor-fixes:
Targets the presubmit fails on (including `//astore/server:css`) are broken on master, so submitting despite the failure.

PR #307, opened 2021-11-10
Create Websocket Connection Pool

aaahrens:
- Makes a websocket connection pool
- Makes a threadsafe wrapper for reading and writing around websockets
- Create a type that determines how the pool can find it's clients.

### DIFF
--- a/proxy/enfuse/websocket.go
+++ b/proxy/enfuse/websocket.go
@@ -26,11 +26,6 @@ func (w *WebsocketLocker) WriteMessage(messageType int, payload []byte) error {
 	return w.c.WriteMessage(messageType, payload)
 }
 
-// Raw exists just in case something needs to be inspected on the underlying connection. Use at own risk.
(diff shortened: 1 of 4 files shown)


proxy/enfuse/websocket.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	uid := data[:scp.prefixLen]
+	if scp.Fetch(uid) == nil {
+		scp.websocketMap[string(uid)] = NewWebsocketLock(conn)
+	}
+	return scp.srvWebsocket.WriteMessage(msgType, data)
+}
+
+func (scp *WebsocketPool) ServerPresent() bool {
+	return scp.srvWebsocket != nil
```
minor-fixes, line 76: Lock?

minor-fixes (approved):

proxy/enfuse/websocket.go
aaahrens: Added in race detection and in test also

PR #308, opened 2021-11-10
Socket shims 

aaahrens:
- Depends on #307 
- Create two socket shims. 
- One a ReadWriter that writes and translates the uuid strategy of the payloads
- Another is a websocket half duplex to a  tcp listener. Both are designed to run locally.


proxy/enfuse/socket.go
```diff
@@ -0,0 +1,117 @@
...
+			w.l.Warnf("Error reading from websocket %v", err)
+			continue
+		}
+		uid := data[:w.prefixLen]
+		remainingData := data[w.prefixLen:]
+		clientConn := w.clientMap[string(uid)]
+		if clientConn == nil {
+			clientConn, err = net.Dial(w.netConn.Addr().Network(), w.netConn.Addr().String())
+			if err != nil {
+				w.l.Warnf("Error dialing to existing listener %v", err)
```
minor-fixes, line 90: Why warnings not errors? This seems like the kind of thing that is definitely broken, that will definitely result in failure down the line -> error? I like to reserve warnings for "things that may or may not be wrong, depending on the circumstances" or deprecations, etc.

proxy/enfuse/socket.go
```diff
@@ -0,0 +1,117 @@
...
+		if err != nil {
+			w.l.Warnf("Error writing to client %v", err)
+			continue
+		}
+	}
+}
+
+// handleRead will read from the net.Conn and write the full payload to webConn
+func (w *WebsocketTCPShim) handleReadToWebsocket(c net.Conn, uid []byte) {
+	go func() {
```
minor-fixes, line 106: This is launched via goroutine on line 94. Should this function be launched async via `go`, or launched sync and start its own goroutine in the call? (I prefer the former slightly)

proxy/enfuse/socket.go
```diff
@@ -0,0 +1,117 @@
...
+		if _, err := io.Copy(SocketShim{WebConn: w.webConn, Prefix: uid, PrefixLen: len(uid)}, c); err != nil {
+			fmt.Println(err, "err copying for client")
```
minor-fixes, line 108: log error?

proxy/enfuse/socket.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+	}()
+}
+
+//TODO(adam): decide whether or not it's even necessary to have websocketTCPShim be a struct.
+//entirely possible for it to just be a function
```
minor-fixes, line 114: I guess make a decision now and remove this TODO

proxy/enfuse/socket.go
```diff
@@ -0,0 +1,117 @@
...
+func (w *WebsocketTCPShim) Close() error {
+	return nil
```
minor-fixes, line 116: Does this need to close the stuff it wraps?

proxy/enfuse/socket.go
```diff
@@ -0,0 +1,117 @@
...
+	return len(realData[s.PrefixLen:]), io.EOF
+}
+
+func (s SocketShim) Write(p []byte) (n int, err error) {
+	// websockets always write the full buffer
+	realPayload := append(s.Prefix, p...)
+	return len(p), s.WebConn.WriteMessage(websocket.BinaryMessage, realPayload)
+}
+
+func NewSocketShim(start PayloadAppendStrategy, Conn *websocket.Conn) (*SocketShim, error) {
```
minor-fixes, line 41: `strat`?

Maybe `strategy` would be a little easier to visually parse/disambiguate from `start`

proxy/enfuse/socket.go
```diff
@@ -0,0 +1,117 @@
...
+	pLen, f := start()
+	pre, err := f()
```
minor-fixes, line 43: How come `strat` couldn't have returned `pre` and `err` directly - why is that behind another function call?

proxy/enfuse/socket.go
aaahrens: I liek handling io.Copy() exits on their won goroutines, so that's why I settled on the former

proxy/enfuse/socket.go
aaahrens: Yes, this should be an error.

proxy/enfuse/socket.go
aaahrens: yes should be errorf

proxy/enfuse/socket.go
aaahrens: removed

proxy/enfuse/socket.go
aaahrens: It's necessary for readwrite closer interface, so It' a NoOp.  But it dones't need to close it's resources

proxy/enfuse/socket.go
aaahrens: The idea is that the Payload strategies return back the information about the strategy, and then the factory to generate that information. Right now it's an int, but in the future will probably be a struct

proxy/enfuse/socket.go
aaahrens: I don't have an issue with this, refactored

proxy/enfuse/socket.go
```diff
@@ -105,13 +104,11 @@ func (w *WebsocketTCPShim) handleWrites() {
 func (w *WebsocketTCPShim) handleReadToWebsocket(c net.Conn, uid []byte) {
 	go func() {
 		if _, err := io.Copy(SocketShim{WebConn: w.webConn, Prefix: uid, PrefixLen: len(uid)}, c); err != nil {
-			fmt.Println(err, "err copying for client")
+			w.l.Errorf("err copying for client %v", err)
```
minor-fixes, line 107: Make sure Errorf's use `%w` everywhere

minor-fixes (approved):

PR #309, opened 2021-11-11
flextape: Add config definition and Xilinx config

minor-fixes:
This change adds:

* A protobuf config definition, for a config file passed to the server
  on startup
* A flag on the server to provide the config file at runtime
* Server code to parse the config

Jira: INFRA-159


ahungrynacho (approved):
Looks good. We will worry about Cadence and Synopsys license queuing with flextape later because those EDA tools natively support the `-licqueue` flag which is just a basic FIFO queue according to their documentation. Any sort of advanced priority queuing beyond just a simple FIFO will require Cadence and Synopsys to be integrated with flextape.

PR #315, opened 2021-11-17
gee: add codeowners command

jonathan-enf:
gee: add a "codeowners" command to display detailed information about what set
of reviewers are required to approve the changes in this branch.  For example:

  jonathan-enf

A more complicated PR might say something like:

This PR requires approval from at least one reviewer from each line:
  jonathan-enf
  tao carlo abhishek
  ben ahungrynacho

So this way the reviewers will know exactly whether or not a PR is blocked by
their review, and authors will know exactly which users still need to be poked
when they only have partial approval for their PR.

Additionally, `gh pr_make` creates a second comment (separate from the PR body
comment) that contains information gleaned from codeowners.  This comment can
be generated again and again using `gee codeowners --comment` command, and thus
can be used to "ping" reviewers if necessary.

TODO(jonathan): Keep track of which reviewers have already commented, so that
only reviewers whose input is still required get pinged.

Tested:
  Used `gee codeowners --comment` command to ping reviewers on a separate PR.

PR generated by jonathan from branch gee_codeowners.

Commits:
*  63c0173 add owners command
*  a9bda03 gee: add codeowners command
*  014b415 annotate a PR request with the approval requirements.
*  3156a18 debugging
*  043f901 debugging
*  4757dce fixed pythonism that crept into my bash script
*  49acb0e fixed another pythonism that crept into my bash script
*  ce559a4 add a little more whitespace to the pr body


minor-fixes:
One thing I currently do is use the first PR description (first comment?) as a "draft" commit message - it contains all the context of the PR, is updated if the context changes, contains bug links, testing info, etc. I then copy/paste this first comment as the squash commit message.

I think the above is useful, and if it is more widely adopted then we can start reviewing the PR description as the final commit message as part of the review (when we start caring about commit messages post-submit)

That said, I don't think the list of codeowners would belong in such a commit message - would it be possible to add it as a follow-up comment instead? If/when we start having bots comment with helpful info (like presubmit bazel failing result links), I imagine it would fit that pattern rather well.

jonathan-enf:
@minor-fixes -- not everybody commits the same way, though.  But maybe what I can do for you in a future PR is: add the full commit log to the PR message as a set of comments.  gee strips comments out of the PR message before sending, so the user has the freedom to pick and choose what (if any) information from their commits they want to copy-paste into their PR body.

I agree that it's suboptimal to have codeowners metadata in our PR commit log, but I feel like not having this information easily visible in each PR is enough of a pain point that it's justified.  Though, I could move this information to a comment that gets added via "gh pr comment" right after "gh pr create".  Let me defer that to a future PR, because more PRs = more testing of this feature.

ccontavalli:
As a meta comment: `git shortlog` and similar display the first line of the commit message. It is strongly advised for the first line to be short (< 80 chars) and self contained, otherwise common git log commands will... be horrible.

CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+*.sh jonathan-enf
+/scripts jonathan-enf
```
ccontavalli, line 2: you won't be able to merge code in with just you as a CODEOWNER? github requires a codeowner that is NOT the author of the code. Add at least another person.

jonathan-enf: Done.

jonathan-enf:
> One thing I currently do is use the first PR description (first comment?) as a "draft" commit message - it contains all the context of the PR, is updated if the context changes, contains bug links, testing info, etc. I then copy/paste this first comment as the squash commit message.
> 
> I think the above is useful, and if it is more widely adopted then we can start reviewing the PR description as the final commit message as part of the review (when we start caring about commit messages post-submit)
> 
> That said, I don't think the list of codeowners would belong in such a commit message - would it be possible to add it as a follow-up comment instead? If/when we start having bots comment with helpful info (like presubmit bazel failing result links), I imagine it would fit that pattern rather well.

codeowners metadata is now moved to a follow-up comment, so it's separate from the first comment that will eventually become the merge-commit message.

jonathan-enf:
  ccontavalli jonathan-enf minor-fixes

ccontavalli (approved):

PR #320, opened 2021-11-19
flextape: Create server container

minor-fixes:
This change adds a container image definition for the flextape server.

The tag is left at `testing` for now, and can be changed after flextape is launched.

Tested: Ran `docker run -e PORT=6432 -p 6432:6432 gcr.io/devops-284019/infra/flextape:testing` and observed the following:

* Server started up successfully
* Server respected the `$PORT` env var
* Was able to curl the metrics endpoint

Jira: INFRA-108


flextape/server/BUILD.bazel
```diff
@@ -18,5 +19,40 @@ go_library(
...
+pkg_tar(
+    name = "flextape_tar",
+    srcs = [":flextape"],
+    package_dir = "/enfabrica/bin",
+)
+
+pkg_tar(
+    name = "flextape_config_tar",
+    srcs = ["flextape_config.textproto"],
+    package_dir = "/enfabrica/config",
```
aaahrens, line 33: This is okayish for now, usually config files (if they're in the container) and labelled under /default or smth. Usually, they're just documented

aaahrens (approved):

flextape/server/BUILD.bazel
minor-fixes: ack

PR #329, opened 2021-11-23
Fix flakey e2e test with item comparison

aaahrens:
changes a assert.Equal to be assert.ElementsMatch, was causing a flakey error when the data is right but in the wrong order


minor-fixes (approved):

minor-fixes:
Filed https://enfabrica.atlassian.net/browse/INFRA-195 for the presubmit failure

minor-fixes (commented):
TL;DR rebase your change/merge a recent master into your change to pick up `a1f75ba` to fix the presubmit

This fails the presubmit because while the presubmit does rebase changes, certain files take effect based on your PR before rebase during the presubmit process:
* the cloud build YAML definition
* the helper script that does the actual rebasing

`a1f75ba` fixed the helper script which was previously doing a botched merge; your PR doesn't include that PR.

PR #330, opened 2021-11-23
cloud build: Add postsubmit

minor-fixes:
This change adds a postsubmit build that builds and tests all bazel
targets in enkit. This will be used to detect bad changes that should
trigger someone to roll back.

Jira: INFRA-189


aaahrens:
We need to wait until the repo is in good shape (Already known but commenting for publicity)

minor-fixes:
> We need to wait until the repo is in good shape (Already known but commenting for publicity)

We don't; in fact, adding this while it's broken incentivizes us to fix it.

ahungrynacho (approved):

PR #331, opened 2021-11-23
bestie: Add initial skeleton

minor-fixes:
This change adds enough of a gRPC and HTTP application to:
* get something running on Google Cloud Run
* develop Bazel rules to deploy a container to Google Cloud Run

The initial gRPC service is a hello-world type service that is expected
to be deleted quickly; this is simply the quickest path to a working RPC
service for testing Cloud Run.

The binary also exports a /metrics endpoint, which we can use to test
HTTP fetches as a surrogate for an actual UI.

Tested: `bazel run //bestie/server:bestie` and successfully:
* did `grpc_cli` call of `Greet` RPC: `grpc_cli call localhost:6433 bestie.proto.Greeter.Greet 'name: "Scott"'`
* fetched metrics: `curl localhost:6433/metrics`

Jira: INFRA-193

### DIFF
--- a/bestie/server/main.go
+++ b/bestie/server/main.go
@@ -1,0 +1,33 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	hpb "github.com/enfabrica/enkit/bestie/proto"
+	"github.com/enfabrica/enkit/lib/server"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"google.golang.org/grpc"
+)
+
+type GreeterService struct {
+}
+
+func (s *GreeterService) Greet(ctx context.Context, req *hpb.GreetRequest) (*hpb.GreetResponse, error) {
+	return &hpb.GreetResponse{
+		Greeting: fmt.Sprintf("Hello, %s!", req.GetName()),
+	}, nil
+}
+
+func main() {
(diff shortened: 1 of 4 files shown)


bestie/server/main.go
```diff
@@ -0,0 +1,33 @@
...
+	}, nil
+}
+
+func main() {
+	grpcs := grpc.NewServer()
+	s := &GreeterService{}
+	hpb.RegisterGreeterServer(grpcs, s)
+
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.Handler())
```
ccontavalli, line 30: looks good as the first PR of a skeleton. I'd like to add:
1) Example prometheus metric (is there any existing gRPC metric exporter we can use?)
2) Logging.

3) UI, with typescript and react, using grpcweb. (probably a separate directory, separate main?)

Architecture has a BES endpoint ingestor, storing data in the database, and a UI pulling data from the database.

ccontavalli (approved):

bestie/server/main.go
minor-fixes: Agreed; this is just enough to test Cloud Run connectivity for both HTTP and gRPC requests (and maybe dive into how we do security).

Added tasks to https://enfabrica.atlassian.net/browse/INFRA-190 for the above.

PR #334, opened 2021-11-24
gee: use --quiet option with push universally.

jonathan-enf:
gee: Consistently use "--quiet" option with git push.

PR generated by jonathan from branch gee_push_quiet.

Commits:
*  bf29436 gee: use --quiet option with push universally.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -2502,7 +2502,7 @@ function gee__commit() {
     fi
     # We always push upstream so that users have a backup in case they lose their
     # laptop:
-    _git push -u origin "${CURRENT_BRANCH}"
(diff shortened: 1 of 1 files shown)


scripts/gee
```diff
@@ -2502,7 +2502,7 @@ function gee__commit() {
     fi
     # We always push upstream so that users have a backup in case they lose their
     # laptop:
-    _git push -u origin "${CURRENT_BRANCH}"
+    _git push --quiet -u origin "${CURRENT_BRANCH}"
```
ccontavalli, line 2505: should there be a `_git_push` function? so we don't have the same option over and over? Your call.

ccontavalli (approved):

PR #335, opened 2021-11-25
lib/token: add go docstrings.

ccontavalli:
This PR is a nop: code is not changed in any way. It only adds comments
explaining the use of the library, and the use of most functions.

### DIFF
--- a/lib/token/token.go
+++ b/lib/token/token.go
@@ -1,3 +1,37 @@
+// Package token provides primitives to create and decode cryptographic tokens.
+//
+// The library is built around the concept of Encoders: objects capable of turning
+// a byte array into another, by, for example, adding a cryptographic signature
+// created with an asymmetric key, encrypting the data, adding an expiry time, or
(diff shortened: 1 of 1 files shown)


lib/token/token.go
```diff
@@ -1,3 +1,37 @@
...
+//        Username, Lang string
+//      }{"myname", "english"}
+//
+//      b64string, err := encoder.Encode(uData)
+//
+// will convert a struct into a byte array, add the time the serialization happened,
+// encrypt all with a symmetric key, and then convert to base64.
+//
+// On Decode(), the original array will be returned after applying all the necessary
+// transformation and verifications. For example, Decode() will error out is the data
```
minor-fixes, line 32: `is` -> `if`

minor-fixes (approved):

lib/token/token.go
ccontavalli: addressed, and rebased.

PR #338, opened 2021-11-28
lib/token: add multi encryption support.

ccontavalli:
Add support to generate tokens that can be decrypted by different
keys.

This is in preparation for future code, but it allows to generate tokens
meant for specific systems to consume them, while issuing a different
key to each such system.

It can also be used to implement schemes where data is stored with
an escrow key alongside a per-user/per-session key, or where data
is encrypted to be sent to different recipients.

### DIFF
--- a/lib/token/BUILD.bazel
+++ b/lib/token/BUILD.bazel
@@ -4,6 +4,7 @@ go_library(
     name = "go_default_library",
     srcs = [
         "asymmetric.go",
+        "multi.go",
         "symmetric.go",
         "time.go",
         "token.go",
@@ -11,6 +12,7 @@ go_library(
     importpath = "github.com/enfabrica/enkit/lib/token",
     visibility = ["//visibility:public"],
     deps = [
+        "//lib/multierror:go_default_library",
         "@org_golang_x_crypto//nacl/box:go_default_library",
         "@org_golang_x_crypto//nacl/sign:go_default_library",
     ],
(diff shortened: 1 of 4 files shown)


ccontavalli:
Rebased on master.

aaahrens (approved):
I actually already think this exists in the stlib with rsa.OEAP and the fact that all curve keys support hybrid encryption, but I don't want to block you further

ccontavalli:
> I actually already think this exists in the stlib with rsa.OEAP and the fact that all curve keys support hybrid encryption, but I don't want to block you further

If you know of a golang library that supports encrypting / decrypting with multiple keys, I'd happily switch to that. I spent about an hour looking, and could not find anything.

Yes, most common uses of asymmetric encryption I know of end up using a symmetric algorithm for the actual encryption. The private/public key pair is used to encrypt/decrypt (or compute) a shared secret - CMK, Content Message Key. Eg, Curve25519 in the NACL library uses salsa underneath, while JWT with rsa.OEAP commonly ends up using either AES or cha cha underneath.

The library function added here is a very generic "compute a 'shared secret, a CMK' for... whatever encoder is used", and then "encrypt this CMK with multiple encoders", and "decrypt it with one that works".

This allows to use multiple public/private keys of possibly different types (eg, RSA, Curve25519, ...), or even a symmetric key (eg, for escrow), to store the same message. It's not efficient - as that it does not try to exploit any internal key computed by those algorithms (would be hard to re-use, unless all keys were guaranteed of the same type?), but should be safe.

Reason I had implemented this originally was exactly a) having an escrow key, b) having things stored so multiple groups/users can access them.

PR #339, opened 2021-11-28
lib/config/directory: minor bugfix - leaking file descriptor.

ccontavalli:
This PR fixes a minor file descriptor leak homedir Write, where
a tempfile is created, but the corresponding file object is never
closed.

This is generally not impacting, as configs are read/written to
at startup only.

But the config store API effectively provides a key value store,
and this could be used outside initial configuration handling.

Tested: unit tests pass before and after. No real test to detect leaks at this point.


minor-fixes (approved):
LGTM; presubmit failure looks like a flaky test (relying on map ordering?)

PR #340, opened 2021-11-28
lib/token: use config/marshal from TypeEncoder

ccontavalli:
Before this change:
TypeEncoder in lib/token would always use gob encoding, and could
use gob encoding only. While we had support for arbitrary marshallers
in lib/config/marshal, gob was not usable.

In this PR:
- add support for gob encoding in config/marshal, so gob is
  treated just like any other format supported.
- change token.TypeEncoder to use config/marshal. Specifically:
  - it now users marshal.Marshal/marshal.Unmarshal instead of
    hardcoding gob support.
  - it now provides a setter so an arbitrary marshaller can be
    selected.

Additionally:
- add initial tests for config/marshal. Goal for now is to just
  cover the added functionality.
- add tests in token_test.go to cover marshalling/marshalling with
  an arbitrary marshaller passed to type encoder.

Tested: all unit tests in lib/token and lib/config pass.

Note that there is no API change in this PR: all changes are within
the internal implementation of the config/marshal and token/ library.

### DIFF
--- a/lib/config/marshal/BUILD.bazel
+++ b/lib/config/marshal/BUILD.bazel
@@ -1,4 +1,4 @@
-load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 
 go_library(
     name = "go_default_library",
@@ -14,4 +14,14 @@ go_library(
         "@com_github_pelletier_go_toml//:go_default_library",
         "@in_gopkg_yaml_v2//:go_default_library",
     ],
+)
+
+go_test(
+    name = "go_default_test",
+    srcs = [
+        "marshal_test.go",
+        "select_test.go",
+    ],
+    embed = [":go_default_library"],
+    deps = ["@com_github_stretchr_testify//assert:go_default_library"],
 )
--- a/lib/config/marshal/marshal.go
+++ b/lib/config/marshal/marshal.go
@@ -1,6 +1,8 @@
 package marshal
 
 import (
+	"bytes"
+	"encoding/gob"
 	"encoding/json"
 	"github.com/pelletier/go-toml"
 	"gopkg.in/yaml.v2"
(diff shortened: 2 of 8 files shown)


minor-fixes (approved):

lib/token/asymmetric.go
```diff
@@ -0,0 +1,238 @@
+package token
```
aaahrens, line 1: I really would like for this to be labelled asymmetric_nacl or something, because this essentially is the multiencoder's interface over a asymmetric nacl

lib/token/symmetric.go
```diff
@@ -0,0 +1,194 @@
+package token
+
```
aaahrens, line 2: Uhm, can we just use the crypto blowfish library here instead of symmetric encryption? seems like it's more standard

lib/token/symmetric.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// SymmetricCreator is a CryptoFactory suitable for use with NewMultiKeyCryptoEncoder().
+var SymmetricCreator CryptoFactory = func(rng *rand.Rand, key []byte) (BinaryEncoder, []byte, error) {
+	if key != nil {
+		be, err := NewSymmetricEncoder(rng, UseSymmetricKey(key))
+		return be, key, err
+	}
+
+	key, err := GenerateSymmetricKey(rng, 0)
```
aaahrens, line 155: I don't like this behaviour. I feel like it shouldn't default to random if not set. I don't know feels like it could lead to a lot of misleading tests for applications.

lib/token/symmetric.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	return key, nil
+}
+
+// Creates a new random key and stores it in settings, or return error.
+//
+// size represents the key size in bits. If size is 0, uses a default key size
+// of 256 bits.
+func WithGeneratedSymmetricKey(size int) SymmetricSetter {
```
aaahrens, line 61: I feel like this one and the other factories should be "WithGenerateSymmetricKeyFromSize"

aaahrens (changes requested):

lib/token/symmetric.go
ccontavalli: Orthogonal to this PR? Also, that's pretty much the only parameter to generate a symmetric key? In hindsight, this should probably have taken no parameter by default.

lib/token/symmetric.go
ccontavalli: In this specific function, CryptoFactory, it has to be a random key. This is used by the MultiCrypto encoder, which needs "something" to a) provide a random/unguessable/secret key to use to encrypt the message once, and then 2) an encoder capable of encrypting that data.

Then, this unguessable key is encrypted with each key that should have access to the message in turn, so any of those *other* keys can be used for crypto purposes.

The other encoders default to use a random key as a) it makes development simpler (start server, it works), b) it is the safest (eg, no secret stored on disk), and c) is "acceptable" in many deployments (single backend, restarted rarely, tokens are valid until timeout or until backend is restarted).

But also orthogonal to this PR?

ccontavalli (commented):
The other commits part of this PR were sent as individual PRs, most of the comments you added were about code in other PRs? Replying to them anyway, rebased and updated as those PRs have been merged a while back.

ccontavalli:
Merging on Scott's approval. If there's further changes requested for this PR, will gladly put together another PR.

PR #342, opened 2021-11-30
bazel: Add cloud_run rules

minor-fixes:
This change adds rules to deploy a Docker image to Cloud Run, and
provides an example deployment for bestie.

This change also adds a new common server factory function, since
the current factory doesn't seem to allow normal HTTP2 requests to succeed.
This new server sets up gRPC and HTTP2 but not grpc-web, which is
currently not needed. Ideally they should be coalesced into one method,
which we may be able to do once we understand the nuances of the
various protocols.

Tested:

```
# Built and deployed image
bazel run //bestie/server:bestie_image_push && bazel run //bestie/server:bestie_cloud_run_deploy

# Test grpc request
grpcurl -d '{"name": "scott"}'  bestie-24mnsxopwq-uw.a.run.app:443 bestie.proto.Greeter/Greet
# which returns
{
  "greeting": "Hello, scott!"
}

# Test HTTP2 request
curl --http2-prior-knowledge https://bestie-24mnsxopwq-uw.a.run.app/metrics
# which returns Prometheus metrics page
```

Jira: INFRA-191

### DIFF
--- a/bestie/server/BUILD.bazel
+++ b/bestie/server/BUILD.bazel
@@ -1,6 +1,7 @@
 load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 load("@rules_pkg//:pkg.bzl", "pkg_tar")
 load("@io_bazel_rules_docker//container:container.bzl", "container_image", "container_push")
+load("//bazel/cloud_run:defs.bzl", "cloud_run_deploy")
 
 go_library(
     name = "go_default_library",
@@ -12,8 +13,6 @@ go_library(
         "//lib/server:go_default_library",
         "@com_github_prometheus_client_golang//prometheus/promhttp:go_default_library",
         "@org_golang_google_grpc//:go_default_library",
-        "@org_golang_google_grpc//codes:go_default_library",
-        "@org_golang_google_grpc//status:go_default_library",
     ],
 )
 
@@ -48,4 +47,14 @@ container_push(
     repository = "devops-284019/infra/bestie",
     # TODO: Change this tag to "live"
     tag = "testing",
+)
+
+cloud_run_deploy(
+    name = "bestie_cloud_run_deploy",
+    access = "public",
+    image = "gcr.io/devops-284019/infra/bestie",
+    image_version = "testing",
+    project = "bestie-builds",
+    region = "us-west1",
+    service = "bestie",
 )
--- a/lib/server/run.go
+++ b/lib/server/run.go
@@ -78,6 +78,9 @@ func Run(mux *http.ServeMux, grpcs *grpc.Server) {
 // ✗ grpc-web via websockets
 // ✗ HTTP 1.1
 // ✔ HTTP 2.0
+//
+// TODO(INFRA-211): Merge with above Run() function if possible, and add tests
+// that verify the protocol compatibility table above either way.
 func CloudRun(mux http.Handler, grpcs *grpc.Server) {
 	if mux == nil {
 		mux = http.NewServeMux()
--- a/lib/server/run.go
+++ b/lib/server/run.go
@@ -8,6 +8,8 @@ import (
 
 	"github.com/improbable-eng/grpc-web/go/grpcweb"
 	"github.com/soheilhy/cmux"
+	"golang.org/x/net/http2"
+	"golang.org/x/net/http2/h2c"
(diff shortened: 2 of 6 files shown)


bestie/server/BUILD.bazel
```diff
@@ -49,3 +48,13 @@ container_push(
     # TODO: Change this tag to "live"
     tag = "testing",
 )
+
+cloud_run_deploy(
```
ccontavalli, line 52: ok for this PR: as mentioned, would be good to have a single step deployment, maybe using "multirun", or maybe without going through a container registry. Let me expand.

1) I don't think there's a good way to run "big query" locally. So, eg, for development and testing, developers will need to deploy all the time.

2) I would imagine a workflow where:
* Developer builds/deploys on cloud run all the time, his own instance, to see it working/testing, etc. Ideally, this does not require pushing a new container image all the time, if it does, it is some form of ephemeral table.
* When a new release is ready, a new container is prepared.
* Eventually it is deployed to staging and then production.

lib/server/run.go
```diff
@@ -56,3 +65,57 @@ func Run(mux *http.ServeMux, grpcs *grpc.Server) {
...
+// provided (is nil), one with default routes/services will be started,
+// respectively.
+//
+// CloudRun() is a helper for servers that run in Google Cloud run, and thus
+// starts a server that supports the following protocols:
+//
+// ✔ grpc
+// ✗ grpc-web via websockets
+// ✗ HTTP 1.1
+// ✔ HTTP 2.0
```
ccontavalli, line 80: I'd like this matrix here (and in the other Run()) method to be covered by some form of unit test, so to catch regressions, and more easily allow us to iterate if necessary on a single main.

I'd add it as a TODO, though, this is a good starting point.

ccontavalli (approved):

bestie/server/BUILD.bazel
minor-fixes: > I don't think there's a good way to run "big query" locally

Maybe we can run locally, while talking to a shared bigquery instance?

ccontavalli: It's possible and maybe ok. I was doing this for astore, but there are two main drawbacks:

1) it's a bit annoying in that it requires creating a service account, *downloading* a credentials file that we absolutely cannot check in in the repository, plumbing in flags so we can pass it, and having some process so we can revoke it when team members leave. So eg, we'd need to provide instructions on how to get this file, and be diligent in managing it.

2) when running locally .. it's a different environment than on cloud run/GAE. Can be small things like logging/monitoring not being available/searchable, to things like gRPC behaving slightly differently because there is no proxy in front of it. It can be desireable at times, but can also be a source of bugs/issues.

minor-fixes: Thinking about this more:
* The only way I can find to deploy to Cloud Run via bazel is via the `gcloud run deploy` command
* `gcloud run deploy` can deploy given an image and version (used here) or the `--source` parameter, which tries to build from the CWD by looking for a Dockerfile or using buildpacks. Bazel integration doesn't look promising here.
* As a result, we will always have to push before deploying
* Therefore, we probably need to parameterize both the service name and image tag, so there are no collisions between iteration cycles. Putting these in the BUILD file is suboptimal, as if someone forgets to change it before running they will clobber something else. We can parameterize the new deploy rule with a custom flag, requiring the dev to specify it on the commandline, but we can't with the container push rule, since we don't control it (we'd have to fork)

So I'm thinking (if this is necessary):
* leave the image tag at "latest" - latest will always get clobbered but hopefully that doesn't happen often
* add a flag for the developer to set the service suffix (required) - so a prod push would require e.g. `bazel run //bestie/server:bestie_cloud_run_deploy --//f/cloud_run:service_suffix=prod` - allows for arbitrary "environments"
* use multirun to reduce the possibility of races

...or if testing locally against a remote bigquery instance is possible, maybe we can punt on the flag, and just have a "staging" and "prod" rule instances. I think this will become more clear after initial bigquery support is added, so let's punt until then?

lib/server/run.go
minor-fixes: Done

bestie/server/BUILD.bazel
ccontavalli: sgtm

PR #348, opened 2021-12-04
enproxy: add rules to release docker image.

ccontavalli:
This PR adds a few rules to build a distroless container with enproxy,
and to upload it to gcr.

I realized today that while I had used this rule multiple times in
the past, I had never merged it in.

### DIFF
--- a/proxy/BUILD.bazel
+++ b/proxy/BUILD.bazel
@@ -44,3 +44,46 @@ go_library(
     importpath = "github.com/enfabrica/enkit/proxy/credentials",
     visibility = ["//visibility:public"],
 )
+
+load("@io_bazel_rules_docker//go:image.bzl", "go_image")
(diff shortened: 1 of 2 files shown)


ccontavalli:
@msgoldflam

proxy/BUILD.bazel
```diff
@@ -44,3 +44,46 @@ go_library(
...
+    env = {
+      "HOME": "/cache",
+    },
+)
+
+go_image(
+    name = "enproxy-image",
+    base = ":enproxy-base",
+    binary = ":enproxy",
+    visibility = ["//visibility:public"],
```
minor-fixes, line 73: ...is visibility:public everywhere a gazelle-ism? Can we turn that off? If not, why public?

minor-fixes (approved):

PR #354, opened 2021-12-07
flextape: Add Docker deployment script

minor-fixes:
This change contains a script with the Docker command used to bring up
Flextape on build-00. This script can get deleted once flextape is
migrated to the cluster.

Tested: Ran this command on build-00; flextape is currently running.

Jira: INFRA-252

### DIFF
--- a/flextape/server/run.sh
+++ b/flextape/server/run.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+PORT=8000
+
+docker run \
+  -d \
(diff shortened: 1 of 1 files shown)


ccontavalli:
Generally recommend using docker-compose yaml files over those scripts lately, I can point you to example in our repository.
Given the expected short lifetime of this script, approving.

ccontavalli (approved):

PR #355, opened 2021-12-07
Refactor and Test MultiError  

aaahrens:
Current issues with MultiError:

Not tested and according to the source code: 

https://cs.opensource.google/go/go/+/refs/tags/go1.17.4:src/errors/wrap.go;l=90

The Unwrap() function is supposed to be the equivalent of a list.pop(). This is not how multierror currently functions.

Right now, it essentially calls errors.As recursively, and for {} loops all get executed in a single pass recursively. 
With these changes, it follows the intended behaviour. 

Does the following things:

- Tests MultiError
- Updates the docs to specify why the library exists
- Changes Unwrap() to return an error
- Adds in a spread creator
- Cleans up the receiver names to be less generic (linting)

### DIFF
--- a/lib/multierror/multierror.go
+++ b/lib/multierror/multierror.go
@@ -87,5 +87,5 @@ func (multi MultiError) Error() string {
 		}
 		messages = append(messages, err.Error())
 	}
-	return strings.Join(messages, Separator)
+	return "Multiple errors:\n  " + strings.Join(messages, Separator)
 }
--- a/lib/multierror/multierror.go
+++ b/lib/multierror/multierror.go
@@ -17,7 +17,6 @@ type MultiError []error
 //		> errors.Is(err, &myTypedErr)
 //
 //
-// This is just a convenience wrapper to ensure that if there is no error,
 // if errs has len() == 0, or all errors are nil, nil is returned.
 // * Capture tracing from fmt.Errorf
 // 		> myTypedErr := errors.New("ssh agent failed to query keyring")
@@ -42,7 +41,6 @@ func Wrap(err ...error) error {
 
 // NewOr creates a MultiError from a list of errors, or returns the fallback error.
 //
-// Just like New, but gurantees that an error is returned. If the list of errors is
 // empty, it will return the supplied error instead.
 func NewOr(errs []error, fallback error) error {
 	if len(errs) == 0 {
@@ -63,27 +61,23 @@ func (multi MultiError) Unwrap() error {
 	if len(multi) == 0 {
 		return nil
 	}
-	return multi[0]
+	return multi[1:]
 }
 
 // As for MultiError returns the first error that can be considered As the specified target.
 func (multi MultiError) As(target interface{}) bool {
-	for _, err := range multi {
-		if errors.As(err, target) {
-			return true
-		}
+	if len(multi) == 0 {
+		return false
 	}
-	return false
+	return errors.As(multi[0], target)
 }
 
 // Is for MultiError returns true if any of the errors listed can be considered of the target type.
 func (multi MultiError) Is(target error) bool {
-	for _, err := range multi {
-		if errors.Is(err, target) {
-			return true
-		}
+	if len(multi) == 0 {
+		return false
 	}
-	return false
+	return errors.Is(multi[0], target)
 }
 
 func (multi MultiError) Error() string {
--- a/lib/multierror/multierror.go
+++ b/lib/multierror/multierror.go
@@ -10,6 +10,7 @@ const Separator = "\n  "
 type MultiError []error
 
 // New creates a MultiError from a list of errors.
+// if errs has len() == 0, or all errors are nil, nil is returned.
 // It has two primary intended uses:
 // * Capture maybe returns of ACID-style transactions or stack tracing.
 // 		> myTypedErr := errors.New("my specific query failed in transaction")
@@ -17,7 +18,6 @@ type MultiError []error
 //		> errors.Is(err, &myTypedErr)
 //
 //
-// if errs has len() == 0, or all errors are nil, nil is returned.
 // * Capture tracing from fmt.Errorf
 // 		> myTypedErr := errors.New("ssh agent failed to query keyring")
 //		> return multierror.Wrap(myTypedErr, fmt.Errorf("mw raw log %v", realAgentErr))
@@ -40,7 +40,6 @@ func Wrap(err ...error) error {
 }
 
 // NewOr creates a MultiError from a list of errors, or returns the fallback error.
-//
 // empty, it will return the supplied error instead.
 func NewOr(errs []error, fallback error) error {
 	if len(errs) == 0 {
--- a/lib/multierror/multierror.go
+++ b/lib/multierror/multierror.go
@@ -5,26 +5,48 @@ import (
 	"strings"
 )
 
-type MultiError []error
+const Seperator = "\n "
+
+type MultiError struct {
+	err1 error
+	err2 error
+}
+
+var (
+	_ error                           = &MultiError{}
+	_ interface{ Unwrap() error }     = &MultiError{}
+	_ interface{ Is(err error) bool } = &MultiError{}
+)
 
 // New creates a MultiError from a list of errors.
+// It has two primary intended uses:
+// * Capture maybe returns of ACID-style transactions or stack tracing.
+// 		> myTypedErr := errors.New("my specific query failed in transaction")
+//		> return multierror.Wrap(myTypedErr, tx.Rollback())
+//		> errors.Is(err, &myTypedErr)
 //
-// This is just a convenience wrapper to ensure that if there is no error,
-// if errs has len() == 0, or all errors are nil, nil is returned.
 //
-// In facts, you could normally convert a []error{} into a MultiError by simply
-// using a cast: MultiError(errs). However, a simple cast would lead to a non
-// nil error when errs has lenght 0.
+// * Capture tracing from fmt.Errorf
+// 		> myTypedErr := errors.New("ssh agent failed to query keyring")
+//		> return multierror.Wrap(myTypedErr, fmt.Errorf("mw raw log %v", realAgentErr))
+//		> if errors.Is(err, &myTypedErr); fmt.Println("I know it was the keyring, here is the stack ", err.Error())
+//
 func New(errs []error) error {
 	if len(errs) == 0 {
 		return nil
 	}
-	for _, err := range errs {
-		if err != nil {
-			return MultiError(errs)
-		}
+	l := len(errs)
+	if l >= 2 {
+		return &MultiError{err1: errs[0], err2: New(errs[1:])}
+	}
+	if l == 1 {
+		return &MultiError{err1: errs[0], err2: nil}
 	}
 	return nil
+}
+
+func Wrap(ers ...error) error {
+	return New(ers)
 }
 
 // NewOr creates a MultiError from a list of errors, or returns the fallback error.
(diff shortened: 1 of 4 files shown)


ccontavalli:
Can you expand on the PR description? What problem does this PR solve?
Eg, `Is`, `As`, `Unwrap` should already be working? Or are they not?

lib/multierror/multierror.go
```diff
@@ -46,13 +59,16 @@ func NewOr(errs []error, fallback error) error {
...
+	if len(multi) == 0 {
+		return nil
+	}
+	return multi[0]
 }
 
 // As for MultiError returns the first error that can be considered As the specified target.
-func (me MultiError) As(target interface{}) bool {
-	for _, err := range me {
+func (multi MultiError) As(target interface{}) bool {
```
minor-fixes, line 70: One letter should be sufficient; `me` was fine - why the change to `multi`?

lib/multierror/multierror_test.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func TestSanityIsError(t *testing.T) {
+	err := multierror.Wrap(OneErr, ThreeErr, FourErr)
+	assert.True(t, errors.Is(err, OneErr))
+	assert.False(t, errors.Is(err, TwoErr))
+	assert.True(t, errors.Is(err, ThreeErr))
+	assert.True(t, errors.Is(err, FourErr))
+	realErrStrings := []string{OneErr.Error(), ThreeErr.Error(), FourErr.Error()}
+	assert.Equal(t, strings.Join(realErrStrings, multierror.Separator), err.Error())
```
minor-fixes, line 37: This last assert is simply a change detector - it reimplements `MultiError.Error` such that if the impl of the latter changes, the test will break.

If we expect `.Error()` to contain the error strings of all errors, better would be to assert that `err.Error()` contains substrings `OneErr.Error()`, `TwoError.Error()`, etc.

lib/multierror/multierror.go
```diff
@@ -46,13 +59,16 @@ func NewOr(errs []error, fallback error) error {
 // Unwrap for MultiError always returns nil, as there is no reasonable way to implement it.
 //
 // Use As and Is methods, or loop over the list directly to access the underlying errors.
-func (me MultiError) Unwrap() error {
-	return nil
+func (multi MultiError) Unwrap() error {
+	if len(multi) == 0 {
```
minor-fixes, line 63: Why always the first error?

lib/multierror/multierror.go
```diff
@@ -46,13 +59,16 @@ func NewOr(errs []error, fallback error) error {
 // Unwrap for MultiError always returns nil, as there is no reasonable way to implement it.
```
minor-fixes, line 59: Needs to be updated if the method is changed

lib/multierror/multierror.go
```diff
@@ -61,26 +77,22 @@ func (me MultiError) As(target interface{}) bool {
...
+func (multi MultiError) Error() string {
+	var messages []string
+	for _, err := range multi {
 		if err == nil {
 			continue
 		}
 		messages = append(messages, err.Error())
 	}
-	return "Multiple errors:\n  " + strings.Join(messages, "\n  ")
+	return strings.Join(messages, Separator)
```
minor-fixes, line 97: "Multiple errors:" is dropped?

lib/multierror/multierror.go
aaahrens: once we start linting it would come up as too generic. Superficial change.

lib/multierror/multierror.go
ccontavalli: I'd prefer to keep it, makes for a more human friendly and readable output.
(assuming we keep "Multiple errors:", please keep the special casing for a single error - again, goal is user friendliness / easier to understand output)

lib/multierror/multierror.go
aaahrens: I went back because of this comment, and realized I had implemented it wrong! :> now it properly unwraps

lib/multierror/multierror.go
```diff
@@ -46,41 +57,36 @@ func NewOr(errs []error, fallback error) error {
...
-	for _, err := range me {
-		if errors.As(err, target) {
-			return true
-		}
+func (multi MultiError) As(target interface{}) bool {
+	if len(multi) == 0 {
+		return false
 	}
-	return false
+	return errors.As(multi[0], target)
```
ccontavalli, line 72: this does not seem right? eg, `As(` should work if any of the errors in the array can be a target?

lib/multierror/multierror.go
```diff
@@ -46,41 +57,36 @@ func NewOr(errs []error, fallback error) error {
...
-	for _, err := range me {
-		if errors.Is(err, target) {
-			return true
-		}
+func (multi MultiError) Is(target error) bool {
+	if len(multi) == 0 {
+		return false
 	}
-	return false
+	return errors.Is(multi[0], target)
```
ccontavalli, line 80: same here, we still need the loop?

lib/multierror/multierror.go
aaahrens: I'm generally allergic to  uneditable formating  inside errors themselves. 
It would be much more idiomatic if instead we have another function that handles the error printing themselves. 
Something like this 

`errhandle.PrintErr(err)` <= and inside this we can check if it's a multierror or not and do error handling things. 

Ideally no one should just fmt.Printtln(err.Error())

minor-fixes: I suspect this change also makes it easier to test in the manner done in this PR - though if the test suggestions are followed (test the final error contains substrings for each included error, instead of joining the test errors manually and testing for equality) then this change can be dropped.

I think it's open for debate, but definitely not the most high-priority debate we have today, and there's no need for this to stall an otherwise good change that adds needed unit tests.

lib/multierror/multierror_test.go
aaahrens: You're right, changed testing to reflect this

lib/multierror/multierror.go
aaahrens: so here is the source code for errors.Is and As 
https://cs.opensource.google/go/go/+/refs/tags/go1.17.4:src/errors/wrap.go;l=90

The Unwrap() function is supposed to be the equivalent of a list.pop(), which is why all of the tests still pass with these changes.

Right now, it essentially calls errors.As recursively, and for {} loops all get executed in a single pass recursively. 

These changes make it so that it behaves as expected. (I updated this PR's description to include this)

lib/multierror/multierror.go
aaahrens: See above

lib/multierror/multierror.go
ccontavalli: uhm .. this regresses on error messages commonly shown by enkit login / enkit tunnel etc.
It goes from a clear message showing that multiple errors happened, to a naked array.

Pretty much all errors in go are full of hardcoded strings. eg, just grep the source code for fmt.Errorf of base libraries?

lib/multierror/multierror.go
ccontavalli: The reason this was not implemented is that semantically the errors in a MultiError are NOT nested.

Let me expand on this. Let's say you get a DataBaseError. You Unwrap it, and discover that you got a DatabaseError because there was a ConnectionError. You Unwrap it again, and discover that you got a ConnectionError because of a DNSResolutionError.
In the code, you can say "if DataBaseError IS a DNSResolutionError", then ... retry without DNS, for example.

Now, let's say that I now have a MultiError containing 3 DataBaseError. The caller invokes Unwrap(). The current code returns just another MultiError, with 2 DatabaseError. Instead of returning THE CAUSE of an error, it just moves to the next element in the array. It then calls Unwrap() again, and another MultiError with 1 element is returned. So again, not the cause of the error. Finally, nil is returned.

Can you add a test with error nesting, and see what happens with Is and As? It's hard to follow the code in the golang library, but my guess is that the change in this PR are semantically breaking the expected behavior of Unwrap.

Can you try a test with normally nested errors? I suspect this test would pass BEFORE your change, but would FAIL after your change:
```
  errs := []error{
    fmt.Errorf("outer caused by %w", fmt.Errorf("inner caused by %w", OneErr)),
    fmt.Errorf("outer caused by %w", fmt.Errorf("inner caused by %w", TwoErr)),
  }

  assert.True(t, errors.Is(errs, OneErr))
  assert.True(t, errors.Is(errs, TwoErr))
  assert.False(t, errors.Is(errs, ThreeErr))
```

ccontavalli: (please don't merge unless this comment is addressed - until I'm convinced otherwise, while I think the tests are valuable, I think this is breaking the contract for Unwrap(), Is() and As())

lib/multierror/multierror.go
ccontavalli: Also, the recommended way to print an error is `%v` or `%w` in `fmt.Errorf`. I've almost never seen go code special casing an error for the purpose of printing.

aaahrens: This is more of a molehill to die on than a hill, so I'll just revert it.

aaahrens: But I want to point out that having multiple errors is NOT and error, which is why I wanted to remove it.

lib/multierror/multierror.go
aaahrens: 1. I already have nested testing [here](https://github.com/enfabrica/enkit/pull/355/files#diff-b0d1f1c620fbee8ca09faef28317d5a966fb36cc07eefd0808b8f51a5ab65cf5R68) I also added in your example.
2. The code snippet you passed in passes both before and after. These changes as a whole are intentionally not behavior changing. That's why I made tests.
3. What the current library does is call errors.Is on all of it's existing errors, instead of letting the errors.Is function body itself do that.

minor-fixes (commented):
(unresolved 1)

lib/multierror/multierror.go
minor-fixes: The reason the code above works is that it relies on `errors.Is` to call `Unwrap()` when the current error doesn't match, which effectively recursively replaces the loop in the old `multierror.Is` implementation.

The debate is on whether `multierror.Unwrap` iterating over errors makes sense. `Unwrap()` is supposed to reveal the cause of errors; in the case of multierror, the next error is probably not the cause of the previous error, so this will be confusing to multierror users.

There are useful implementations of `As()` and `Is()` for `multierror`, as previously shown. Its clear that this code can be refactored by implementing `Unwrap()` (which `errors.As` and `errors.Is` leverages), but `Unwrap()` by itself does something different than it should (iterating over disparate errors rather than finding the cause of any particular error) and I think that's the main objection.

minor-fixes (commented):
(unresolved 1)

lib/multierror/multierror.go
aaahrens: This would be true if, and only if, Unwrap actually had a reason to exist. It does not. Referenced multiple times in the errors package, like [here](https://cs.opensource.google/go/go/+/refs/tags/go1.17.5:src/errors/wrap.go;l=26), Unwrap is simply a pop from a list. errors.As and errors.Is serve that functionality to "reveal" causes of errors.

To have a list of errors, and when called to Unwrap (aka to pop) and return nil even though the end user knows there are multiple errors in there is contract breaking.

aaahrens: What if, for example, I wrote a errors library that will parse the typing of each err in an err chain and upload it to some statistics database. 

I would not use errors.As or errors.Is, I would use Unwrap to parse all individual errors and their typing. So I would expect that if err != nil for a top level operation, Unwrap shouldn't return nothing.

PR #358, opened 2021-12-08
gee: explicit paths for enkit

jonathan-enf:
Some user's environments or containers don't seem to set /opt/enfabrica/bin in
the default PATH variable.  This replaces all of gee's invocations of enkit
with a fully specified path.

PR generated by jonathan from branch gee_enkit_path.

Commits:
*  a664208 gee: explicit paths for enkit

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -542,13 +542,13 @@ function _check_ssh_agent() {
   # Check that the ssh-agent is loaded and reachable.
   if [[ -z "${SSH_AUTH_SOCK}" ]]; then
     _warn "SSH_AUTH_SOCK is not set."
-    _info "Consider adding \"eval \`enkit agent print\`\" to your .bashrc"
+    _info "Consider adding to your .bashrc: eval \"\$(${ENKIT} agent print)\""
     if _confirm_default_no \
       "Would you like gee to append this line to your .bashrc file now? (y/N)  "
     then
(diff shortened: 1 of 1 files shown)


minor-fixes (commented):
> Some user's environments or containers don't seem to set /opt/enfabrica/bin in the default PATH variable

I think this is a container bug, because we always want `enkit` in `$PATH`; otherwise we have different problems, as lots of docs become less usable. Maybe we just fix the container bug?

Using `enkit` from `$PATH` would also allow for users to override it if desired (though admittedly this is a double-edged sword)

jonathan-enf:
> > Some user's environments or containers don't seem to set /opt/enfabrica/bin in the default PATH variable
> 
> I think this is a container bug, because we always want `enkit` in `$PATH`; otherwise we have different problems, as lots of docs become less usable. Maybe we just fix the container bug?
> 
> Using `enkit` from `$PATH` would also allow for users to override it if desired (though admittedly this is a double-edged sword)

It's been 3 years but I'm not sure if all containers are fixed yet.

I think we need this override -- this treats enkit the same as git, jq, gh, and other tools that gee must depend on in a consistent manner.

minor-fixes (approved):
Well, I can't remove myself as reviewer... but I can approve the PR! I say go for it 😉  👍

mhallmark (approved):

PR #359, opened 2021-12-09
Sketch out rpc for Controlplane state

aaahrens:
Defines an rpc service that is designed to run on the controlplane. It exports the control planes  current state.


machinist/rpc/machinist.proto
```diff
@@ -54,6 +54,23 @@ message DownloadResponse {
   bytes data = 1;
 }
 
+message ControlPlaneStateRequest {
+
+}
+
+// StaticMachine is a singular machine usually pushed from the controlplane. It represents an immutable object
+// scraped from the network
+message ControlPlaneStaticMachine {
```
minor-fixes, line 63: Should this be named just `StaticMachine` - is the `ControlPlane` prefix needed to disambiguate it from another kind of static machine?

machinist/rpc/machinist.proto
```diff
@@ -54,6 +54,23 @@ message DownloadResponse {
...
+  string name = 1;
+  repeated string ips = 2;
+  repeated string tags = 3;
+}
+
+message ControlPlaneStateResponse {
+  repeated ControlPlaneStaticMachine machines = 1;
```
minor-fixes, line 70: `static_machines`?

PR #360, opened 2021-12-09
flextape: Add cadence licenses

minor-fixes:
This change adds cadence licenses for all features that expire in March
2022. Licenses for features that expire soon are skipped.

Jira: INFRA-274

### DIFF
--- a/flextape/server/flextape_config.textproto
+++ b/flextape/server/flextape_config.textproto
@@ -177,3 +177,35 @@ license_configs {
   }
   quantity: 3
 }
+
+license_configs {
+  license {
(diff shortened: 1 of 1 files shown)


minor-fixes:
Once this is merged, I'll rebuild + redeploy flextape and test to make sure cadence tools work with the cutover PR. If not - will fix and create another PR. Risk is low since flextape is not launched yet.

ahungrynacho (approved):
Note that we will get another set of updated licenses from Cadence within the next 2 weeks with different license features and counts since we are still negotiating with them so this file will have to be updated.

PR #365, opened 2021-12-10
Refactor Testing for Socket Shims

aaahrens:
Problem:
- Thought about it, and the way the testing was written was pretty sloppy and couldn't be reused across other tests.
Solution:
- Made a documented //enfuse/testserver package 
- refactored existing tests for pooling and shims to be more descriptive


proxy/enfuse/socket.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (s SocketShim) Write(p []byte) (n int, err error) {
+	// websockets always write the full buffer
+	realPayload := append(s.Prefix, p...)
+	return len(p), s.WebConn.WriteMessage(websocket.BinaryMessage, realPayload)
+}
+
+func NewSocketShim(strategy PayloadAppendStrategy, Conn *websocket.Conn) (*SocketShim, error) {
+	pLen, f := strategy()
+	pre, err := f()
```
minor-fixes, line 42: (future PR) this still smells to me - I suspect that PayloadAppendStrategy should actually be an interface

proxy/enfuse/socket.go
```diff
@@ -0,0 +1,114 @@
...
+func (w *WebsocketTCPShim) handleReadToWebsocket(c net.Conn, uid []byte) {
+	go func() {
+		if _, err := io.Copy(SocketShim{WebConn: w.webConn, Prefix: uid, PrefixLen: len(uid)}, c); err != nil {
+			w.l.Errorf("err copying for client %v", err)
+		}
+	}()
+}
+
+func (w *WebsocketTCPShim) Close() error {
+	return nil
```
minor-fixes, line 113: Nothing to do here is surprising - does it need to kill goroutines? At least a comment that explains that this is TODO/what it should do would be helpful

proxy/enfuse/socket_test.go
```diff
@@ -0,0 +1,83 @@
...
+		Name: "goofy",
+	},
+}
+
+func TestSocketShim(t *testing.T) {
+	s := testserver.NewWebSocketBasicClientServer(t)
+	defer s.Close()
+
+	// Dial all websocket clients to the httptest.Server
+	url := strings.Replace(s.URL, "http", "ws", -1)
```
minor-fixes, line 44: Just the first http needs to be replaced, right?

proxy/enfuse/testserver/hello.go
```diff
@@ -0,0 +1,55 @@
+//go:build !release
+// +build !release
```
minor-fixes, line 2: Why the build tags? Here and in other files

proxy/enfuse/websocket_test.go
```diff
@@ -1,64 +1,95 @@
...
-	time.Sleep(2 * time.Second)
+	messages := []string{"winnie", "piglet", "eeyore", "tigger"}
+	for _, message := range messages {
+		cc, _, err := websocket.DefaultDialer.Dial(url, nil)
+		assert.NoError(t, err)
+		err = p.WriteWebsocketServer(websocket.BinaryMessage, []byte(message), enfuse.NewWebsocketLock(cc))
+		assert.NoError(t, err)
+	}
+	// give time for cpu cycles
+	time.Sleep(200 * time.Millisecond)
```
minor-fixes, line 60: (nit) this scares me - most tests I see with sleeps are either flaky or slow; even tests that should work with a particular timeout experience failures when you change the build machine and/or load up the build machine (like when tests run in parallel with other builds/tests). Ideally there's some sort of condition to wait for in order to proceed with confidence.

No action needed, but hopefully the number of tests that rely on timing can be minimized moving forward

proxy/enfuse/socket.go
aaahrens: Maybe, depending on how complex it gets

proxy/enfuse/socket.go
aaahrens: It's essentially a NoOpWriteCloser (the proxy is one also). It doesn't need to kill it's goroutines.

proxy/enfuse/socket_test.go
aaahrens: I suppose putting the -1 is arbitrary, replace All looks prettier

proxy/enfuse/testserver/hello.go
aaahrens: They get automatically populated by gazelle. It makes it so that the go tool wont build the package for release

proxy/enfuse/socket.go
```diff
@@ -0,0 +1,114 @@
...
+	webConn *WebsocketLocker
+	netConn net.Listener
+
+	clientMap map[string]net.Conn
+	prefixLen int
+
+	l logger.Logger
+}
+
+// handleWrites will handle all writes from webConn to netConn. It reads the content of the payload based on
```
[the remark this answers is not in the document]
aaahrens: I'm actually removing the doc because I have better ones coming up on the next PR

proxy/enfuse/websocket_test.go
aaahrens: I don't like to too, it feels way to unstable for me. I'm hoping to expand the testserver package to check and wait for all write, but that'll be another PR.

proxy/enfuse/socket_test.go
minor-fixes: ...but only the first should be replaced, right? This rewrites `http://somehttpserver.com` into `ws://somewsserver.com` which might be OK in the context of the test but definitely dials a different server

minor-fixes (approved):
Please take another look at the `ReplaceAll` before merging

proxy/enfuse/socket_test.go
aaahrens: This rewrites the httptest.Server, which is guaranteed to be an ip and random port. I don't think it'll ever change to be a url.

PR #375, opened 2021-12-21
Machinist Bug Fixes

aaahrens:
These are just a collection of bug fixes for machinist that needed to get done for NC/MTV
Problems:
- settting the state file and dns port did nothing to the internal application
- machines would error out if they had the same name instead of updating in place (people wanted this changed)

Solutions:
- state.AddMachine no longer errors if the machine is the same name, updates in place instead
- updates our go.mod and updates the golang-protobuf library a minor version (Note, not protobuf, but the go library bindings)
- Puts the DNS port and the state files from the CLI to the running 
- Fixes a flakey e2e test. 
- Removes random print statements

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,30 +1,58 @@
 module github.com/enfabrica/enkit
 
-go 1.14
+go 1.17
+
+require (
+	cloud.google.com/go/datastore v1.1.0
+	cloud.google.com/go/storage v1.10.0
+	github.com/bazelbuild/buildtools v0.0.0-20211007154642-8dd79e56e98e
+	github.com/cheggaaa/pb/v3 v3.0.5
+	github.com/coreos/go-oidc v2.2.1+incompatible
+	github.com/cybozu-go/aptutil v1.4.2-0.20200413001041-3f82d8384481
+	github.com/docker/docker v20.10.3+incompatible
+	github.com/dustin/go-humanize v1.0.0
+	github.com/fatih/color v1.10.0
+	github.com/go-git/go-git/v5 v5.1.0
+	github.com/golang/protobuf v1.5.2
+	github.com/google/go-github v17.0.0+incompatible
+	github.com/google/uuid v1.3.0
+	github.com/gorilla/websocket v1.4.2
+	github.com/improbable-eng/grpc-web v0.13.0
+	github.com/kataras/muxie v1.1.1
+	github.com/kirsle/configdir v0.0.0-20170128060238-e45d2f54772f
+	github.com/mitchellh/mapstructure v1.3.3
+	github.com/pelletier/go-toml v1.8.1
+	github.com/pkg/browser v0.0.0-20180916011732-0a3d74bf9ce4
+	github.com/prometheus/client_golang v1.11.0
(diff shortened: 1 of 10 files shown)


go.mod
```diff
@@ -1,30 +1,58 @@
 module github.com/enfabrica/enkit
 
-go 1.14
+go 1.17
```
minor-fixes, line 3: Do we need to hang back at go 1.16 for appengine support still?

machinist/example/deployment/docker-compose.yml
```diff
@@ -0,0 +1,16 @@
...
+
+services:
+  controlplane:
+    image: gcr.io/devops-284019/infra/machinist/controlplane:latest
+    command:
+      - "--state=/enfabrica/dns.state.json"
+      - "--dns-port=4455"
+      - "--bind-net=0.0.0.0"
+      - "--port=4456"
+      - "--domains=hello.loca"
```
minor-fixes, line 11: loca?

machinist/polling/register.go
```diff
@@ -25,12 +26,18 @@ func SendRegisterRequests(ctx context.Context, client machinist_rpc.ControllerCl
 	}
 	for {
 		if err := pollStream.Send(registerRequest); err != nil {
-			l.Errorf("unable to send request: %w", err)
+			_, err := pollStream.Recv()
+			s, ok := status.FromError(err)
+			if ok {
+				l.Errorf("unable to send register request: %s %s", s.Code(), s.Message())
```
minor-fixes, line 32: Does %w and passing err directly not log both the status and message?

machinist/state/controlplane.go
```diff
@@ -23,12 +22,16 @@ type MachineController struct {
 func AddMachine(mc *MachineController, m *Machine) error {
 	mc.Lock()
 	defer mc.Unlock()
-	for _, mm := range mc.Machines {
-		if mm.Name == m.Name {
-			return fmt.Errorf("machinist: named machine %s already exists", m.Name)
+	modifiedInPlace := false
+	for i := range mc.Machines {
+		if mc.Machines[i].Name == m.Name {
```
minor-fixes, line 27: Should `Machines` be a `map[string]*Machine` keyed by name instead?

go.mod
aaahrens: Ups, changed to 1.16. Re ran gazelle and update-repos, no change.

machinist/example/deployment/docker-compose.yml
aaahrens: changed to changeme.local (which I think gets the point across)

machinist/polling/register.go
aaahrens: It does, I just was just doing it so it woudl log prettier. have no issue %w ing.

machinist/state/controlplane.go
aaahrens: So I intentionally left this an array because I didn't want to be stuck on strings as uniqueness. Eventually I want to move to generics or the Equals interface.

minor-fixes (approved):

PR #378, opened 2021-12-23
gee: starting to rewrite in golang

jonathan-enf:
gee: starting to rewrite in golang

Looking for feedback on general approach.  Right now, only one command is
implemented (commit), and it was used to create the last two commits of this
PR.

This still needs a ton of work, but I'm pushing this just to get some early
feedback on the approach before I get too far.  My golang is *really* rusty.

I threw away my first attempt (gee as an enkit subcommand) and implemented gee
as a stand-alone app using the default cobra libraries (better documentation).

Tested: It compiles, and I made a commit.

PR generated by jonathan from branch gee_golang.


minor-fixes (commented):
I think this is a great start! A couple nits on style, and I think consideration is warranted in a couple places as this gets off the ground:

Singletons

I'm not opposed to the singleton approach (as opposed to forcing callers to create and pipe e.g. log objects through everywhere) as it tends to work for the basic usecases while being more terse than the alternative. There is a lot of `lib.GetSingleton().DoTheThing()` here though - maybe when combined with separate libraries (next point) package-level functions can wrap getting the singleton? This would shorten `lib.Logger().Debugf("..")` to `log.Debugf("...")` which internally gets the singleton.

File organization

Combining lots of modules into one package "lib" has two drawbacks:

* It allows for inter-dependencies between what should be logically-separate modules to creep in over time, as each has access to the other's private types/functions/fields/methods (if logger and runner can indeed be separated)
* It creates more noise in the command code, as every function/type needs a `lib.` prefix and needs a longer name to disambiguate it from within `lib`.

Maybe putting code into separate directories per module (`gee/log/log.go` for a logging package, `gee/command/runner.go` for a command package, etc.) does the following:

* Plays nice with gazelle (don't have to hand-write BUILD files)
* Creates separate packages per library
* Drops `lib` both as a dirname and a filename, which adds noise

Testing

It doesn't look like any tests are added in this PR. If I were to implement a command, I might want to write some tests that ensure:
* the correct git or gh commandline gets created based on the gee args provided
* canned sample git/gh output is correctly interpreted by gee for the expected commandline
* gee prints an appropriate output to the user

Along these lines, it may be worthwhile to define some interfaces on the logging/command seams to capture these inputs/outputs - then the user could pass in a fake logger/fake command runner in tests to have a path to make these assertions.

PR #380, opened 2021-12-28
lib/knetwork/echo: small and naive echo server for testing.

ccontavalli:
Background:
Verifying that tunnels work as expected requires having a TCP endpoint
we can write to, and read from. This is a common enough pattern that
it's worth having a simple object to use that implements this feature.

This PR:
- implements a very naive TCP echo server. What's written to it gets
  written back.
- over features, I chose simplicity. There's very little code here.
  This also means the implementation is very naive: for example, on
  Close(), it terminates only the listening socket. All other sockets
  will terminate when the client disconnects. Also, no timeouts are
  implemented.

### DIFF
--- a/lib/knetwork/echo/BUILD.bazel
+++ b/lib/knetwork/echo/BUILD.bazel
@@ -3,6 +3,7 @@ load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
 go_library(
     name = "go_default_library",
     srcs = ["echo.go"],
+    testonly = True,
     importpath = "github.com/enfabrica/enkit/lib/knetwork/echo",
     visibility = ["//visibility:public"],
 )
--- a/lib/knetwork/echo/BUILD.bazel
+++ b/lib/knetwork/echo/BUILD.bazel
@@ -1,0 +1,15 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+go_library(
+    name = "go_default_library",
+    srcs = ["echo.go"],
+    importpath = "github.com/enfabrica/enkit/lib/knetwork/echo",
+    visibility = ["//visibility:public"],
+)
+
+go_test(
+    name = "go_default_test",
+    srcs = ["echo_test.go"],
+    embed = [":go_default_library"],
+    deps = ["@com_github_stretchr_testify//assert:go_default_library"],
+)
--- a/lib/knetwork/echo/echo.go
+++ b/lib/knetwork/echo/echo.go
@@ -9,6 +9,8 @@
 //
 // This is most useful for testing code that requires a TCP endpoint on
 // the other end.
+//
+// +build !release
 package echo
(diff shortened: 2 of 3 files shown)


aaahrens (changes requested):
Can we actually start moving test servers to their own top level packages instead of separating them?  I also have an echo wss server and several other testing type servers. 
I can put them in whatever package.

edit:

Also can we start attaching the nobuild release annotation for code designed to be importable by test only?

ccontavalli:
> Can we actually start moving test servers to their own top level packages instead of separating them? I also have an echo wss server and several other testing type servers. I can put them in whatever package.

So... the current convention is that test support libraries live alongside the corresponding library. Eg, astore test libraries live in astore, http test libraries live in lib/khttp, and so on. If we want to change the convention, let's 1) discuss in person, and 2) have a plan to migrate what we currently have? Having echo as a one off in a different directory is just going to add confusion?

> Also can we start attaching the nobuild release annotation for code designed to be importable by test only? If you add the annotation gazelle adds its own (and is supported)

uhm ... I've added a `// +build !release` tag, and a `testonly = True` tag in BUILD.bazel file. I've had to do both **manually**. It does not look like gazelle adds *testonly* automatically? Or do you know how to make that happen?

Checking our repository, I see:
```
$ find . -type f -name "*.go" |xargs grep "\+build.*release"
./astore/atesting/mocks.go:// +build !release
./lib/khttp/ktest/server.go:// +build !release
./lib/knetwork/echo/echo.go:// +build !release
```
and
```
$ find . -type f -name BUILD.bazel |xargs grep testonly
./lib/knetwork/echo/BUILD.bazel:    testonly = True,
./lib/errdiff/BUILD.bazel:    testonly = True,
./lib/testutil/BUILD.bazel:    testonly = True,
```
If you exclude echo (manually), it looks like target that have `+build !release` are NOT getting `testonly = True` in bazel, and vice versa? We're also not consistent in our convention.

Which convention do we want to use here? @minor-fixes as well. My vote is:

1) If // build can add testonly automatically, let's use it.
2) Otherwise, I would standardize on `testonly` in the BUILD.bazel file (which is enforced, I don't see // build enforced?)

minor-fixes (approved):

aaahrens (approved):

aaahrens:
The reason why it didn't happen is because that's not a valid go annotation, it should happen before the comments. We can discuss the refactor later.

Just change where the annotation is, then MAW.

ccontavalli:
> The reason why it didn't happen is because that's not a valid go annotation, it should happen before the comments. We can discuss the refactor later.
> 
> Just change where the annotation is, then MAW.

I put it at the top of the file, and it is still not propagating with gazelle. It is on top of the file on astore/.../mocks.go, and still don't see gazelle generating `testonly = True`. Did you get this to work before?

Merging as is, as suggested, let's separately clean this up.

PR #383, opened 2021-12-28
proxy/: refactor for modularity and testing.

ccontavalli:
The enkit proxy creates a MUX that is able to dispatch to an
http proxy, or to the nassh tunnel handling (which happens to
use websockets and GET/POST requests via HTTP).

Creation of this MUX was performed directly from main, together
with a bunch of flags/configuration parsing.

Among other things, this made testing harder.

In this PR:
- break out code out of main() into a dedicated object, under
  enproxy/. This object follows the common conventions in the
  enkit/ repo, allowing direct initialization via flags or via
  simple modifiers/functors.
- add unit test covering the basic functionality of the proxy
  as well as ability to establish tunnels.

Note that this PR is still in preparation for follow up PRs:
THERE WERE NO BEHAVIORAL CHANGES IN THIS PR. Tried to minimize
code changes as much as possible, outside those changes necessary
to break the code out.

Follow up PRs will change a few behavioral things. Specifically:
- add functionality to only test the config file
  (without loading the proxy)
- metrics for prometheus
- a few cleanups (eg, use a single mux).

But this break down is a precursor to those changes.

### DIFF
--- a/proxy/BUILD.bazel
+++ b/proxy/BUILD.bazel
@@ -8,15 +8,10 @@ go_library(
     deps = [
         ":credentials",
         "//lib/client:go_default_library",
-        "//lib/config/marshal:go_default_library",
-        "//lib/kflags:go_default_library",
         "//lib/kflags/kcobra:go_default_library",
-        "//lib/khttp:go_default_library",
-        "//lib/oauth:go_default_library",
         "//lib/srand:go_default_library",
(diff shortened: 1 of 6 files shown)


ccontavalli:
(rebased on latest merges)

minor-fixes (approved):

PR #384, opened 2021-12-29
postsubmit: Define build configuration proto

minor-fixes:
This change adds a documented protobuf that lays out the configuration
for postsubmit builds. Each configuration will be an
`infra.postsubmit.Build` message in its own configuration file;
Terraform and other configuration will be generated from this high-level
config and applied to create the builds in GCP.

Tested: Builds

Jira: INFRA-361

### DIFF
--- a/infra/postsubmit/proto/postsubmit.proto
+++ b/infra/postsubmit/proto/postsubmit.proto
@@ -1,0 +1,42 @@
+syntax = "proto3";
+
+package infra.postsubmit;
+
+message Build {
+  // Name of this build. Needs to be unique among all builds, and should match
+  // the regex `[a-z][a-z_]+`.
+  string name = 1;
+
+  // Repository that this postsubmit builds code from. Should be one of:
+  // * `enfabrica/enkit`
+  // * `enfabrica/internal`
+  string repository = 2;
+
+  oneof trigger {
+    // Interval on which to run the build. This is a normal cron schedule
+    // expression (https://crontab.guru/), except that the minutes fields can
+    // (and should) be a `?` rather than fixed; this will be translated into a
+    // stable value based on the name of this build, which allows builds to not
+    // all hammer infrastructure by starting e.g. every hour on the hour.
+    string cron = 3;
+  }
+
+  // Bazel target patterns to include in the build.
+  repeated string include_patterns = 4;
+
+  // Bazel targets to exclude from the build; should be a subset of
+  // include_patterns.
+  repeated string exclude_patterns = 5;
+
+  // Tags to exclude from the build; build will not build targets that:
+  // * are tagged with at least one of the specified tags
+  // * depend on a target tagged with at least one of the specified tags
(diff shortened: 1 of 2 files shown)


infra/postsubmit/proto/postsubmit.proto
```diff
@@ -0,0 +1,42 @@
+syntax = "proto3";
+
+package infra.postsubmit;
+
+message Build {
+  // Name of this build. Needs to be unique among all builds, and should match
+  // the regex `[a-z][a-z_]+`.
+  string name = 1;
```
aaahrens, line 8: So I don't like using names here, considering there's a high chance of collisions between teams. Why not use path relative from git root?

infra/postsubmit/proto/postsubmit.proto
```diff
@@ -0,0 +1,42 @@
...
+    // Interval on which to run the build. This is a normal cron schedule
+    // expression (https://crontab.guru/), except that the minutes fields can
+    // (and should) be a `?` rather than fixed; this will be translated into a
+    // stable value based on the name of this build, which allows builds to not
+    // all hammer infrastructure by starting e.g. every hour on the hour.
+    string cron = 3;
+  }
+
+  // Bazel target patterns to include in the build.
+  repeated string include_patterns = 4;
```
aaahrens, line 25: I feel like include and exclude are more designed for the use case of include "build/...." and exclude "build:my_heavy_test".

Maybe they should be lumped together in a message of their own and targets is a list of those messages. Makes sense to keep them tightly coupled.

aaahrens (changes requested):

infra/postsubmit/proto/postsubmit.proto
minor-fixes: Individual tests like `my_heavy_test` should probably be excluded via tag, not this mechanism - this is more for defining which parts of the code tree belong to your team (and therefore what your team's build should cover). It's usually the case for a team to own an entire directory, but it's not unheard of for a subteam/someone else to own a subdirectory of that - this mechanism is for excluding such subdirectories.

This shouldn't be the common case - I think splitting it out adds unnecessary nesting, and the config reads nicely as-is: "include //a, //b, //c, but not //a/d, //a/e"

infra/postsubmit/proto/postsubmit.proto
minor-fixes: This makes sense if this message is only supposed to be used from singular config files (which is the current plan), but if this ever changes (nesting this into a higher-level message, having multiple Builds in a file, etc) then one would really want the name field back, since keying off filename no longer makes sense.

Path from git root would also create potentially noisy/non-intuitive names - this forces teams to name their builds, meaning no intuition/translation is necessary (though probably there will be some repetition of the file's basename in this field)

ccontavalli (approved):

aaahrens (approved):

PR #385, opened 2021-12-30
cloud build staging: Add initial Terraform

minor-fixes:
This change adds a Terraform config that attempts to create the
following:

* a Cloud Build for the `enkit` repository
* a cron trigger in Cloud Scheduler for the aforementioned build
* an SMTP notifier that sends build failure notifications via Cloud Run

...in accordance with the manual setup steps described
[here](https://cloud.google.com/build/docs/configuring-notifications/configure-smtp)
and
[here](https://cloud.google.com/build/docs/automating-builds/create-scheduled-triggers).

Since this is our first official foray into Terraform, the directory
includes a README about how to install and use Terraform, as well as
manual setup steps for a new project that can't be done inside
Terraform.

This change commits the Terraform lock files as well as the recorded
state after `terraform apply` (so we can detect out-of-band changes).
Terraform temporary files (the .terraform directory, and the state
backup) are `.gitignore`d; therefore, users must run `terraform init`
commands before any `terraform apply` commands, as per the `README`.

The goal in this change is to create one "golden" config for a single build,
so that it becomes clear what needs to be parameterized when doing
config generation and scaling to N builds.

In future changes, the terraform commands are hopefully replaced with
bazel rules, and the terraform configs themselves hopefully replaced
with some sort of generation mechanism that exposes a cleaner interface.

Tested: Clicked "run now" on Cloud Scheduler trigger - Cloud Scheduler
successfully kicked Cloud Build, which failed and sent me a failure notification mail.

Jira: INFRA-296
Jira: INFRA-297

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ __pycache__/
 .pytest_cache
 .idea/
 .ijwb/
+.terraform
+*.tfstate.backup
--- a/infra/cloudbuild_staging/.terraform.lock.hcl
+++ b/infra/cloudbuild_staging/.terraform.lock.hcl
@@ -1,0 +1,38 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/hashicorp/google" {
+  version = "4.5.0"
+  hashes = [
+    "h1:MWHx6G+emVNvPyxxLCpEEco5F7eoRzX9Tp9lbtO/PmI=",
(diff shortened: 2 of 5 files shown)


jacob-adelmann (approved):
Methodology looks good. Went through the README.md, made it up to terraform plan but don't have permission on projects/enfabrica-cloud-build-staging to go further (which is totally fine). Good start to our Terraform journey!

image

PR #388, opened 2022-01-06
enfuse: Fuse over Websockets!

aaahrens:
This PR does the following:
- Tests running a FUSE filesystem over websockets.
- SocketShim now implements net.Conn
- Establishes the convention of how to use the tcp-to-websocket model (more documentation and refactoring later)
- Makes it so that the test server's no longer write back []byte with NULL characters appended.
- Add GrpcDialOpts to the connect configuration 
- Removes the Toy Caching from the FuseClient, wasn't really necessary

### DIFF
--- a/infra/postsubmit/proto/BUILD.bazel
+++ b/infra/postsubmit/proto/BUILD.bazel
@@ -1,13 +1,23 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
 load("@rules_proto//proto:defs.bzl", "proto_library")
 load("@io_bazel_rules_go//proto:def.bzl", "go_proto_library")
 
 proto_library(
     name = "postsubmit_proto",
     srcs = ["postsubmit.proto"],
+    visibility = ["//visibility:public"],
 )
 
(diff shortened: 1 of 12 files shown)


minor-fixes (commented):
Tests seem to hang on multiple runs: `bazel test //proxy/enfuse/... --runs_per_test=100`

minor-fixes:
Also please investigate presubmit issues: https://console.cloud.google.com/cloud-build/builds/b7fb3a63-cc5f-41ed-9db8-e02a03e28b5f;step=5?project=cloud-build-290921

aaahrens:
So runs per test = 100 really isn't feasible,  the core issue is that the scaffolding for the FUSE system takes a while, and so does verification of the data post-transfer. Later on I can do smoke tests with in memory file systems.  Frame of reference, --runs_per_test=3 takes about 120 seconds on my laptop.

aaahrens:
Actually scratch that, runs_per_test=100 should work now.
I also removed the time.Sleep from testing :D

PR #391, opened 2022-01-07
kunit: Use an ephemeral tmp path as the UML directory

rajachan:
As noted in the comment, this prevents a path string overflow. Without
this change, the mconsole initialization fails:

```
printk: console [stderr0] disabled
umid_file_name : buffer too short
```

And it succeeds with it:

```
printk: console [stderr0] disabled
mconsole (version 2) initialized on /tmp/tmp.tOCqXFlP01/zbnwLP/mconsole
```

### DIFF
--- a/bazel/linux/run_um_kunit_tests.template
+++ b/bazel/linux/run_um_kunit_tests.template
@@ -1,8 +1,17 @@
 #!/bin/bash
 
-set -e
+set -ex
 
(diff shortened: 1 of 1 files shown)


ccontavalli (approved):
I hate those problems :/ we were supposed to have flying cars by now, and we can't handle a path that is too long. Yes, have seen this too many times with unix domain sockets.

PR #393, opened 2022-01-08
gee: ignore "apt update" errors

jonathan-enf:
gee: ignore "apt update" errors

Some of our installations have stale/broken links to certain launchpad
repos ( https://enfabrica.atlassian.net/browse/INFRA-383 ).  This causes
"sudo apt update" to fail with exit code 100, which then causes gee to
fail when trying to install missing tools.  gee now works around this
issue by ignoring all exit codes from "apt update."

Tested:
Ran "gee repair" on hw-dev-00, which has a broken repo configured.

PR generated by jonathan from branch gee_workaround_apt.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -993,7 +993,8 @@ function _install_tools() {
       | sudo /usr/bin/apt-key --keyring "${KRFILE}" add -
     echo "deb [arch=amd64 signed-by=${KRFILE}] https://cli.github.com/packages stable main" \
       | sudo /usr/bin/tee /etc/apt/sources.list.d/githubcli-archive.list
-    sudo /usr/bin/apt update && sudo /usr/bin/apt install gh
+    sudo /usr/bin/apt update || /bin/true  # ignore errors associated with stale repos.
+    sudo /usr/bin/apt install gh
     if [ ! -x ${GH} ]; then
       _fatal "Could not install gh."
     fi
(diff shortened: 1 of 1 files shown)


scripts/gee
```diff
@@ -993,15 +993,17 @@ function _install_tools() {
       | sudo /usr/bin/apt-key --keyring "${KRFILE}" add -
     echo "deb [arch=amd64 signed-by=${KRFILE}] https://cli.github.com/packages stable main" \
       | sudo /usr/bin/tee /etc/apt/sources.list.d/githubcli-archive.list
-    sudo /usr/bin/apt update && sudo /usr/bin/apt install gh
+    sudo /usr/bin/apt update || /bin/true  # ignore errors associated with stale repos.
```
ccontavalli, line 996: not objecting, but for my own understanding: why do you specify the full path for something like /bin/true?

(also apt, to be frank, or tee - eg, full path is a double edged sword: works without PATH, yay, but things do move around, so eg, even across ubuntu versions I've seen the path of pretty basic shell commands change - was recently bit by one)

ccontavalli, line 996: also minor nit: I've seen `apt-get` still being preferred in shell script.

This [not very loved stackoverflow answer](https://unix.stackexchange.com/questions/590699/should-i-use-apt-or-apt-get-in-shell-scripting) provides the rationale so I don't have to type it (source: apt man page).

ccontavalli (approved):
You don't need to address either comment, nits/question.

scripts/gee
jonathan-enf: It's a good security hygiene habit.  One of common attack on scripts is to insert a malicious program into the path and trick a script to run it with elevated privileges.

Doesn't matter for gee, which doesn't run with any special privileges, but I've just gotten into the habit of trying to do things the secure way.

jonathan-enf: Done.

minor-fixes (commented):
Any reason to ignore all error codes instead of just code 100?

minor-fixes (approved):

PR #394, opened 2022-01-08
bestie: Add test metrics processing to BEP server

gpaussaenf:
This change allows the BEP server to process TestResult
events containing test metrics presented in the
test_metrics.pb file that is incorporated in the
outputs.zip file. The test_metrics.proto defines
the protobuf message format, which reflects the
data format needed by BigQuery for Prometheus consumption.

A follow on PR is needed to continue this work, taking
the test metric data and storing it in BigQuery. For now,
the BEP server continues to display the Bazel event info
same as before, and additionally displays the relevant
metric information it finds in the TestResult events.

Tested: Started server, ran 'bazel build //bestie/...
--bes_backend=grpc://localhost:6433' and saw messages
printed on stdout; bazel exited with no error.

Jira: INFRA-326


bestie/server/unzipper.go
```diff
@@ -0,0 +1,84 @@
...
+		if err := os.MkdirAll(filePath, os.ModePerm); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	// For a file element, make sure its path exists.
+	if err := os.MkdirAll(filepath.Dir(filePath), os.ModePerm); err != nil {
+		return err
+	}
```
minor-fixes, line 64: I suspect the biggest improvement that can be made to this code is to not actually unzip the file via Go - instead, we can iterate over the zip to find interesting files, and selectively either read those files into memory, or spill them to temporary files to read. This would help by:
* not having to write/maintain essentially a fork of the zip tool in Go
* not having to lock/unlock in the service (since the service won't contend on the same destination paths)
* not having to worry about overly large zips
* not having to worry about cleaning up zips

Maybe this can be replaced with a function that:
* iterates through the zip file headers
* if the current zip file header is an interesting file (metrics.pb) - do something with it
  * maybe read the file contents into a buffer, and then unmarshal the pb message

minor-fixes (commented):
Setup looks good! The code to unzip a file tree in any language is always pretty hairy, so nice job working through that.

Let's figure out the best way to make the unzipping work long-term (added a comment below)

bestie/server/unzipper.go
aaahrens: @gpaussaenf a toy example of this is https://gist.github.com/indraniel/1a91458984179ab4cf80

gpaussaenf: Thanks for the link. Looks like `tarReader.Read(data)` directly into a []byte slice might be what I need.

gpaussaenf: Made this change in the latest commit.

bestie/proto/test_metrics.proto
```diff
@@ -0,0 +1,60 @@
...
+//    tablename - name of the table to hold these metrics; the same tablename is used across datasets
+message BigQueryTable {
+    string project = 1;
+    string dataset = 2;
+    string tablename = 3;
+}
+
+// BigQuery metric definition compatible with Prometheus remote read adapter.
+//
+// The 'tags' field is a freeform, JSON-like string with comma-separated key=value pairs.
```
ccontavalli, line 20: should we use something like:

```
repeated KeyValue tags;
```
with:
```
message KeyValue {
  string key = 1;
  string value = 2;
}
```
?

We can then turn it into a single json field for bigquery easily, or do other automated processing more easily in the ingestion pipeline if necessary?

(this also guarantees that the value is syntactically valid, and saves the caller from having to implement any form of escaping, which is error prone. Eg, let's say otherwise this field contains invalid values, how do we return and inform the caller of an error? Hard to do from the BES endpoint, we risk having a silent error cause data to be discarded, hard to debug)

bestie/proto/test_metrics.proto
```diff
@@ -0,0 +1,60 @@
...
+message TestMetric {
+    string metricname = 1;
+    string tags = 2;
+    double value = 3;
+    int64 timestamp = 4;  // nanoseconds since epoch (e.g. Python: time.time_ns())
+}
+
+// A set of metrics reported by a test run to be stored in the designated BigQuery table.
+message TestMetrics {
+    BigQueryTable table = 1;
```
ccontavalli, line 58: as mentioned at standup, I would personally make this optional (eg, there's a default, and tests don't normally set it) or remove it entirely. Probably a preference to remove it entirely.

Rationale: let's say the database does not scale, or prometheus explodes with too much data in a single table. We then need to "shard" the data, split it up in multiple tables. We don't want to go around and change each test / metric file to include this information, it'd be a large refactor on the code base? And force the author of the test to think about how to ... adapt his test to make it work with our metric infrastructure?
We'd also need to manually migrate the data, if we care about preserving history.

A better approach, imho, would be for the BES backend itself to have some internal policy to shard the data. For example, it could use different tables based on the directory of the test, or based on the test name prefix, or ... use tags in BUILD files (so the table becomes a "configuration parameter of the test", controlled in BUILD.bazel file), rather than a code change in the test. (those are just examples)

bestie/proto/test_metrics.proto
gpaussaenf: Agreed. Made this change in the latest commit.

minor-fixes (approved):

bestie/server/testresult.go
```diff
@@ -0,0 +1,220 @@
...
+		fmt.Printf("Found output file to process: %s\n", file.Name)
+
+		f, err := file.Open()
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+
+		// Read entire file contents into a byte slice.
+		data, err := ioutil.ReadAll(f)
```
minor-fixes, line 156: We need to explicitly bound the metrics filesize, or come up with a method of streaming this from an io.Reader instead of forcing the whole thing to be loaded into memory (in case someone generates e.g. 1GB of metrics)

bestie/server/testresult.go
```diff
@@ -0,0 +1,220 @@
...
+	if err := extractZippedFiles(stream, ofuri); err != nil {
+		return fmt.Errorf("Error processing %s file: %s", ofname, err)
+	}
+
+	return nil
+}
+
+// Look for and return a []byte slice with its contents.
+func extractZippedFiles(stream bazelStream, zipFile string) error {
+	metricsFileRE := regexp.MustCompile(`(^test_metrics.pb$|^test_metrics[_-]+[[:word:]-]*[[:alnum:]]\.pb$)`)
```
minor-fixes, line 121: We should pick a convention that's easier to check for (e.g. "file must end in `.metrics.pb`") to simplify our regex checking

bestie/server/testresult.go
```diff
@@ -0,0 +1,220 @@
...
+	summaryFileRE := regexp.MustCompile(`(^test_summary.pb$|^test_summary[_-]+[[:word:]-]*[[:alnum:]]\.pb$)`)
```
minor-fixes, line 122: What is a summary file - how is it different than a metrics file?

bestie/server/testresult.go
```diff
@@ -0,0 +1,220 @@
...
+			// TODO: Upload test metrics to BigQuery database table.
+		}
+	}
+	return nil
+}
+
+// Extract the test metrics information from the protobuf message data read
+// from an output file. This is used for processing test_metrics*.pb and
+// test_summary*.pb output files.
+func getTestMetricsFromFileData(pbmsg *[]byte) (*metricTestResult, error) {
```
minor-fixes, line 179: nit: slices don't need to be passed by pointer if they're only read (only if you want to append to it - but in that case it might be better to return a slice rather than have effectively a slice outparam)

bestie/server/testresult.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	if !found {
+		return nil
+	}
+
+	// Strip off any file:// prefix from the URI to access the local file system path.
+	filePrefix := "file://"
+	if strings.HasPrefix(ofuri, filePrefix) {
+		ofuri = ofuri[len(filePrefix):]
+	}
```
minor-fixes, line 108: This almost certainly won't work when we put this up into Cloud Run, because:
* remote build URLs are going to not start with `file://`, but this code will fail to strip the URL prefix, and then try to load the file on-disk
* local builds file URLs won't make sense to the BES endpoint, which is on another system and therefore doesn't have the same filesystem

...but we can fix that later.

minor-fixes (commented):
Nothing blocking here - just notes for the future

bestie/server/testresult.go
gpaussaenf: OK, I'll change this in the current PR before merging.

bestie/server/testresult.go
gpaussaenf: The intent is to put actual test metrics in a "test_metrics*.pb" file and the individual test case pass/fail summaries in a "test_summary*.pb" file. They are formatted the same, but the pass/fail summary "metrics" use a fixed metric name of 'testresult' and a value of 1.0 or 0.0 to indicate pass vs. fail, staying consistent with both the protobuf message definition and the BigQuery table schema, since we want these all to land in the same BigQuery test metrics table. With the regex change mentioned in the previous comment, the actual file names don't need to be as rigid.

bestie/server/testresult.go
gpaussaenf: OK, will fix in this PR.

bestie/server/testresult.go
gpaussaenf: We'll need to address the Cloud Run case later. I could see how the URL was formatted for local builds, but wasn't sure about how this would be formatted for remote builds. That's why I added the HasPrefix check to cover the local case, knowing there will likely be some similar adjustment needed for the remote case. Definitely need to revisit this later on.

bestie/server/testresult.go
```diff
@@ -133,14 +132,8 @@ func extractZippedFiles(stream bazelStream, zipFile string) error {
...
-		//  Look for a matching file name using regular expression patterns.
-		match := ""
-		for _, re := range allRE {
-			match = re.FindString(filepath.Base(file.Name))
-			if len(match) > 0 {
-				break
-			}
-		}
+		//  Look for a matching file name using a regular expression pattern.
+		match := pbFileRE.FindString(filepath.Base(file.Name))
```
minor-fixes, line 136: ...this then becomes `strings.HasSuffix(filename, ".metrics.pb")` :)

gpaussaenf: Yeah, I was too fixated on the regex that I didn't consider the simpler approach. Will change this in my next PR, post merge of this PR.

PR #402, opened 2022-01-18
faketree: make work in remote tests

aaahrens:
Currently the test for fake assumes that /etc is readable by the executing user. this is not the case in most CI environments.

This replaces the dependencies in /etc with random /tmp ones.

### DIFF
--- a/faketree/faketree_test.sh
+++ b/faketree/faketree_test.sh
@@ -30,8 +30,8 @@ test "$?" == 0 || {
 test "$uid" == "0" || {
   fail "running as root failed - $uid"
 }
-
-dir=$($ft --mount /etc:/tmp/root/etc --chdir /tmp/root -- sh -c pwd)
+tmpdir=$(mktemp -d -t faketree-XXXXXXXXXX)
+dir=$($ft --mount $tmpdir:/tmp/root/etc --chdir /tmp/root -- sh -c pwd)
 test "$?" == 0 || {
   fail "unexpected failure in simple mount"
 }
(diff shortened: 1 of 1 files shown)


faketree/faketree_test.sh
```diff
@@ -30,31 +30,32 @@ test "$?" == 0 || {
...
-$ft --mount /etc:/tmp/root/etc --chdir /tmp/root/etc -- sh -c 'exit 12' &>/dev/null
+$ft --mount $tmpdir:/tmp/root/etc --chdir /tmp/root/etc -- sh -c 'exit 12' &>/dev/null
 test "$?" == 12 || {
   fail "faketree was suppsoed to propagate error 12"
 }
 
-$ft --mount /etc:/tmp/root/etc --chdir /tmp/root/etc -- sh -c 'cat ./hosts' &>/dev/null
+tmpfile=$(mktemp $tmpdir/faketree.XXXXXX)
+cmd="cat ./$tmpfile"
+$ft --mount $tmpdir:/tmp/root/etc --chdir /tmp/root/etc -- sh -c $cmd &>/dev/null
```
ccontavalli, line 56: nit: use `"$cmd"` rather than just `$cmd` here.

(the expansion of $cmd should be a single argument to `sh -c`, as is $cmd contains a space, so it will be two arguments. Eg, without "" argv ends up as 0: sh, 1: -c 2: cat 3: ./$tmpfile, rather than 0: sh, 1: -c 2: "cat ./$tmpfile" - 4 arguments rather than 3 - in both cases $tmpfile would be expanded correctly, in facts it is expanded on line 55)

ccontavalli (approved):
thank you! Very necessary!

PR #406, opened 2022-01-21
bestie: Add initial BES Endpoint service metrics

gpaussaenf:
Defined the following metrics for monitoring the BES Endpoint
service through Prometheus and Grafana:

- total number of Bazel builds seen
- count of each Bazel event, tagged by event ID

Also replaced the regex with a much simpler check for any test
output file ending with ".metrics.pb" to be processed for
test result metrics.

Tested: 'bazel run //bestie/server:bestie' to run BES Endpoint
locally on port 6433. Ran a 'bazel test' client command multiple
times, with --bes_backend=grpc://localhost:6433, to generate a
variety of Bazel events through the endpoint. Set up a local
Grafana dashboard to observe Prometheus metrics being updated.

Jira: INFRA-438

### DIFF
--- a/bestie/server/service.go
+++ b/bestie/server/service.go
@@ -1,0 +1,133 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// List of event ID names defined by build_event_stream.proto.
+// These formal event names are mapped to Prometheus metric label names.
+var besEventIds = map[string]string{
+	"Unknown":                       "unknown", // should not normally see any of these
+	"Progress":                      "progress",
+	"Started":                       "started",
+	"CommandLine":                   "unstructured_command_line", // CommandLine replaced by UnstructuredCommandLine
+	"UnstructuredCommandLine":       "unstructured_command_line",
+	"StructuredCommandLine":         "structured_command_line",
+	"WorkspaceStatus":               "workspace_status",
+	"OptionsParsed":                 "options_parsed",
+	"Fetch":                         "fetch",
+	"Configuration":                 "configuration",
+	"TargetConfigured":              "target_configured",
+	"Pattern":                       "pattern",
+	"PatternSkipped":                "pattern_skipped",
+	"NamedSet":                      "named_set",
+	"TargetCompleted":               "target_completed",
+	"ActionCompleted":               "action_completed",
+	"UnconfiguredLabel":             "unconfigured_label",
+	"ConfiguredLabel":               "configured_label",
+	"TestResult":                    "test_result",
+	"TestSummary":                   "test_summary",
+	"TargetSummary":                 "target_summary",
+	"BuildFinished":                 "build_finished",
+	"BuildToolLogs":                 "build_tool_logs",
+	"BuildMetrics":                  "build_metrics",
+	"Workspace":                     "workspace",
+	"BuildMetadata":                 "build_metadata",
+	"ConvenienceSymlinksIdentified": "convenience_symlinks_identified",
+}
+
+// BES Prometheus metric definition.
+// Use this for untagged counters.
+type besPromMetric struct {
+	label string
+	ctr   prometheus.Counter
+}
+
+// BES Prometheus metric vector definition.
+// Use this for tagged counters.
+type besPromMetricVec struct {
+	ctr *prometheus.CounterVec
+}
+
+// BES Prometheus histogram definition.
+type besPromHistogram struct {
+	ctr prometheus.Histogram
+}
+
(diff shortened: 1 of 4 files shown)


bestie/server/service.go
```diff
@@ -0,0 +1,133 @@
...
+type besPromHistogram struct {
+	ctr prometheus.Histogram
+}
+
+// Number of BES messages seen by event type.const
+type besEvents struct {
+	bazelBuildsTotal besPromMetric
+	bazelEventsTotal besPromMetricVec
+}
+
```
gpaussaenf, line 65: This file is intentionally structured this way to allow for the easy addition of BigQuery service metrics in the PR that's coming for INFRA-423.

bestie/server/service.go
```diff
@@ -0,0 +1,133 @@
...
+// Increment per-event count.
+func (s *serviceStats) bazelEventsTotal(bevid interface{}) {
+	x := strings.Split(fmt.Sprintf("%T", bevid), "_")
+	id := x[len(x)-1] // use last split item
+	label := besEventIds["Unknown"]
+	if promId, ok := besEventIds[id]; ok {
+		label = promId
+	} else {
+		// Make note of this condition. Probably means .proto file definition was updated.
+		fmt.Printf("Detected unknown event id: %s\n\n", id)
```
minor-fixes, line 130: nit: Not necessary to log - we'll catch it in the metrics

bestie/server/service.go
```diff
@@ -0,0 +1,133 @@
...
+func (s *serviceStats) init() {
+	s.registerPrometheusMetrics()
+}
+
+//
+// Accessor methods to update Prometheus metrics.
+//
+
+// Increment total number of Bazel builds seen.
+func (s *serviceStats) bazelBuildsTotal() {
```
minor-fixes, line 117: nit: Start these functions with a verb: `incrementTotalBuilds()`, `incrementTotalEvents()`

bestie/server/service.go
```diff
@@ -0,0 +1,133 @@
...
+type besEvents struct {
+	bazelBuildsTotal besPromMetric
+	bazelEventsTotal besPromMetricVec
+}
+
+// Define BES Endpoint service counters being exposed through Prometheus.
+var (
+	promBazelBuildsTotal = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: "bes",
```
minor-fixes, line 70: Namespace == bestie?

bestie/server/service.go
```diff
@@ -0,0 +1,133 @@
...
+			Subsystem: "srv",
```
minor-fixes, line 71: Do we know what the subsystems will be? If not, let's omit it for now until we figure out the best way to use that field

bestie/server/service.go
```diff
@@ -0,0 +1,133 @@
...
+	)
+)
+
+// Service statistics.
+type serviceStats struct {
+	// General Build Event Stream stats.
+	besEvents
+}
+
+// Statistic values to report to Prometheus for this service.
```
minor-fixes, line 92: Why all the structs wrapping the metric types, only to declare the metrics as global? Why not use the global metrics directly?

e.g. https://github.com/enfabrica/enkit/blob/f61a0f0aa15fd7cf74fa0f94af10c43cc4499ab6/flextape/service/service.go#L21 defines some global metrics, that the service increments directly. This has extra wrapper structs, but global metrics are still used under the hood - what value do the wrapper structs provide?

bestie/server/service.go
gpaussaenf: The "unknown" metric will increment, but we won't know what was the unknown metric ID that caused it.

bestie/server/service.go
gpaussaenf: Will do.

bestie/server/service.go
gpaussaenf: OK

bestie/server/service.go
gpaussaenf: Initial thought is "srv" for general service metrics, "bq" for BigQuery-related service metrics, etc. But I can remove Subsystem and prefix the Name field with "bq_" to identify those.

bestie/server/service.go
gpaussaenf: I was thinking that there might be other information related to each metric in addition to the actual Prometheus metric type itself. For example, the metric name used by Prometheus. I'll take another look at this to see if the wrapper provides any meaningful value-add.

bestie/server/service.go
gpaussaenf: I left this in for now.

minor-fixes (approved):

PR #408, opened 2022-01-25
bestie: kbuildbarn url parser

aaahrens:
The way to translate between bytestream urls and how to fetch them from buildbarn is a little obfuscatory. This is an attempt to document some of the current methods of translating betwene the two in code.

### DIFF
--- a/lib/kbuildbarn/README.md
+++ b/lib/kbuildbarn/README.md
@@ -1,0 +1,4 @@
+This package is designed for integration between buildbarn and other first party applications written in enkit.
+
+Buildbarn and other remote executors format their external tools around already defined .proto definitions. This library
+is designed to be configurable based on external templates.
--- a/lib/kbuildbarn/README.md
+++ b/lib/kbuildbarn/README.md
@@ -2,3 +2,17 @@ This package is designed for integration between buildbarn and other first party
 
 Buildbarn and other remote executors format their external tools around already defined .proto definitions. This library
 is designed to be configurable based on external templates.
+
+Examples:
+
+
+    MyApplication --buildbarn-dsn=http://foo.internal
+    Delivered: bytestream://127.0.0.1/blobs/curry/444
+    ====== 
+    hash, size, err := kbuildbarn.ByteStreamUrl("bytestream://127.0.0.1/blobs/curry/444")
+
+    bbparam := kbuildbarn.NewBuildBarnParams("http://foo.internal", "foo.bar", hash, size)
+    
+    ======
+    MyApplication.Upload(bbparams.FileUrl())
+
--- a/lib/kbuildbarn/urls.go
+++ b/lib/kbuildbarn/urls.go
@@ -10,7 +10,7 @@ import (
 	"strings"
 )
 
-// ByteStreamUrl retieves the CAS id and bytes of an action based on the input url.
+// ByteStreamUrl retrieves  the CAS id and bytes of an action based on the input url.
 // For example, bytestream://build.local.enfabrica.net:8000/blobs/a9a664559b4d29ecb70613fad33acfb287f2fa378178e131feaaebb5dafa231a/465
 // should return (a9a664559b4d29ecb70613fad33acfb287f2fa378178e131feaaebb5dafa231a, 465, nil)
 // which is a BuildBarnParams.Hash, BuildBarnParams.Size, error.Error
@@ -38,10 +38,10 @@ type BuildBarnParams struct {
 	Scheme  string
 
 	// These are the default buildbarn templates for their different types inside the CAS
-	FileTemplate      string
 	ActionTemplate    string
 	CommandTemplate   string
 	DirectoryTemplate string
+	FileTemplate      string
 }
 
 // the following default values are arbitrary, based on what current works with buildbarn
@@ -52,6 +52,12 @@ var (
 	DefaultDirectoryTemplate = "/blobs/directory/%s-%s/"
 )
 
+// NewBuildBarnParams generates a new Buildbarn url translator for other apis.
+// The intended use case is something like RetryUntilSuccess where:
+// The application retrieves a dead bytestream//: url, and extracts the artifact information from that using ByteStreamUrl
+// The external application can then use baseurl (which changes based on the location of the buildbarn instance),
+// fileName (which is optional) and the extracted hash and size to generate valid urls via BuildBarnParams.ActionUrl etc.
+// External applications can use them e.g. save the now valid urls; load balance onto the urls that are valid, etc.
 func NewBuildBarnParams(baseUrl, fileName, hash, size string) *BuildBarnParams {
 	// These are prefilled defaults, we can change at will
 	return &BuildBarnParams{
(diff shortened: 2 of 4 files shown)


lib/kbuildbarn/urls.go
```diff
@@ -0,0 +1,130 @@
...
+	Hash         string
+	Size         string
+	InvocationID string
+
+	// This is the base Url
+	BaseUrl string
+	Scheme  string
+
+	// These are the default buildbarn templates for their different types inside the CAS
+	FileTemplate      string
```
gpaussaenf, line 41: nit: Suggest listing these four items in alphabetical order for ease of future maintenance. Likewise for the Default templates and in NewBuildBarmParams().

lib/kbuildbarn/urls.go
```diff
@@ -0,0 +1,130 @@
...
+		params.FileUrl(), params.ActionUrl(), params.CommandUrl(), params.DirectoryUrl(),
+	}
+	var errs []error
+	for _, uri := range urls {
+		rc, err := performRequest(http.DefaultClient, uri)
+		if err != nil {
+			errs = append(errs)
+			continue
+		}
+		return ioutil.ReadAll(rc)
```
gpaussaenf, line 127: By doing a ReadAll() and returning a byte slice, what happens to the ReadCloser object returned by performRequest()? The caller won't see it, and there is no Close() (deferred or otherwise) being done here.

lib/kbuildbarn/urls.go
```diff
@@ -0,0 +1,130 @@
...
+		Scheme: bbp.Scheme,
+		Host:   bbp.BaseUrl,
+		Path:   fmt.Sprintf(bbp.CommandTemplate, bbp.Hash, bbp.Size),
+	}
+	return u.String()
+}
+
+// RetryUntilSuccess just blasts through all possible urls until it hits one that works. This is intended for
+// applications that are blind to the type of artifact
+func RetryUntilSuccess(params BuildBarnParams) ([]byte, error) {
```
gpaussaenf, line 116: Is the intent that this is the only public API to be used by applications? What if the application already knows which template type it is interested in (e.g. File). Since performRequest() is private, applications don't have the option of working from their own ReadCloser object -- is that needed?

lib/kbuildbarn/README.md
```diff
@@ -0,0 +1,4 @@
+This package is designed for integration between buildbarn and other first party applications written in enkit.
+
+Buildbarn and other remote executors format their external tools around already defined .proto definitions. This library
+is designed to be configurable based on external templates.
```
gpaussaenf, line 4: It would be helpful to give a short example of intended API usage in this readme to end up with a []byte slice from the input bytestream:// URL. Application developers are more likely to look here than in the unit test file for this info.

lib/kbuildbarn/urls.go
```diff
@@ -0,0 +1,130 @@
...
+func ByteStreamUrl(byteStream string) (string, string, error) {
+	u, err := url.Parse(byteStream)
+	if err != nil {
+		return "", "", err
+	}
+	splitUrl := strings.Split(u.Path, "/")
+	if len(splitUrl) != 4 {
+		return "", "", fmt.Errorf("ByteStreamUrl() bytestream url is not well formed %s", byteStream)
+	}
+	return splitUrl[2], splitUrl[3], nil
```
gpaussaenf, line 26: Shouldn't this function also return the base URL that's needed by NewBuildBarnParams()? How else can the application get it from the original bytestream: URL?

gpaussaenf (commented):
Overall, looks good and will be helpful for the BES Endpoint to use this. There are a few comments to address before I can approve.

lib/kbuildbarn/urls.go
aaahrens: Yup, I need to close the request body

lib/kbuildbarn/urls.go
aaahrens: Done

lib/kbuildbarn/urls.go
aaahrens: The general idea is that the external application knows what to do with the generated URL. RetryUntilSuccess is a function that sits on top of that abstraction layer that's pretty much "i don't care about the type, just give me whatever is in this bytestream url"

lib/kbuildbarn/urls.go
aaahrens: The Buildbarn base url is not necessarily what's in the bytestream URL.
For example, if you communicate over ssh to remote exec, the url in the bytestream url is 127.0.0.1. If it's via unix sockets the url is a file <.<. Neither are where the CAS is stored.

lib/kbuildbarn/README.md
aaahrens: Added more docs in both the readme and the New factor constructor

ccontavalli:
uhm .. so, this PR provides code to turn "bytestream://..." urls into "http" URLs that work with buildbarn.

But... isn't buildbarn exposing a bytestream protocol endpoint? Eg, from what I read online (quick google search, so maybe wrong), bytestream:// is basically a gRPC based protocol. Using a client like `https://pkg.go.dev/google.golang.org/api/transport/bytestream` you can access bytestream urls without converting them to http? and without having custom logic to parse the url?

Not sure, but if that's the case, the string "bytestream://build.local.enfabrica.net:8000/blobs/a9a664559b4d29ecb70613fad33acfb287f2fa378178e131feaaebb5dafa231a/465" would allow to connect to `build.local.enfabrica.net:8000` via gRPC, and just fetch that resource by full path, and buildbarn would return it?

AFAIK, things like buildbuddy don't need to code support for different bazel clustering, eg, they don't need a library to convert those urls for buildbarn, for buildfarm, for....

ccontavalli:
at first sight, buildbarn does implement a bytestream endpoint, see file: https://github.com/buildbarn/bb-storage/blob/master/pkg/blobstore/grpcservers/byte_stream_server.go#L36.

The Read() function then invokes [NewDigestFromBytestreamReadPath()](https://github.com/buildbarn/bb-storage/blob/cd279b287defd412b65f29d6d2313c5fe9ed145f/pkg/digest/digest.go#L123) that does a very similar thing to the code here (splits the path by /, looks for "blobs", then takes the hash and value).

aaahrens:
You could theoretically do a read from the grpc endpoint. There is no advantage to doing it because 
1. The host and port still need to be rewritten. *.local.** always goes to loopback. Ports are different than the host. Intranet builds will even have randomized ports. Rewrite work still needs to be done. 
2. The rpc endpoint is not more efficient because CAS downloads are not chunked out.

However just rewriting to http has the following benefits:
1. human readable and accessible from any browser. it's also shortenable.
2. can be used outside of intra cluster workloads agnostically.
3. Does not require passing around a storage grpc client. 

And it looks suspiciously similar to ReadDigest because I had to trawl though the source code to figure out how the urls were constructed <.<

aaahrens:
And FTR buildbuddy does need code like this to access artifacts. This code is live on the shim I wrote for the /file/download endpoint, which I hijack at L7.

gpaussaenf (approved):
The latest revisions look good. Thanks for updating the README. I now realize that the point I was missing is that the baseUrl comes into my code as a command line parameter.

Also FYI, I tried using the `google.golang.org/api/transport/bytestream` API in some sample code and found it to be very confusing (and not many usable examples out there to learn from). I much prefer the translation and access functions provided in this PR -- I think it'll simplify the BES Endpoint code.

ccontavalli:
> The latest revisions look good. Thanks for updating the README. I now realize that the point I was missing is that the baseUrl comes into my code as a command line parameter.
> 
> Also FYI, I tried using the `google.golang.org/api/transport/bytestream` API in some sample code and found it to be very confusing (and not many usable examples out there to learn from). I much prefer the translation and access functions provided in this PR -- I think it'll simplify the BES Endpoint code.

Sounds good.

PR #409, opened 2022-01-25
machinist: add _info endpoint

aaahrens:
Adds a feature where people can see what machines are currently in the dns server by calling the _info record

### DIFF
--- a/machinist/mserver/controller.go
+++ b/machinist/mserver/controller.go
@@ -145,6 +145,7 @@ func (en *Controller) addNodeToDns(name string, ips []net.IP, tags []string) {
 
 // ServeAllAndInfoRecords will continuously poll Nodes() and create multiple _all.<domain> records containing the ip addresses
 // of all machines attached.
+// It also serves the current state of the dns server via the _info record.
 // TODO(adam): be able to pass in a wrapped ticker for testing intervals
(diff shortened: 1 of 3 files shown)


machinist/mserver/controller.go
```diff
@@ -143,27 +143,36 @@ func (en *Controller) addNodeToDns(name string, ips []net.IP, tags []string) {
...
-func (en *Controller) ServeAllRecords(killChannel chan struct{}, killChannelAck chan struct{}) {
+func (en *Controller) ServeAllAndInfoRecords(killChannel chan struct{}, killChannelAck chan struct{}) {
 	for {
 		select {
 		case <-time.After(en.allRecordsRefreshRate):
 			ns := en.Nodes()
 			for _, d := range en.dnsServer.Domains {
 				dnsName := dns.CanonicalName(fmt.Sprintf("%s.%s", "_all", d))
-				var rs []dns.RR
+				infoDnsName := dns.CanonicalName(fmt.Sprintf("%s.%s", "_info", d))
```
gpaussaenf, line 157: Seems like this section could be streamlined using another for loop that iterates over []string{"all", "info"}. The only thing that varies based on the current iterator value is the string format used for the call to dns.NewRR().

gpaussaenf (approved):

PR #417, opened 2022-01-28
Bazel FUSE: Add enkit commands

minor-fixes:
This change adds enkit subcommands for bazel artifacts mounting, as laid
out in
https://docs.google.com/document/d/1L9skX1BKY-AZ15GVypQWxiy360UkUo5mNXglC9vvO9g/edit#.

These commands will need to tie together functionality from faketree,
bb_clientd management, and zipfile/tempdir management, so unlike other
enkit subcommand trees there is no pre-existing logical source location
to put the commands themselves. For now, they are added as a package
under `enkit` itself.

Tested: Output for help of each command:

`enkit outputs mount`:

```
Mount the build outputs of a particular invocation

Usage:
  enkit outputs mount [invocation ID] [flags]

Examples:
  $ enkit outputs mount 73d4a9f0-a0c4-4cb2-80eb-b4b4b9720d07
        Mounts outputs from build 73d4a9f0-a0c4-4cb2-80eb-b4b4b9720d07 to the
        default location.

Flags:
  -h, --help   help for mount

Global Flags:
      --help-all              Show all the flags available, even those that are less useful and normally hidden.
      --outputs_root string   Root dir of mounted outputs (default "/home/bminor/outputs")
```

`enkit outputs unmount`:

```
Unmount the build outputs of a particular invocation

Usage:
  enkit outputs unmount [invocation ID] [flags]

Aliases:
  unmount, umount

Examples:
  $ enkit outputs unmount 73d4a9f0-a0c4-4cb2-80eb-b4b4b9720d07
        Unmounts outputs from build 73d4a9f0-a0c4-4cb2-80eb-b4b4b9720d07 from the
        default location.

Flags:
  -h, --help   help for unmount

Global Flags:
      --help-all              Show all the flags available, even those that are less useful and normally hidden.
      --outputs_root string   Root dir of mounted outputs (default "/home/bminor/outputs")
```

`enkit outputs run`:

```
Mount build artifacts and execute an optional command on them

Usage:
  enkit outputs run [flags]

Examples:
  $ enkit outputs run --invocation=73d4a9f0-a0c4-4cb2-80eb-b4b4b9720d07
        Launches a shell with build outputs from a particular build re-rooted so that
        paths are correct within the outputs themselves.

  $ enkit outputs run --dir=/tmp/some_dir
        Launches a shell with build outputs in /tmp/some_dir re-rooted so that paths
        are correct within the outputs themselves.

  $ enkit outputs run --zip=/tmp/some.zip
        Unpacks the named zip into a tempdir, and then reroots the artifacts so that
        paths are correct within the outputs themselves.

  $ enkit outputs run --zip=/tmp/some-zip -- find .
        Runs "find ." in the unpacked zip after it is rerooted.

Flags:
      --dir_path string        If set, the path to already-unpacked artifacts to reroot
  -h, --help                   help for run
      --invocation_id string   If set, the build's invocation ID from which to mount artifacts
      --zip_path string        If set, the path to a zipped outputs file to unpack and re-root

Global Flags:
      --help-all              Show all the flags available, even those that are less useful and normally hidden.
      --outputs_root string   Root dir of mounted outputs (default "/home/bminor/outputs")
```

`enkit outputs shutdown`:

```
Unmount all builds under particular directory

Usage:
  enkit outputs shutdown [flags]

Examples:
  $ enkit outputs shutdown
        Unmounts all builds in the given output root and resets the output root to a pristine state.

Flags:
  -h, --help   help for shutdown

Global Flags:
      --help-all              Show all the flags available, even those that are less useful and normally hidden.
      --outputs_root string   Root dir of mounted outputs (default "/home/bminor/outputs")
```

Jira: INFRA-469
Jira: INFRA-488

### DIFF
--- a/enkit/outputs/commands.go
+++ b/enkit/outputs/commands.go
@@ -174,7 +174,7 @@ func NewShutdown(root *Root) *Shutdown {
 			Use:   "shutdown",
 			Short: "Unmount all builds under particular directory",
 			Example: `  $ enkit outputs shutdown
-	Unmounts bb_clientd, and deletes all links to build artifacts.`,
+	Unmounts all builds in the given output root and resets the output root to a pristine state.`,
 		},
 		root: root,
 	}
--- a/enkit/outputs/commands.go
+++ b/enkit/outputs/commands.go
@@ -1,0 +1,187 @@
+package outputs
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/enfabrica/enkit/lib/client"
+	"github.com/spf13/cobra"
+)
+
+type Root struct {
+	*cobra.Command
+	*client.BaseFlags
+
+	OutputsRoot string
+}
+
+func New(base *client.BaseFlags) (*Root, error) {
+	root, err := NewRoot(base)
+	if err != nil {
+		return nil, err
+	}
+
+	root.AddCommand(NewMount(root).Command)
+	root.AddCommand(NewUnmount(root).Command)
+	root.AddCommand(NewRun(root).Command)
+	root.AddCommand(NewShutdown(root).Command)
+
+	return root, nil
+}
+
+func NewRoot(base *client.BaseFlags) (*Root, error) {
+	rc := &Root{
+		Command: &cobra.Command{
+			Use:           "outputs",
+			Short:         "Commands for mounting remotely-built Bazel outputs",
+			SilenceUsage:  true,
+			SilenceErrors: true,
+			Long:          `outputs - commands for mounting remotely-built Bazel outputs`,
+		},
+		BaseFlags: base,
+	}
+
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return nil, fmt.Errorf("failed to detect $HOME: %w", err)
+	}
+	defaultOutputsRoot := filepath.Join(homeDir, "outputs")
+
+	rc.PersistentFlags().StringVar(&rc.OutputsRoot, "outputs_root", defaultOutputsRoot, "Root dir of mounted outputs")
+	return rc, nil
+}
+
+type Mount struct {
+	*cobra.Command
+	root *Root
+}
+
+func NewMount(root *Root) *Mount {
+	command := &Mount{
+		Command: &cobra.Command{
+			Use:   "mount [invocation ID]",
+			Short: "Mount the build outputs of a particular invocation",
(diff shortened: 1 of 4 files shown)


enkit/outputs/commands.go
```diff
@@ -0,0 +1,187 @@
...
+	root *Root
+}
+
+func NewShutdown(root *Root) *Shutdown {
+	command := &Shutdown{
+		Command: &cobra.Command{
+			Use:   "shutdown",
+			Short: "Unmount all builds under particular directory",
+			Example: `  $ enkit outputs shutdown
+	Unmounts bb_clientd, and deletes all links to build artifacts.`,
```
ccontavalli, line 177: remove references to bb_clientd, we are using bb_clientd now as implementation, but our average developer may not know / will likely not have direct visibility into it?

enkit/outputs/commands.go
```diff
@@ -0,0 +1,187 @@
...
+
+type Shutdown struct {
+	*cobra.Command
+	root *Root
+}
+
+func NewShutdown(root *Root) *Shutdown {
+	command := &Shutdown{
+		Command: &cobra.Command{
+			Use:   "shutdown",
```
ccontavalli, line 174: shutdown -> are we using a single mount? or multiple mounts? also, "deletes all links"...

In the document I was proposing using "scratch/" for *hard links* that are hard for tools to distinguish. Is that still the plan?

I was proposing using a bb_clientd mount per invocation as I don't know if scratch/ allows to create directories. If it does, than a single bb_clientd mount is feasible. If it doesn't, then we must have a different bb_clientd mount per invocation id.

minor-fixes: @aaahrens says you can `mkdir` in the scratch directory, so we get by with one bb_clientd instance per `output_root` directory. `shutdown` will operate on one `output_root` by unmounting bb_clientd (which removes the scratch dir and thus all the hardlinks) and deleting the convenience symlinks within

enkit/outputs/commands.go
minor-fixes: Agreed; I'll replace this with something less specific.

enkit/outputs/commands.go
ccontavalli: sgtm, then! and I assume the mounting is done dynamically, no need for shutdown.

ccontavalli (approved):

PR #425, opened 2022-01-30
proxy/amux: handle domains with trailing ".", add conformance tests.

ccontavalli:
Background:
DNS allows domain names ending with a trailing ".", eg, "www.enfabrica.net."
or "www.google.com.". Adding a trailing dot is considered best practice
when specifying domain names in cli tools, bots, or automated systems
in general, as - without a trailing . - the name is considered "relative"
by DNS resolver library, and will be looked up multiple times in DNS
using the "search" domains in resolv.conf, increasing the load on DNS,
and possibly leading to very unexpected results.

An http server should generally treat "enfabrica.net" and "enfabrica.net." as
exactly the same domain. Golang muxes (and most web servers) don't do this by
default, introducing various kind of problems (see a good discussion here:
https://news.ycombinator.com/item?id=5384864, using a trailing period on
some sites allows to even bypass authentication or subscriptons).

In enkit, however, enkit/lib/khttp had code to deal with it by default,
and transparently.

Problem:
When moved away from the enkit/lib/khttp library for domain match,
the automated handling of domains with trailing "." was accidentally
dropped, meaning that tools trying to access "enfabrica.net." rather
than "enfabrica.net" started getting an error.

In this PR:
- update the documentation of amux, further specifying the expected contract.
- update the amux wrapper for muxie to transparently handle trailing ".",
  just like the previous mux used by enproxy.
- while testing the code, I realised that enproxy also requires at least
  one form of pattern matching, which has been added to the documentation.

Tested:
- this PR adds a conformance test, covering a lot of the edge cases
  in how the mux behaves, including trailing ".".

### DIFF
--- a/proxy/amux/mux_test.go
+++ b/proxy/amux/mux_test.go
@@ -1,0 +1,124 @@
+package amux_test
+
+import (
+	"testing"
+	  "github.com/stretchr/testify/assert"
+	"github.com/enfabrica/enkit/proxy/amux"
+	"github.com/enfabrica/enkit/proxy/amux/amuxie"
+	"net/http"
+	"net/http/httptest"
+	"fmt"
+)
+
+func CountHandler(counter *int) http.Handler {
+return http.HandlerFunc(func (w http.ResponseWriter, r *http.Request) {
+	*counter += 1
+})
+}
+
+var NilHandler = http.HandlerFunc(
+func (w http.ResponseWriter, r *http.Request) {
+},
+)
+
+func Request(m amux.Mux, host, path string) int {
+	r := httptest.NewRequest("GET", fmt.Sprintf("http://%s%s", host, path), nil)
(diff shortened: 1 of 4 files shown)


proxy/amux/mux_test.go
```diff
@@ -0,0 +1,124 @@
+package amux_test
+
+import (
+	"testing"
+	  "github.com/stretchr/testify/assert"
+	"github.com/enfabrica/enkit/proxy/amux"
+	"github.com/enfabrica/enkit/proxy/amux/amuxie"
```
minor-fixes, line 7: nit: gofmt, and separate stdlib imports from third-party

minor-fixes (approved):

proxy/amux/mux_test.go
ccontavalli: done.

PR #426, opened 2022-01-31
kbuildbarn: start formatting symlink urls

aaahrens:
What this does:

Generates out a generic LL of symlinks out of a NamedSetOfFiles protobuff message. Each node of the LL contains a source and destination symlink, formatted for bb_clientd cas and scratch directory.

### DIFF
--- a/lib/kbuildbarn/options.go
+++ b/lib/kbuildbarn/options.go
@@ -1,7 +1,5 @@
 package kbuildbarn
 
-import "fmt"
-
 type options struct {
 	Scheme       string
 	PathTemplate string
@@ -69,13 +67,36 @@ func WithDirectoryUrlTemplate() Option {
 	return pathTemplateOption(DefaultDirectoryTemplate)
 }
 
-type templateArgsOption []interface{}
+type templateArgsOption [][]interface{}
 
 func (ta templateArgsOption) apply(opts *options) {
+	opts.TemplateArgs = ta[0]
+}
+
+func WithTemplateArgs(args []interface{}) Option {
+	return templateArgsOption{args}
+}
+
+type templateAppendArgsOption []interface{}
+
+func (ta templateAppendArgsOption) apply(opts *options) {
 	opts.TemplateArgs = append(opts.TemplateArgs, ta...)
-	fmt.Println(opts.TemplateArgs)
+}
+
+func WithAdditionalTemplateArgs(args ...interface{}) Option {
+	return templateAppendArgsOption{args}
 }
 
 func WithFileName(s string) Option {
-	return multipleOption{pathTemplateOption(DefaultFileTemplate), templateArgsOption{s}}
+	return multipleOption{pathTemplateOption(DefaultFileTemplate), templateAppendArgsOption{s}}
+}
+
+type customTemplate string
+
+func (ct customTemplate) apply(opts *options) {
+	opts.PathTemplate = string(ct)
+}
+
+func WithFileTemplate(s string) Option {
+	return customTemplate(s)
 }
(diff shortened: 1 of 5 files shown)


lib/kbuildbarn/protoparse.go
```diff
@@ -0,0 +1,68 @@
...
+	DefaultBBClientdCasFileTemplate     = "cas/%s/blobs/file/%s"
+	DefaultBBClientdScratchFileTemplate = "scratch/%s/%s"
+)
+
+// BBClientDLink represents a single symlink of a file in the cas to a hard symlink in the scratch directory
+type BBClientDLink struct {
+	Src  string
+	Dest string
+	Next *BBClientDLink
+}
```
minor-fixes, line 19: Why a linked list over a slice?

lib/kbuildbarn/protoparse_test.go
```diff
@@ -0,0 +1,60 @@
...
+		},
+		{
+			Name: "tarball.tar", Digest: "digest3", Length: 777,
+		},
+	}}
+	baseName := "/foo/bar"
+	clusterName := "duster"
+	invocationPrefix := "invocation"
+	expected := map[string]string{
+		"/foo/bar/scratch/invocation/simple.txt":       "/foo/bar/cas/duster/blobs/file/digest0",
```
minor-fixes, line 45: Any advantage to generating relative symlinks?

lib/kbuildbarn/protoparse.go
aaahrens: Change to slice

lib/kbuildbarn/protoparse_test.go
aaahrens: Not really, we have 0 clarity as to where the commands to create symlinks gets run etc. absolute paths are much safer, especially for cleanup.

ccontavalli:
If we use the scratch space under buildbarn, btw, we should use hard links rather than symlinks. They are entirely invisible to the application, and what scratch/ was created for.

lib/kbuildbarn/options.go
```diff
@@ -69,13 +67,36 @@ func WithDirectoryUrlTemplate() Option {
...
 
-type templateArgsOption []interface{}
+type templateArgsOption [][]interface{}
 
 func (ta templateArgsOption) apply(opts *options) {
+	opts.TemplateArgs = ta[0]
+}
+
+func WithTemplateArgs(args []interface{}) Option {
+	return templateArgsOption{args}
```
minor-fixes, line 77: Maybe one slice dimension can be dropped if this is `return templateArgsOption(args)`?

...and then just `opts.TemplateArgs = ta` above?

aaahrens: So fun fact, if you wrap []interface{} with a custom type,  it is now [][]interface{} (functionally).

so if on line 76 you `fmt.Println(args)` => `[ foo bar ]`

and on line 96, you fmt.Println(ta) => `[ [ foo bar ] ] `


*but* you can declare TA as an []interface{} (where the interface{} is another []interface{}. This requires a type check.

*or* you can declare ta as a [][]interface{}, and the go compiler knows that you can unwrap.

lib/kbuildbarn/options.go
```diff
@@ -69,13 +67,36 @@ func WithDirectoryUrlTemplate() Option {
 	return pathTemplateOption(DefaultDirectoryTemplate)
 }
 
-type templateArgsOption []interface{}
+type templateArgsOption [][]interface{}
```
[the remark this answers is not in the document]
aaahrens: See below

lib/kbuildbarn/options.go
aaahrens: Changed to just convert the []string input to []interface on the flag itself, easy fix.

minor-fixes (approved):

PR #429, opened 2022-02-01
bestie: Add test metrics upload to BigQuery (#423)

gpaussaenf:
This continues the work of #394 and #406 by adding support
for uploading test metrics to BigQuery. The changes in this
PR do the following:

- Read and process metrics contained in *.metrics.pb file(s)
produced during bazel test runs. Format and push these
test metrics to the BigQuery 'bestie-builds.<dataset>.testmetrics'
table, where <dataset> is specified by the --dataset command line
parameter.

- Supports both file:// and bytestream:// schemes for the
outputs.zip file in the TestResult event, which correspond to
local and cluster operation of the BES Endpoint, respectively,
using APIs provided by enkit/lib/kbuildbarn. A new --baseurl
command line parameter allows a deployment-specific base URL
to be specified.

- Adds several service metrics related to BigQuery interactions
for monitoring the BES Endpoint operation.

Tested: Ran BES Endpoint service both locally and in the staging
cluster, where it detected the test.metrics.pb file produced by
a standalone test app. Authentication issues prevented the staging
cluster service from communicating with BigQuery, but that will be
resolved in a separate PR by creating a new service account.

Jira: INFRA-423

### DIFF
--- a/bestie/server/bigquery_metrics.go
+++ b/bestie/server/bigquery_metrics.go
@@ -4,39 +4,28 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
 	"strings"
 	"time"
 
-	bq "cloud.google.com/go/bigquery"
+	"cloud.google.com/go/bigquery"
 )
 
-type bigQuerySession struct {
-	ctx    context.Context
-	client *bq.Client
-}
-
 type bigQueryMetric struct {
-	metricname string
+	metricName string
 	tags       string // stringified JSON
 	value      float64
-	timestamp  string // "YYYY-MM-DD hh:mm:ss.uuuuuu"
+	timestamp  string // must be: "YYYY-MM-DD hh:mm:ss.uuuuuu"
 }
 
 type bigQueryTable struct {
 	project   string
 	dataset   string
-	tablename string
-}
-
-type keyValuePair struct {
-	key   string
-	value string
+	tableName string
 }
 
 type testMetric struct {
-	metricname string
-	tags       []keyValuePair
+	metricName string
+	tags       map[string]string
 	value      float64
 	timestamp  int64
 }
@@ -46,59 +35,93 @@ type metricTestResult struct {
 	metrics []testMetric
 }
 
-// BigQuery service location.
-const databaseLocation = "us-west1"
+// Timestamp datetime format required by BigQuery.
+// Go dictates that these particular field values must be used
+// when formatting datetime as a custom string.
+const timestampFormat = "2006-01-02 15:04:05.000000"
+
+// BigQuery test metrics table schema definition.
+// Although no longer creating tables in the code, this is kept for reference.
+//
+//var databaseSchema = bigquery.Schema{
+//	{Name: "metricname", Type: bigquery.StringFieldType, Description: "metric name", Required: true},
+//	{Name: "tags", Type: bigquery.StringFieldType, Description: "metric attribute tags"},
+//	{Name: "value", Type: bigquery.FloatFieldType, Description: "metric value"},
+//	{Name: "timestamp", Type: bigquery.TimestampFieldType, Description: "sample collection timestamp", Required: true},
+//}
 
 // Define default BigQuery table to use if not specified in Bazel TestResult event message.
 //
 // NOTE: If BES Endpoint needs to shard metrics into multiple tables, it will add a suitable
-// suffix to the tablename specified here.
+// suffix to the table name specified here.
 var bigQueryTableDefault = bigQueryTable{
 	project:   "bestie-builds",
 	dataset:   "",            // Must be specified as --dataset arg on the command line.
-	tablename: "testmetrics", // Can be overridden from the --tablename arg on the command line.
+	tableName: "testmetrics", // Can be overridden from the --table_name arg on the command line.
 }
 
 // Save implements the ValueSaver interface.
 // This example disables best-effort de-duplication, which allows for higher throughput.
-func (i *bigQueryMetric) Save() (map[string]bq.Value, string, error) {
-	ret := map[string]bq.Value{
-		"metricname": i.metricname,
+func (i *bigQueryMetric) Save() (map[string]bigquery.Value, string, error) {
+	ret := map[string]bigquery.Value{
+		"metricname": i.metricName,
 		"tags":       i.tags,
 		"value":      i.value,
 		"timestamp":  i.timestamp,
 	}
-	return ret, bq.NoDedupeID, nil
+	return ret, bigquery.NoDedupeID, nil
 }
 
-func getMetricsTableSchema() bq.Schema {
-	databaseSchema := bq.Schema{
-		{Name: "metricname", Type: bq.StringFieldType, Description: "metric name", Required: true},
-		{Name: "tags", Type: bq.StringFieldType, Description: "metric attribute tags"},
-		{Name: "value", Type: bq.FloatFieldType, Description: "metric value"},
-		{Name: "timestamp", Type: bq.TimestampFieldType, Description: "sample collection timestamp", Required: true},
+// Keep track of dataset and table names that were requested,
+// but do not exist. Display a log message the first time
+// each is encountered and every "rate" occurrence thereafter.
+var missingResource = make(map[string]int)
+var missingResourceMsgRate = 10
+
+// Report a missing dataset or table.
+func reportMissingResource(id, resourceType string) error {
+	if _, ok := missingResource[id]; !ok {
+		missingResource[id] = 0
 	}
-	return databaseSchema
+	missingResource[id]++
+	errMsg := fmt.Sprintf("The BigQuery %s %q does not exist", resourceType, id)
+	if (missingResource[id] % missingResourceMsgRate) == 1 {
+		logger.Printf("%s. Please create it and try again.", errMsg)
+	}
+	return fmt.Errorf("%s", errMsg)
+}
+
+// UnixMicro returns the local Time corresponding to the given Unix time,
+// usec microseconds since January 1, 1970 UTC.
+// NOTE: This function is copied from src/time/time.go in the Go library (version 1.17.6).
+func UnixMicro(usec int64) time.Time {
+	return time.Unix(usec/1e6, (usec%1e6)*1e3)
 }
 
 // Produce a formatted string of the table identifier.
-func (t bigQueryTable) formatDatasetId() string {
-	return fmt.Sprintf("`%s.%s`", t.project, t.dataset)
+func (t *bigQueryTable) formatDatasetId() string {
+	return fmt.Sprintf("%s.%s", t.project, t.dataset)
 }
 
 // Produce a formatted string of the dataset identifier.
-func (t bigQueryTable) formatTableId() string {
-	return fmt.Sprintf("`%s.%s.%s`", t.project, t.dataset, t.tablename)
+func (t *bigQueryTable) formatTableId() string {
+	return fmt.Sprintf("%s.%s.%s", t.project, t.dataset, t.tableName)
+}
+
+// Check if a BigQuery dataset exists based on reading its metadata.
+func (t *bigQueryTable) isDatasetExist(ctx context.Context, client *bigquery.Client) bool {
+	_, err := client.Dataset(t.dataset).Metadata(ctx)
+	return err == nil
 }
 
 // Check if a BigQuery table exists based on reading its metadata.
-func (t bigQueryTable) isTableExist(session bigQuerySession) bool {
-	_, err := session.client.Dataset(t.dataset).Table(t.tablename).Metadata(session.ctx)
+func (t *bigQueryTable) isTableExist(ctx context.Context, client *bigquery.Client) bool {
+	_, err := client.Dataset(t.dataset).Table(t.tableName).Metadata(ctx)
 	return err == nil
 }
 
 // Normalize BigQuery table references by filling in with defaults, as needed.
-func normalizeTableRef(t *bigQueryTable) {
+func (t *bigQueryTable) normalizeTableRef() {
 	// Always use the default project.
 	t.project = bigQueryTableDefault.project
 
@@ -107,68 +130,18 @@ func normalizeTableRef(t *bigQueryTable) {
 	if len(t.dataset) == 0 {
 		t.dataset = bigQueryTableDefault.dataset
 	}
-	if len(t.tablename) == 0 {
-		t.tablename = bigQueryTableDefault.tablename
+	if len(t.tableName) == 0 {
+		t.tableName = bigQueryTableDefault.tableName
 	}
-}
-
-// Create a new dataset using an explicit destination location.
-func createDataset(w io.Writer, t bigQueryTable, location string, session bigQuerySession) error {
-	meta := &bq.DatasetMetadata{
-		Location: location, // See https://cloud.google.com/bigquery/docs/locations
-	}
-	if err := session.client.Dataset(t.dataset).Create(session.ctx, meta); err != nil {
-		return fmt.Errorf("Error creating dataset '%s': %s\n", t.formatDatasetId(), err)
-	}
-	fmt.Fprintf(w, "Created dataset %s\n", t.formatDatasetId())
-	return nil
-}
-
-// Create a BigQuery table with a predefined schema.
-func createTable(w io.Writer, t bigQueryTable, session bigQuerySession) error {
-	// Always attempt to create the dataset that holds the table.
-	// This is faster than first querying the table metadata,
-	// then creating the dataset if not present.
-	// We expect an "already exists" error if the dataset currently
-	// exists for the project, which is the normal case.
-	if err := createDataset(w, t, databaseLocation, session); err != nil {
-		if !strings.Contains(strings.ToLower(err.Error()), "already exist") {
-			return err
-		}
-	}
-
-	// Create the table within the dataset.
-	tableSchema := getMetricsTableSchema()
-	fmt.Fprintln(w, "Metrics Table Schema:")
-	for _, field := range tableSchema {
-		fmt.Fprintf(w, "  %s (%s)\n", field.Name, field.Type)
-	}
-	fmt.Fprintln(w)
-
-	// NOTE: Not setting an expiration for BigQuery tables created by the endpoint.
-	// If an expiration was used and someone forgets to extend the deadline,
-	// the table and all of its data are deleted by BigQuery without warning.
-	metaData := &bq.TableMetadata{
-		Schema: tableSchema,
-		//ExpirationTime: time.Now().AddDate(1, 0, 0), // Table will be automatically deleted in 1 year
-	}
-	tableRef := session.client.Dataset(t.dataset).Table(t.tablename)
-	if err := tableRef.Create(session.ctx, metaData); err != nil {
-		if !strings.Contains(strings.ToLower(err.Error()), "already exist") {
-			return fmt.Errorf("Error creating table '%s': %s\n", t.formatTableId(), err)
-		}
-	} else {
-		fmt.Fprintf(w, "Created table %s\n", t.formatTableId())
-	}
-	return nil
 }
 
 // Translate protobuf metric to BigQuery metric.
-func translateMetric(stream bazelStream, m testMetric) (bigQueryMetric, error) {
-	// Process the slice of key/value pairs representing individual metric tags.
+// TODO: Decide whether to override client metric time with BES server current time.
+func translateMetric(stream *bazelStream, m *testMetric) (*bigQueryMetric, error) {
+	// Process the map of key/value pairs representing individual metric tags.
 	dat := make(map[string]string)
-	for _, kv := range m.tags {
-		dat[kv.key] = kv.value
+	for k, v := range m.tags {
+		dat[k] = v
 	}
 	// Insert additional tags using information that is only available
 	// to the BES endpoint through the Bazel event message itself.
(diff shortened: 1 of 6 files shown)


bestie/server/bigquery_metrics.go
```diff
@@ -0,0 +1,288 @@
...
+// Upload this set of metrics to the specified BigQuery table.
+func uploadTestMetrics(w io.Writer, stream bazelStream, r *metricTestResult) error {
+	// Normalize the BigQuery table identifier based on whether one was specified
+	// in the protobuf message.
+	normalizeTableRef(&r.table)
+	fmt.Fprintf(w, "Normalized table ref: %s\n", r.table.formatTableId())
+
+	// Get client context for this BigQuery operation.
+	ctx := context.Background()
+	client, err := bq.NewClient(ctx, r.table.project)
```
gpaussaenf, line 212: This is currently failing in the staging cluster due to lack of auth credentials for accessing BigQuery. Adam is looking into a fix; will revisit this call once the change is determined.

bestie/server/bigquery_metrics.go
```diff
@@ -0,0 +1,288 @@
...
+
+// Create a BigQuery table with a predefined schema.
+func createTable(w io.Writer, t bigQueryTable, session bigQuerySession) error {
+	// Always attempt to create the dataset that holds the table.
+	// This is faster than first querying the table metadata,
+	// then creating the dataset if not present.
+	// We expect an "already exists" error if the dataset currently
+	// exists for the project, which is the normal case.
+	if err := createDataset(w, t, databaseLocation, session); err != nil {
+		if !strings.Contains(strings.ToLower(err.Error()), "already exist") {
```
minor-fixes, line 135: This should probably be something like:

```
var apiErr *googleapi.Error
if errors.As(err, &apiErr)  && apiErr.Code == /* whatever the HTTP code for already exists is */ {
  // ignore
} else {
  // some other error worth handling; add context and return
}
```

bestie/server/bigquery_metrics.go
```diff
@@ -0,0 +1,288 @@
...
+	timestamp  string // "YYYY-MM-DD hh:mm:ss.uuuuuu"
+}
+
+type bigQueryTable struct {
+	project   string
+	dataset   string
+	tablename string
+}
+
+type keyValuePair struct {
```
aaahrens, line 32: These should probably look something like this

`type QueryTags map[string][]string `

Just a suggestion, but it makes it more readable and idiotmatic.

bestie/server/service.go
```diff
@@ -97,3 +139,38 @@ func (s *serviceStats) incrementEventsTotal(bevid interface{}) {
...
+	s.bigQueryInsertTotal.WithLabelValues("timeout").Inc()
+}
+
+// BigQuery total metrics processed.
+func (s *serviceStats) incrementBigQueryMetricDiscard(count int) {
+	s.bigQueryMetricTotal.WithLabelValues("discard").Add(float64(count))
+}
+func (s *serviceStats) incrementBigQueryMetricUpload(count int) {
+	s.bigQueryMetricTotal.WithLabelValues("upload").Add(float64(count))
+}
```
minor-fixes, line 176: Can all this be inlined?

bestie/server/bigquery_metrics.go
```diff
@@ -0,0 +1,288 @@
...
+	if len(t.dataset) == 0 {
+		t.dataset = bigQueryTableDefault.dataset
+	}
+	if len(t.tablename) == 0 {
+		t.tablename = bigQueryTableDefault.tablename
+	}
+}
+
+// Create a new dataset using an explicit destination location.
+func createDataset(w io.Writer, t bigQueryTable, location string, session bigQuerySession) error {
```
aaahrens, line 116: Usually something like this, e.g. S3 bucket should always exist, should be a Must type function.

Instead of returning the error, how about renaming ti something like MustCreateDataset() and if the error returned is not "already exists", panic.

Because the program can't run unless the bucket's been made, a panic is good.  Also it should be called as soon as possible in the program.

bestie/server/bigquery_metrics.go
aaahrens: +1

bestie/server/bigquery_metrics.go
```diff
@@ -0,0 +1,288 @@
...
+		return bigQueryMetric{}, fmt.Errorf("Error converting JSON to string: %s", err)
+	}
+
+	// Change the metric timestamp from integer to formatted string,
+	// since that is what BigQuery expects for a TIMESTAMP column.
+	//
+	// The incoming timestamp is UTC nanoseconds since epoch, but
+	// storing with microsecond granularity in BigQuery.
+	ts := m.timestamp / 1000
+	tsf := time.Unix(ts/1000000, ts%1000000).Format("2006-01-02 15:04:05.000000")
```
minor-fixes, line 193: What if `i.timestamp` is a raw `*time.Time`? Does the client lib do the right thing? (avoids this translation code)

bestie/server/bigquery_metrics.go
```diff
@@ -0,0 +1,288 @@
...
+
+	// Attempt to upload the metrics, assuming the dataset and table
+	// both exist. If a "not found" error occurs, sleep for a while
+	// then try uploading again.
+	//
+	// Using a finite delay loop here to give BigQuery time to instantiate
+	// the table, which can take a relativly long time (i.e. tens of seconds).
+	ok := false
+	insertionDelay := 0
+	sleepTime := 10
```
minor-fixes, line 259: I don't think having lots of code to try-create tables and datasets is worth it, especially if it includes arbitrary sleeps. Can we make the simplifying assumption that the tables and datasets will exist before we start bestie and point it at them, or else bestie is allowed to error? This simplifies the code greatly, and I think is fine since we're going to make a constant number of tables/datasets (new tables is not dynamic)

bestie/server/bigquery_metrics.go
```diff
@@ -0,0 +1,288 @@
...
+			fmt.Fprintf(w, ".")
+		} else {
+			ok = true
+			break
+		}
+		time.Sleep(time.Duration(sleepTime) * time.Second)
+		insertionDelay += sleepTime
+	}
+	fmt.Fprintln(w)
+	if !ok {
```
aaahrens, line 279: It is probably better to start a goroutine for incrementing the prometheus stats. Just so that it's non blocking.

aaahrens: so just call for example `go ServiceStats.incrementBigQueryMetricUpload(len(rows))`

bestie/server/testresult.go
```diff
@@ -168,6 +205,7 @@ func extractZippedFiles(stream bazelStream, zipFile string) error {
 func getTestMetricsFromFileData(pbmsg []byte) (*metricTestResult, error) {
 	tmet := &tpb.TestMetrics{}
 	if err := proto.Unmarshal(pbmsg, tmet); err != nil {
+		ServiceStats.incrementBigQueryProtobufError()
```
aaahrens, line 208: Same as above `go <metricBlah>`

aaahrens (changes requested):

bestie/server/bigquery_metrics.go
```diff
@@ -0,0 +1,288 @@
+package main
```
minor-fixes, line 1: This file can be moved to it's own library package (how about `package metrics` in `lib/metrics`?)

This would allow for:
* shortening of identifiers: `bigQueryMetric` -> `metric.Metric`, `bigQueryTable` -> `metric.Table`, etc (bigQuery is dropped because it's an implementation detail)
* addition of unit tests for these functions

minor-fixes (commented):
A lot of comments; as always, don't hesitate to reach out over chat or drop something on my calendar if you'd like to go over the code in-person, or would like tips/clarification

bestie/server/bigquery_metrics.go
minor-fixes: I agree with the sentiment (have bestie require the dataset/tables to already exist, and error if they don't) - disagree with the implementation - don't panic in production code; instead, return the error to the top-level and have it exit gracefully. Since the function doesn't panic, don't prefix the name with `Must`.

We could either:
* check that the tables exist at startup, and error early if they don't
* error on every request if the table doesn't exist

...this depends partially on what's more convenient from the bigquery library, with the former being a little easier to deal with operationally.

gpaussaenf: This negates the ability for a test client application to define its own dataset and/or tablename in the *.metrics.pb file as an indication of where it wants to store its metrics in BigQuery. My main concern is that I don't know how other parts of the org plan on storing their metrics, so I wanted to allow for an optional means of specifying their destination. Am I being too flexible with this?

bestie/server/bigquery_metrics.go
gpaussaenf: OK, will do.

bestie/server/testresult.go
gpaussaenf: OK.

bestie/server/main.go
```diff
@@ -76,9 +80,44 @@ func (s *BuildEventService) PublishBuildToolEventStream(stream bpb.PublishBuildE
 	return nil
 }
 
+// Command line arguments.
+var (
+	argBaseUrl   = flag.String("baseurl", "", "Base URL for accessing output artifacts in the build cluster (required)")
```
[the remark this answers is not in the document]
gpaussaenf: OK

bestie/server/bigquery_metrics.go
```diff
@@ -0,0 +1,288 @@
...
+	// Note: bigQueryMetric implements the ValueSaver interface.
+	var bqMetrics []bigQueryMetric
+	var rows []*bigQueryMetric
+	idx := 0
+	for _, m := range r.metrics {
+		// TODO: Decide whether to override client metric time with BES server current time.
+		bqMetric, err := translateMetric(stream, m)
+		if err != nil {
+			ServiceStats.incrementBigQueryMetricDiscard(1)
+			fmt.Fprintf(w, "Discarding metric %s due to error: %s\n", m, err)
```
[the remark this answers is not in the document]
gpaussaenf: Addressing this in the following ways:
1. Created a logger to replace the fmt.Printf's, etc.
2. Defined debugPrintf() and debugPrintln() functions that call the logger if the isDebugMode global variable is set to true, which comes from a new `--debug` command line arg. Decided which are normal vs. debug log messages.
3. Used strings.Builder to create a single string to be logged / debug logged all at once, rather than logging it piecemeal.

bestie/server/bigquery_metrics.go
gpaussaenf: Update: Discussed this with Mike G. We like having the current options to specify a different dataset and/or tablename, but agree that it's reasonable to expect those to be pre-created manually to simplify the BES Endpoint in that regard. I'll probably do something like track non-existing "project.dataset.tablename" requests in a map, so that we can log the first occurrence and just bump a service counter on subsequent occurrences.

bestie/server/bigquery_metrics.go
```diff
@@ -0,0 +1,288 @@
...
+	"fmt"
+	"io"
+	"strings"
+	"time"
+
+	bq "cloud.google.com/go/bigquery"
+)
+
+type bigQuerySession struct {
+	ctx    context.Context
```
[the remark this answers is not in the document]
gpaussaenf: OK

bestie/server/bigquery_metrics.go
```diff
@@ -0,0 +1,288 @@
...
+	timestamp  int64
+}
+
+type metricTestResult struct {
+	table   bigQueryTable
+	metrics []testMetric
+}
+
+// BigQuery service location.
+const databaseLocation = "us-west1"
```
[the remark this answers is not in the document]
gpaussaenf: True, but this is no longer needed now that we're no longer creating datasets from within the BES Endpoint.

bestie/server/testresult.go
```diff
@@ -133,20 +167,19 @@ func extractZippedFiles(stream bazelStream, zipFile string) error {
...
 			continue
 		}
-		fmt.Printf("Found output file to process: %s\n", file.Name)
 
+		// TODO (PR-394): May need to chunk the .pb file into multiple messages, or enforce a max metrics file size.
+		// Read entire file contents into a byte slice.
 		f, err := file.Open()
 		if err != nil {
-			return err
+			continue
```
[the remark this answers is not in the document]
gpaussaenf: It's an error, but it occurs inside a for loop where other .metrics.pb files may still be processed. I'll add an error message to a multierror slice and allowing the loop to continue.

bestie/server/bigquery_metrics.go
```diff
@@ -0,0 +1,288 @@
...
+	metricname string
+	tags       string // stringified JSON
+	value      float64
+	timestamp  string // "YYYY-MM-DD hh:mm:ss.uuuuuu"
+}
+
+type bigQueryTable struct {
+	project   string
+	dataset   string
+	tablename string
```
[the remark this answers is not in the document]
gpaussaenf: OK

bestie/server/bigquery_metrics.go
```diff
@@ -0,0 +1,288 @@
...
+	ret := map[string]bq.Value{
+		"metricname": i.metricname,
+		"tags":       i.tags,
+		"value":      i.value,
+		"timestamp":  i.timestamp,
+	}
+	return ret, bq.NoDedupeID, nil
+}
+
+func getMetricsTableSchema() bq.Schema {
```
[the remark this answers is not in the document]
gpaussaenf: Yes, though it's no longer needed since not creating database tables in the code.

bestie/server/bigquery_metrics.go
gpaussaenf: I verified that this approach works, checking for http.StatusConflict (409 already exists). However, this is only needed when creating a dataset or table in BigQuery, which the endpoint no longer does, so the entire function containing this logic is being removed.

bestie/server/testresult.go
```diff
@@ -71,6 +55,40 @@ func identifyStream(bazelBuildEvent bes.BuildEvent, streamId *build.StreamId) ba
 	return stream
 }
 
+// Read the contents of a bytestream file.
+func readBytestreamFile(fileName, bytestreamUri string) ([]byte, error) {
```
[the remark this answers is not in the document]
gpaussaenf: Not sure if this is feasible, but will take some investigation. I don't have time to pursue this now, so I've opened [INFRA-504](https://enfabrica.atlassian.net/browse/INFRA-504) to track it. Feel free to reassign, as appropriate.

bestie/server/testresult.go
```diff
@@ -85,46 +103,62 @@ func handleTestResultEvent(bazelBuildEvent bes.BuildEvent, streamId *build.Strea
...
+		// The outputs.zip file was not found.
 		return nil
 	}
 
-	// Strip off any file:// prefix from the URI to access the local file system path.
-	// TODO (PR-394): Handle similar adjustment for Cloud Run URI.
-	filePrefix := "file://"
-	if strings.HasPrefix(ofuri, filePrefix) {
-		ofuri = ofuri[len(filePrefix):]
+	var fileBytes []byte = []byte{}
```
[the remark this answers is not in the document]
gpaussaenf: Fixed. Went with the first style.

bestie/server/testresult.go
```diff
@@ -85,46 +103,62 @@ func handleTestResultEvent(bazelBuildEvent bes.BuildEvent, streamId *build.Strea
...
+	var readErr error = nil
+	urlParts := strings.Split(outFileUri, "://")
+	if len(urlParts) == 2 {
```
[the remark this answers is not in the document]
gpaussaenf: OK

bestie/server/testresult.go
```diff
@@ -85,46 +103,62 @@ func handleTestResultEvent(bazelBuildEvent bes.BuildEvent, streamId *build.Strea
...
+		switch scheme {
+		case "bytestream":
+			// Handle cluster build scenario, translating bytestream:// URL to file URL for http.Get().
+			fileBytes, readErr = readBytestreamFile(outFileName, outFileUri)
+		case "file":
+			// Use URI without file:// prefix to access the local file system path.
+			fileBytes, readErr = ioutil.ReadFile(fileRef)
+		default:
+			// Silently ignore file: not a supported URL scheme prefix.
+			return nil
```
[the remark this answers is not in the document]
gpaussaenf: Adding a log message.

bestie/server/testresult.go
```diff
@@ -71,6 +55,40 @@ func identifyStream(bazelBuildEvent bes.BuildEvent, streamId *build.StreamId) ba
...
+
+	// Read response body regardless of response status
+	// so it can be used in the HTTP error message below.
+	respBody, readErr := ioutil.ReadAll(resp.Body)
+
+	if respStatus != http.StatusOK {
+		return nil, fmt.Errorf("HTTP error status %d: response: %s", respStatus, respBody)
+	}
+	if readErr != nil {
+		return nil, fmt.Errorf("%s file read error: %s", fileName, readErr)
```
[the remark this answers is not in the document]
gpaussaenf: OK

bestie/server/main.go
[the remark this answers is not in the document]
gpaussaenf: OK. BTW, I looked up dataset and it is acceptable to be used as a single word.

bestie/server/bigquery_metrics.go
```diff
@@ -0,0 +1,288 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"strings"
+	"time"
+
+	bq "cloud.google.com/go/bigquery"
```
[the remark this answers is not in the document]
gpaussaenf: OK

bestie/server/bigquery_metrics.go
gpaussaenf: I agree that this would simplify the endpoint quite a bit. I do want to retain the ability for overriding the default BigQuery dataset and/or tablename that is used, but with the caveat that these must be manually created in BigQuery beforehand.

bestie/server/bigquery_metrics.go
gpaussaenf: I looked into this for quite some time and concluded that we still need to do the translation. I cleaned up the code a little, copying a small time.UnixMicro() function from a more recent version of Go, but ultimately the .Format() is required to conform the TIMESTAMP string to what BigQuery requires. For example, time.UnixMicro().String() creates a nicely formatted timestamp, but it ends with "+0000 UTC", which is enough to prevent BigQuery from adding the metrics to the table. Hence, producing a custom format acceptable to BigQuery requires a .Format().

bestie/server/testresult.go
```diff
@@ -85,46 +103,62 @@ func handleTestResultEvent(bazelBuildEvent bes.BuildEvent, streamId *build.Strea
...
+	urlParts := strings.Split(outFileUri, "://")
+	if len(urlParts) == 2 {
+		scheme, fileRef := urlParts[0], urlParts[1]
+		switch scheme {
+		case "bytestream":
+			// Handle cluster build scenario, translating bytestream:// URL to file URL for http.Get().
+			fileBytes, readErr = readBytestreamFile(outFileName, outFileUri)
+		case "file":
+			// Use URI without file:// prefix to access the local file system path.
+			fileBytes, readErr = ioutil.ReadFile(fileRef)
```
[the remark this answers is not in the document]
gpaussaenf: See previous comment.

bestie/server/bigquery_metrics.go
```diff
@@ -0,0 +1,288 @@
...
+		{Name: "metricname", Type: bq.StringFieldType, Description: "metric name", Required: true},
+		{Name: "tags", Type: bq.StringFieldType, Description: "metric attribute tags"},
+		{Name: "value", Type: bq.FloatFieldType, Description: "metric value"},
+		{Name: "timestamp", Type: bq.TimestampFieldType, Description: "sample collection timestamp", Required: true},
+	}
+	return databaseSchema
+}
+
+// Produce a formatted string of the table identifier.
+func (t bigQueryTable) formatDatasetId() string {
```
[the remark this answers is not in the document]
gpaussaenf: OK

bestie/server/service.go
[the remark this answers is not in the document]
gpaussaenf: I wanted to abstract away the specific details of the Prometheus metric API from the rest of the code. I've now changed this completely and instead of having individual accessor methods, I defined a couple of general functions that are called throughout the code using a counter ID. These in turn invoke a common function as a goroutine (per Adam's comments above) to update the actual Prometheus counter.

bestie/server/bigquery_metrics.go
gpaussaenf: Agreed. I've eliminated the extra type and incorporated `tags map[string]string` directly as an element of the `testMetric` struct.

bestie/server/bigquery_metrics.go
gpaussaenf: Decided to defer this work for now. Not sure of the benefit of making this a separate library, since the code is "BigQuery aware" by definition. This change does not affect the operation of the BES Endpoint and can be considered later on, as time permits. Opened [INFRA-506](https://enfabrica.atlassian.net/browse/INFRA-506) to track as a future work item.

bestie/server/service.go
gpaussaenf: GitHub seems to have lost my first attempt at replying, so I'll try again...

The intent of these methods is to abstract away the details of Prometheus metrics from the rest of the code. I've completely changed my approach, replacing these individual methods with a few general purpose functions that increment the service counters based on counter IDs. These new functions conveniently wrap a call to a common update function that is invoked as a goroutine, thus addressing Adam's comment on this subject.

bestie/server/testresult.go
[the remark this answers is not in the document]
minor-fixes: Ok - this is going to be very important to follow up on - we are planning on sending builds that are going to generate giant output.zip files (GBs) and as written right now this will try and load it into RAM and bestie will probably fall over

minor-fixes (approved):

aaahrens (approved):

bestie/server/bigquery_metrics.go
gpaussaenf: This has since been fixed by Adam. Confirmed this latest BES Endpoint code is able to push metrics into BigQuery from the staging cluster.

PR #430, opened 2022-02-01
kbuildbarn: integration with buddy

aaahrens:
What this does:

- Implements GenerateSymlinks which takes in a build buddy client.
- Implements response side filtering so that all destination symlinks are unique.


lib/bes/buildbuddy.go
```diff
@@ -20,6 +20,8 @@ var (
 	getInvocationEndpoint = mustParseURL("rpc/BuildBuddyService/GetInvocation")
 )
 
+var _ httpDoer = http.DefaultClient
+
```
minor-fixes, line 24: Why is this needed?

lib/bes/buildbuddy.go
```diff
@@ -34,7 +36,7 @@ type BuildBuddyClient struct {
 // specified URL. If not nil, auth cookies are discovered via BaseFlags and
 // added to every request. apiKey must be a valid BuildBuddy API key (other
 // forms of auth are not currently supported).
-func NewBuildBuddyClient(u *url.URL, bf *client.BaseFlags, apiKey string) (*BuildBuddyClient, error) {
+func NewBuildBuddyClient(u *url.URL, bf *client.BaseFlags, apiKey string, client httpDoer) (*BuildBuddyClient, error) {
```
minor-fixes, line 39: I think this would be better as a separate function: `func NewBuildBuddyClientForTest(mockHttp httpDoer) *BuildBuddyClient` - simpler implementation - and the fact that the cookie doesn't get attached to a pre-created client doesn't become confusing if the `client != nil` codepath is used in production code

lib/kbuildbarn/buddy.go
```diff
@@ -0,0 +1,44 @@
...
+		nsof := event.GetNamedSetOfFiles()
+		if nsof != nil {
+			return GenerateLinksForFiles(nsof.Files, baseName, invocation, clusterName)
+		}
+		return nil
+	}
+}
+
+func GenerateSymlinks(client *bes.BuildBuddyClient, baseName, invocation, clusterName string, options ...FilterOption) (BBClientdList, error) {
+	ctx := context.Background()
```
minor-fixes, line 32: This context should be a param - only tests/main should grab the background context

lib/kbuildbarn/buddy.go
```diff
@@ -0,0 +1,44 @@
...
+	return func(event *bespb.BuildEvent, baseName, invocation, clusterName string) BBClientdList {
+		nsof := event.GetNamedSetOfFiles()
+		if nsof != nil {
+			return GenerateLinksForFiles(nsof.Files, baseName, invocation, clusterName)
+		}
+		return nil
+	}
+}
+
+func GenerateSymlinks(client *bes.BuildBuddyClient, baseName, invocation, clusterName string, options ...FilterOption) (BBClientdList, error) {
```
minor-fixes, line 31: I feel like `BBClientdList` isn't aging well already - it's a list of hardlinks, right? Not a list of bb_clientd's

aaahrens: Yeah, should be renamed. Changed to SymlinkList and Symlink

lib/bes/buildbuddy.go
aaahrens: Made a function NewTestClient() and reverted the normal factory change

lib/bes/buildbuddy.go
aaahrens: When duck typing a struct for testing usually it's polite to name where it came from.

lib/kbuildbarn/buddy.go
aaahrens: Done

lib/kbuildbarn/protoparse.go
```diff
@@ -10,17 +10,17 @@ const (
 	DefaultBBClientdScratchFileTemplate = "scratch/%s/%s"
 )
 
-type BBClientdList []*BBClientDLink
+type SymlinkList []*Symlink
```
minor-fixes, line 13: From the googling I'm doing, symbolic link is most commonly associated with soft link, and hard links are different. Can we name these Hardlink/HardlinkList (or even just Hardlinks)

aaahrens: Done

minor-fixes (approved):

PR #433, opened 2022-02-02
faketree: Add exec library

minor-fixes:
This change adds a sub-library of `faketree` that exec's out to a
`faketree` binary found on `$PATH`. `enkit outputs` subcommands will use
this function to manage a faketree process in order to re-root artifacts
in some cases.

When launching the default shell in faketree, `--norc` and `--noprofile`
are passed in order to allow the prompt to be modified in a helpful way;
otherwise, user modifications to `$PS1` in their `.bashrc` can override
any environment passed into the faketree invocation.

Tested:

* Added rudimentary unit tests that ensure that the return code of
  faketree'd processes is properly propagated
* Manual testing with a sample binary to inspect faketree behavior (not
  included in this PR)

Jira: INFRA-473


faketree/exec/BUILD.bazel
```diff
@@ -0,0 +1,15 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+go_library(
+    name = "go_default_library",
+    srcs = ["exec.go"],
+    importpath = "github.com/enfabrica/enkit/faketree/exec",
+    visibility = ["//visibility:public"],
+)
+
+go_test(
```
aaahrens, line 10: Probably because of the pre submit failures, mark this as manual?

aaahrens (approved):

PR #436, opened 2022-02-02
gee: use standard key location

jonathan-enf:
gee: use standard key location

Rather than creating an ssh key specifically for gee connecting to github,
it is a lower friction for our users to just use the standard keyfile that
ssh defaults to trying whenever it connects to a new host.  This enables
the key to "just work" even if not explicitly added using `ssh-add` or
configured via `~/.ssh/config`.

For backwards compatibility, gee will fall back on the old keyfile if it
exists and the new (standard) keyfile is missing.

I validated the key flow by going through the ssh key enrollment flow again,
and make some improvements to that flow:

1.  gee no longer runs `gh auth login` if `gh auth status` is okay.
2.  gee handles `gh ssh-key add` failure gracefully.

While at it, I added some gh API validation to the `gee repair` flow.

Tested:

* `gee hello` still works.
* `gee create_ssh_key` works better.
* `gee repair` confirms gh connectivity.

PR generated by jonathan from branch gee_use_standard_key.


minor-fixes (approved):

minor-fixes (approved):
From the presubmit:

```
In scripts/gee line 162:
readonly -a SSHKEYFILES=("${HOME}/.ssh/id_ed25519", "${HOME}/.ssh/gee_github_ed25519")
                        ^-- SC2054: Use spaces, not commas, to separate array elements.
```

PR #437, opened 2022-02-02
enfuse: allow to pull in remote config files from a private astore instance

aaahrens:
What this PR does:
- Sets an example of how to pull in from a private astore repository and parse in the config as command line arguments
- Pulls in spf13/viper since it's already done most of the heavy lifting

Problems it solves:
- We are unable to read in from a private astore repository and marshal into cobra flags.


astore/cobra.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (r *realFileFetcher) FileContent(ctx context.Context, path string) (string, error) {
+	f, err := r.c.Retrieve(ctx, &astorepb.RetrieveRequest{Path: path})
+	if err != nil {
+		return "", err
+	}
+	b := new(strings.Builder)
+	h := progress.NewDiscard()
+	if err := astore.Download(context.TODO(), progress.WriterCreator(h, knetwork.NopWriteCloser(b)), f.Url); err != nil {
```
minor-fixes, line 32: Pipe `ctx` through Download

astore/cobra.go
```diff
@@ -0,0 +1,60 @@
...
+var _ FileFetcher = (*realFileFetcher)(nil)
+
+type realFileFetcher struct {
+	c astorepb.AstoreClient
+}
+
+func (r *realFileFetcher) FileContent(ctx context.Context, path string) (string, error) {
+	f, err := r.c.Retrieve(ctx, &astorepb.RetrieveRequest{Path: path})
+	if err != nil {
+		return "", err
```
minor-fixes, line 28: Add context to errors where appropriate

enkit/outputs/configuration/README.md
```diff
@@ -0,0 +1,7 @@
+### Updating the configs to astore:
+> assuming you're using the default paths, running something like 
+```
+    enkit astore upload realconfig.yaml@enkitconfigs/outputs.yaml
+```
+
+will update the prefilled configs for the command
```
minor-fixes, line 7: How will someone know when this needs to be done?

Let's make updating the config an explicit list of steps, like:

```
1. Download the current config: `insert command here`
2. Change paramenters in the downloaded config
3. Upload modified config back to astore: `insert command here`
4. Set `live` tag in astore (or whatever else needs to be done - command for each)
```

I'm thinking that providing a blank template gives people space to fill in the wrong parameters because they know e.g. the new API key but in the process of creating a new config, mess up the cluster name

...and then make sure there's a link here in the relevant place (playbook for enkit?) or just inline this in that one place

enkit/outputs/commands.go
```diff
@@ -82,11 +94,14 @@ func NewMount(root *Root) *Mount {
...
 		},
-		root: root,
+		root:   root,
+		config: &MountConfig{},
 	}
-	command.Flags().StringVar(&command.BuildBuddyApiKey, "api-key", "", "build buddy api key used to bypass oauth2")
-	command.Flags().StringVar(&command.BuildBuddyUrl, "url", "", "build buddy url instance")
-	command.Flags().StringVar(&command.ClusterName, "cluster", "", "name of the cluster")
+
+	command.Flags().StringVar(&command.config.ApiKey, "api-key", "", "build buddy api key used to bypass oauth2")
```
minor-fixes, line 101: nit: making these flag names a bit more verbose makes commands more self-documenting:
* `api-key` -> `buildbuddy-api-key`
* `url` -> `buildbuddy-url`
* `cluster` -> `cluster-name`

astore/cobra.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	b := new(strings.Builder)
+	h := progress.NewDiscard()
+	if err := astore.Download(context.TODO(), progress.WriterCreator(h, knetwork.NopWriteCloser(b)), f.Url); err != nil {
+		return "", err
+	}
+	return b.String(), err
+}
+
+func ReadConfig(v *viper.Viper, bf *client.BaseFlags, sf *client.ServerFlags, fileName string, obj interface{}) error {
```
ccontavalli, line 38: Please use the populator library and kflags. Don't roll your own.

Why?
- there is no retry strategy here.
- there is no caching (eg, no papering over config push errors)
- if we change how we do auth, we have to come here and change it.
- no flags to change or override defaults...

Follow what the main here is doing:
https://github.com/enfabrica/enkit/blob/master/astore/client/main.go#L15

@minor-fixes

ccontavalli: Another major problem is that those configs are different based on the username@domain provided by the user.
If you hard code a path to download the file from, ... we're sort of back to square zero.

aaahrens: @ccontavalli can you show me where we can pull in **private** yaml file from astore and populate flags.  I dug through kflags and kcobra for a time and found nothing. I sent a DM on Feb 2nd about it also.

 Also going to address these in order:
 
 * retry is trivial, but can add.
 * we **don't** want local caching because these configs contain api keys which can refresh at a moments notice.  
 * I use all the native functionality from client.BaseFlags and ServerFlags, so any changes to auth will propogate. I don't roll my own at all.  
 * Viper unmarshall doesn't override if the property is not present. This is why I pulled it in. If we want smart merging it's trivial to add also. More than happy to do so.

=== Aside ===

kcobra has some serious overlap with viper, and has some seriously deep rooted bugs that make it not play nice. For example:

* If I registered *client.ServerFlags in this command like astore, the kcobra.populator never touches them. This is why I have to pass in the serverFlags to outputs. This is the case even if I register it with the **same** prefix and command. 
* Any command line parsing from cobra is clobbered flat out. cobra.ExactArgs(1) panics (or errors unexpectedly don't remember).  
* This is my opinion, but the api on commands be the same as it is on projects 50x the size of enkit. `
`
NewCommand().Execute()
` 
 expecting the command to be monkey patched by the flags is an anti pattern.

PR #438, opened 2022-02-02
bugfixes: update yarn lock

aaahrens:
Problems: 
- once of the dependencies was not on the right version in the yarn.lock, causing `bazelisk build //...` to fail. Regenerating yarn.lock fixes this. 
- changes the ordering of proto declaration in machinist and astore 

...works now with bazel 5.0


minor-fixes:
Cool, thanks for the fix!

(may need to raise the presubmit timeout in order to see it succeed - maybe 30m is enough?)

minor-fixes (approved):

PR #441, opened 2022-02-03
enkit outputs: Start tunnel when necessary

minor-fixes:
This change adds a few helper functions:
* A function to determine if a particular host requires a tunnel for
  communication
* A function to start and background a tunnel to a particular host
  listening on a port

These helper functions are used to allow `enkit outputs` subcommands to
determine:
* if a tunnel should be started to communicate with buildbarn
* if so, start a tunnel and replace the destination port with the local
port the tunnel is currently listening on

Tested: `enkit outputs mount` gets past tunnel initialization to a
BuildBuddy error (no invocation ID supplied)

Jira: INFRA-507

### DIFF
--- a/enkit/outputs/commands.go
+++ b/enkit/outputs/commands.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net"
 	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
 	"time"
 
 	"github.com/enfabrica/enkit/lib/bes"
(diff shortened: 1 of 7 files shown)


enkit/outputs/commands.go
```diff
@@ -80,16 +89,48 @@ func NewMount(root *Root) *Mount {
 		root: root,
 	}
 	command.Flags().StringVar(&command.BuildBuddyApiKey, "api-key", "", "build buddy api key used to bypass oauth2")
-	command.Flags().StringVar(&command.BuildBuddyUrl, "url", "", "build buddy url instance")
-	command.Flags().StringVar(&command.ClusterName, "cluster", "", "name of the cluster")
+	command.Flags().StringVar(&command.BuildBuddyUrl, "url", "https://buddy.bazel.corp.enfabrica.net", "build buddy url instance")
+	command.Flags().StringVar(&command.ClusterHost, "buildbarn_host", "build.local.enfabrica.net:8000", "host+port of BuildBarn instance")
```
minor-fixes, line 93: Defaults probably need to come out; convenience for testing

aaahrens (approved):

PR #445, opened 2022-02-05
fuse: fix outputs mount command

aaahrens:
problems: 
- cluster host in the url should not include port
- default port was wrong

[INFRA-518]

[INFRA-518]: https://enfabrica.atlassian.net/browse/INFRA-518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/enkit/outputs/commands.go
+++ b/enkit/outputs/commands.go
@@ -24,7 +24,7 @@ import (
 )
 
 var (
-	localTunnelPort = 8822
+	localTunnelPort = 8866
 )
 
 type Root struct {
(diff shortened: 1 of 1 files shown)


enkit/outputs/commands.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 var (
-	localTunnelPort = 8822
+	localTunnelPort = 8866
```
minor-fixes, line 27: Instead of changing this - fix line 148 to pass port instead

minor-fixes (approved):
Please add `Jira: INFRA-whatever` to the change description as well

PR #447, opened 2022-02-07
fuse: implement shutdown command

aaahrens:
Implements a the basic unmount function, which just deletes the soft symlink at <root>/invocation  

Also removes an ineffectual assignment.

### DIFF
--- a/enkit/outputs/commands.go
+++ b/enkit/outputs/commands.go
@@ -199,7 +199,8 @@ func (c *Mount) Run(cmd *cobra.Command, args []string) error {
 
 type Unmount struct {
 	*cobra.Command
-	root *Root
+	root       *Root
+	Invocation string
 }
 
 func NewUnmount(root *Root) *Unmount {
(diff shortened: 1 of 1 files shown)


enkit/outputs/commands.go
```diff
@@ -220,7 +220,14 @@ func NewUnmount(root *Root) *Unmount {
 }
 
 func (c *Unmount) Run(cmd *cobra.Command, args []string) error {
-	return fmt.Errorf("`enkit outputs unmount` is unimplemented")
+	if len(args) != 1 {
```
minor-fixes, line 223: Other commands take an invocation_id flag (I think) - can you make sure this command matches the others as much as possible?

minor-fixes (approved):

enkit/outputs/commands.go
aaahrens: Done, with -i flag

enkit/outputs/commands.go
```diff
@@ -220,7 +220,14 @@ func NewUnmount(root *Root) *Unmount {
...
+		return errors.New("must provide a valid invocation")
+	}
+	invoPath := filepath.Join(c.root.OutputsRoot, args[0])
+	if err := os.RemoveAll(invoPath); err != nil {
```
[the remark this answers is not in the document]
aaahrens: it is now os.Remove

PR #448, opened 2022-02-07
fuse: implement shutdown command

aaahrens:
What this does: 
- implements the shutdown command by:
* unmounting the mount directory
* killing the bb_clientd process
* deleting the output base

### DIFF
--- a/enkit/outputs/commands.go
+++ b/enkit/outputs/commands.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strconv"
 	"time"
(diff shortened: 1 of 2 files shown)


enkit/outputs/commands.go
```diff
@@ -303,5 +303,24 @@ func NewShutdown(root *Root) *Shutdown {
...
 func (c *Shutdown) Run(cmd *cobra.Command, args []string) error {
-	return fmt.Errorf("`enkit outputs Shutdown` is unimplemented")
+	bbOpts := bbexec.NewClientOptions(
+		c.root.Log,
+		0, /* tunnel port does not matter in this case */
+		c.root.OutputsRoot,
+	)
+	bbClient, err := bbexec.MaybeStartClient(bbOpts)
+	if err != nil {
+		return err
```
minor-fixes, line 313: I think we still want to attempt the other steps even if this fails

enkit/outputs/commands.go
```diff
@@ -303,5 +303,24 @@ func NewShutdown(root *Root) *Shutdown {
 }
 
 func (c *Shutdown) Run(cmd *cobra.Command, args []string) error {
-	return fmt.Errorf("`enkit outputs Shutdown` is unimplemented")
+	bbOpts := bbexec.NewClientOptions(
+		c.root.Log,
+		0, /* tunnel port does not matter in this case */
+		c.root.OutputsRoot,
+	)
+	bbClient, err := bbexec.MaybeStartClient(bbOpts)
```
minor-fixes, line 311: This merits a comment - maybe `// MaybeStartClient is used here to bind a client handle to an existing process, so that we can kill it. It may start a process that will be then killed quickly, which is acceptable but not ideal.`

enkit/outputs/commands.go
```diff
@@ -303,5 +303,24 @@ func NewShutdown(root *Root) *Shutdown {
...
+	if err != nil {
+		return err
+	}
+	var errs []error
+	if err := bbClient.Shutdown(); err != nil {
+		errs = append(errs, fmt.Errorf("error maybe? killing the process of existing bb_clientd %v", err))
+	}
+	if err := os.RemoveAll(c.root.OutputsRoot); err != nil {
```
minor-fixes, line 319: Is there a method on bbOpts we can use/make public instead? Incidentally, I think there's already code that effectively deletes everything inside OutputsRoot, but leaves OutputsRoot intact (the user might be surprised if OutputsRoot is gone after a shutdown instead of just empty)

enkit/outputs/commands.go
```diff
@@ -303,5 +303,24 @@ func NewShutdown(root *Root) *Shutdown {
...
+		errs = append(errs, err)
+		c.root.Log.Errorf("error removing output root %s %v", c.root.OutputsRoot, err)
```
minor-fixes, line 321: This should be the only real error - if we fail to clean the FS as expected the command should fail. The other steps (fusermount, killing bb_clientd) are supposed to increase the likelihood that the FS clean goes well

enkit/outputs/commands.go
```diff
@@ -303,5 +304,29 @@ func NewShutdown(root *Root) *Shutdown {
...
+	)
+	if err := bbOpts.MaybeKill(); err != nil {
+		errs = append(errs, err)
+		return fmt.Errorf("error maybe? killing the process of existing bb_clientd %v", err)
+	}
+
+	if err := os.RemoveAll(c.root.OutputsRoot); err != nil {
+		errs = append(errs, err)
+		c.root.Log.Errorf("error removing output root %s %v", c.root.OutputsRoot, err)
+	}
```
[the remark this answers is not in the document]
minor-fixes: (ignore)

minor-fixes (approved):

PR #449, opened 2022-02-07
bestie: Use zipstream package to read outputs.zip

gpaussaenf:
Change the handling of the outputs.zip file so that it is not
read into memory prior to processing each of its file elements.
Use the https://github.com/xenking/zipstream package to iterate
through each of the files, looking for file names ending in
.metrics.pb.

Tested: Ran BES Endpoint service both locally and in the staging
cluster, where it detected the test.metrics.pb file produced by
a standalone test app.

Jira: INFRA-504

### DIFF
--- a/bestie/server/testresult.go
+++ b/bestie/server/testresult.go
@@ -1,13 +1,13 @@
 package main
 
 import (
-	"archive/zip"
-	"bytes"
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -16,6 +16,7 @@ import (
 	"github.com/enfabrica/enkit/lib/kbuildbarn"
 	"github.com/enfabrica/enkit/lib/multierror"
 	bes "github.com/enfabrica/enkit/third_party/bazel/buildeventstream" // Allows prototext to automatically decode embedded messages
+	"github.com/xenking/zipstream"
 	"google.golang.org/genproto/googleapis/devtools/build/v1"
 	"google.golang.org/protobuf/proto"
 )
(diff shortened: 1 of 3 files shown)


bestie/server/testresult.go
```diff
@@ -142,77 +141,89 @@ func handleTestResultEvent(bazelBuildEvent bes.BuildEvent, streamId *build.Strea
 		// Attempt to read the zip file failed.
 		return fmt.Errorf("Error reading %s file: %w", outFileName, readErr)
 	}
```
minor-fixes, line 143: It's a bit easier to verify correctness if the close happens in the same function, close to where the open happened (via `defer()`)

In this case - maybe `defer zipFileCloser.Close()` right after the switch statement, and then `extractZipFiles` can take an `io.Reader`, and doesn't have to close

bestie/server/testresult.go
```diff
@@ -142,77 +141,89 @@ func handleTestResultEvent(bazelBuildEvent bes.BuildEvent, streamId *build.Strea
...
-	if len(fileBytes) == 0 {
-		// Something went wrong reading zip file -- no data.
-		return fmt.Errorf("Error reading %s file: no data", outFileName)
-	}
 	// Process test metrics output file(s).
 	// Each output file contains a single (potentially large) protobuf message.
 	// For now, it's up to the client to split large metric datasets into multiple .metrics.pb files.
-	if err := extractZippedFiles(stream, fileBytes); err != nil {
+	if err := extractZippedFiles(stream, zipFileCloser); err != nil {
```
minor-fixes, line 147: nit: maybe extractZippedFiles gets renamed to `processZip`, since it doesn't extract all files anymore?

minor-fixes (approved):

bestie/server/testresult.go
gpaussaenf: OK, makes sense.

bestie/server/testresult.go
gpaussaenf: OK

PR #450, opened 2022-02-08
Bazel FUSE: Implement run with zip_path, dir_path

minor-fixes:
This change implements `enkit outputs run` when either a zip file is
passed (perhaps from a previous `enkit outputs mount`) or a dir is
passed (maybe from an already-unzipped outputs.zip)

Tested: `enkit outputs run --zip_path ~/tmp/outputs.zip` manually on
local zip file - reroots the zip to `/enfabrica`, and starts a shell at
that path.

`enkit outputs run --zip_path ~/tmp/outputs.zip cat cartoons.txt` skips
the shell and instead prints the contents of the file.

Jira: INFRA-514

### DIFF
--- a/enkit/outputs/BUILD.bazel
+++ b/enkit/outputs/BUILD.bazel
@@ -6,8 +6,10 @@ go_library(
     importpath = "github.com/enfabrica/enkit/enkit/outputs",
     visibility = ["//visibility:public"],
     deps = [
+        "//faketree/exec:go_default_library",
         "//lib/bes:go_default_library",
         "//lib/client:go_default_library",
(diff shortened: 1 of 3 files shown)


ccontavalli (approved):

aaahrens (approved):
Works good, issue, but has to be prefixed with 
SHELL=/bin/bash

faketree adds in flags that are bash only, We should fix that upstream

minor-fixes:
> Works good, issue, but has to be prefixed with SHELL=/bin/bash
> 
> faketree adds in flags that are bash only, We should fix that upstream

ack

PR #451, opened 2022-02-09
Bazel FUSE: Separate request host from tunnel endpoint

minor-fixes:
When using a tunnel, two hostnames are requried:
* The host to tunnel to (`pbzl.service.enfabrica.net`)
* The host to make requests to (`build.local.enfabrica.net`)

The latter is required because the tunnel host is typically a
non-routable internal IP, whereas this host resolves to localhost if a
tunnel is required; additionally, ingress will route requests based on
the destination host, which is expected to be the build.local... host.

This change breaks the single host flag into two flags: One is used to
resolve to a non-localhost address (if no tunnel is required) or a
localhost address (if a tunnel is required), and requests are to be
directed to this URL. The new flag specifies the endpoint to tunnel to
if a tunnel is required.

Tested:

```
bazelisk run //enkit:enkit -- \
           outputs \
           mount \
           --buildbarn_host=<redacted> \
           --buildbarn_tunnel_target=<redacted> \
           --api_key=<redacted> \
           --buildbuddy_url=<redacted> \
           --invocation=532263bf-6522-4d05-8c63-f15621e6de4c &&
cat ~/outputs/532263bf-6522-4d05-8c63-f15621e6de4c/test.log
```

...on the corp network (no tunnel required)

Case that requires tunnel not tested.

Jira: INFRA-518

### DIFF
--- a/enkit/outputs/commands.go
+++ b/enkit/outputs/commands.go
@@ -76,11 +76,12 @@ type Mount struct {
 	*cobra.Command
 	root *Root
 
-	BuildBuddyApiKey string
-	BuildBuddyUrl    string
-	ClusterHost      string
-	DryRun           bool
-	InvocationID     string
+	BuildBuddyApiKey      string
+	BuildBuddyUrl         string
+	BuildbarnHost         string
+	BuildbarnTunnelTarget string
+	DryRun                bool
+	InvocationID          string
 }
 
 func NewMount(root *Root) *Mount {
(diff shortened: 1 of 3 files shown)


enkit/outputs/commands.go
```diff
@@ -94,11 +95,12 @@ func NewMount(root *Root) *Mount {
 		},
 		root: root,
 	}
-	command.Flags().StringVar(&command.BuildBuddyApiKey, "api-key", "", "build buddy api key used to bypass oauth2")
-	command.Flags().StringVar(&command.BuildBuddyUrl, "url", "", "build buddy url instance")
-	command.Flags().StringVar(&command.ClusterHost, "buildbarn_host", "", "host:port of BuildBarn instance")
+	command.Flags().StringVar(&command.BuildBuddyApiKey, "api_key", "", "build buddy api key used to bypass oauth2")
```
ccontavalli, line 98: as per other PR: most of enkit uses `-` for flag separation. I would stick to the convention used throughout, it is odd to have a single command behave differently.

ccontavalli (approved):

enkit/outputs/commands.go
minor-fixes: Ack; filed INFRA-550

PR #453, opened 2022-02-09
enkit outputs: Don't sleep waiting for bb_clientd

minor-fixes:
This change replaces the 5-second sleep waiting for bb_clientd to start
and create the mount with a polling loop that looks for an external
signal that bb_clientd is ready. The external signal chosen is that one
of the expected mount directories exists. The code will poll for up to 5
seconds instead, which should allow it to move on more quickly once
bb_clientd is ready

Tested: Ran:

```
bazel run //enkit -- shutdown
bazel run //enkit -- outputs mount \
        --buildbarn_host=<redacted> \
        --buildbarn_tunnel_target=<redacted> \
        --api_key=<redacted> \
        --buildbuddy_url=<redacted> \
        --invocation=6c01fa88-3fe5-47e8-8a5a-6fb17b2330eb
```

which mounted successfully much more quickly.

Jira: INFRA-547


lib/kbuildbarn/exec/bb_clientd.go
```diff
@@ -75,6 +75,25 @@ func MaybeStartClient(o *ClientOptions) (*Client, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error starting bb_clientd: %w", err)
 	}
+
+	// Wait for a directory under the mount dir to appear, so that we know
+	// bb_clientd is ready, up until `timeout`
+	stopTime := time.Now().Add(timeout)
```
ccontavalli, line 81: retry lib? 5 times retry, 100 ms in between? not sure it's a good idea, but would save some typing here.

lib/kbuildbarn/exec/bb_clientd.go
```diff
@@ -75,6 +75,25 @@ func MaybeStartClient(o *ClientOptions) (*Client, error) {
...
+	for {
+		time.Sleep(100*time.Millisecond)
+		if time.Now().After(stopTime) {
+			return nil, fmt.Errorf("failed to find mount directory ready after %v", timeout)
```
ccontavalli, line 85: when you return here for errors, you are leaking the process? eg, next run of kbuildbarn may fail / be unreliable / be flaky, as a background bb_clientd may (or may not) still be running (and trying, or not to be using the directory)?

I would recommend some form of defer to kill the process if this function returns with error. Ideally with a waitpid() equivalent, so you can log if it failed in any way (right now failure is silently discarded?)

ccontavalli (approved):

lib/kbuildbarn/exec/bb_clientd.go
minor-fixes: Retry lib seems to work great! Thanks for the pointer

lib/kbuildbarn/exec/bb_clientd.go
minor-fixes: Done; now using named returns, and after the process is created a defer checks the returned error and runs the existing Shutdown codepath to clean up if necessary.

lib/kbuildbarn/exec/bb_clientd.go
```diff
@@ -75,11 +76,40 @@ func MaybeStartClient(o *ClientOptions) (*Client, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error starting bb_clientd: %w", err)
 	}
+	// From here on out, any returns should return a client handle instead of
+	// nil, so that we can shutdown in case of an error to avoid leaking the
```
ccontavalli, line 80: is this comment correct? well, ok, if ret is declared as a return value. If not, it is just a variable captured by the functor? (can you "name the return value something else/not name it", and just have `if ret != nil && err != nil {
ret.Shutdown()`)?

That way, you still return `nil`, which is more idiomatic for the caller, but the client is still shutdown.
I've also often seen `retErr` simply called `err`.

ccontavalli (approved):
thank you - note that I would have been fine with a TODO/next PR.

lib/kbuildbarn/exec/bb_clientd.go
minor-fixes: I see what you're saying - creating a client and capturing it in the functor makes the rest of the code less error-prone; switched to that instead.

> I've also often seen retErr simply called err.

With the aliasing that can happen in e.g. `if` blocks, my mind melts. For this reason I like to leave `err` for normal use in the function, and a specifically-named error to be checked in the defer, to disambiguate the two.

gpaussaenf (approved):
I'll approve since I'm listed as a reviewer due to the bestie/server/BUILD.bazel change. Carlo has already commented on the bb_clientd code changes.

PR #457, opened 2022-02-10
bestie: Rename metric tag test_name to test_target

gpaussaenf:
Just renaming one of the metric tags that get inserted by the
BES Endpoint using information from the BES TestResult event.
The 'test_name' tag identifier can be easily confused with a
testcase name tag supplied by the test client application.
Since this represents the test target specified for the
'bazel test' command, the 'test_target' tag identifier
seems more appropriate.

Tested: Verified using a localhost instance of the BES Endpoint
that the new tag identifier is processed correctly and appears
in the BigQuery testmetrics table.

Jira: INFRA-551


msgoldflam (approved):
Looks like you got all the right places for this minor change

minor-fixes (approved):

PR #458, opened 2022-02-12
flextape: introduce prioritization support.

ccontavalli:
Background:
Before this PR, licenses were always allocated in FIFO order.

In this PR:
- allow to specify a prioritization strategy (per license).
- define an interface to provide different prioritization strategies,
  refactor the code to allow for the plumbing.
- create a strategy that spreads licenses evenly across users.
- added extensive tests.

Note that this PR still does not allow to allocate multiple licenses
at once. The rough plan to allow for that is to expand the concept
of "invocation", so it links to all the types it is trying to allocate,
and extend the prioritizer to account for that.

For now, I just wanted to allow better license spreading, and figure out a
reasonable interface for queue reordering.

Tested: added significant unit tests, all passing.


ccontavalli:
For review: I recommend starting from the prioritization.go file, which explains the logic and algorithm. The rest is plumbing.

flextape/service/prioritizer.go
```diff
@@ -0,0 +1,154 @@
...
+// EvenOwnerAllocationsPrioritizer is a prioritizer that tries to spread
+// license allocations evenly across owners.
+//
+// Time is not part of the equation: if user A has 10 licenses allocated
+// and an user B comes by, it will prioritize B over A until there is a
+// 5 ~ 5 distribution in number of licenses.
+type EvenOwnersPrioritizer struct {
+	// Key: invocation.Owner, value represents # of requests queued,
+	// monothonically increasing.
+	enqueued map[string]uint64
```
minor-fixes, line 61: Are we worried about race conditions/protecting these with mutexes? I suspect that a top-level mutex ends up protecting races here in practice, but on the flipside if mutexes were added here as appropriate probably the top-level mutex would result in the lower-level ones being uncontended and the performance impact being negligible.

flextape/service/queue.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+// invocationq is a simple queue of invocations.
+//
+// Beside implementing the classic Enqueue and Dequeue methods:
+// - It partially implements the sort.Interface to allow the easy
+//   implementation of invocation prioritization policies.
+// - It maintains per-entry state (QueueID) so it's easy to tell
+//   the position of each invocation by just looking at that value.
+type invocationq []*invocation
```
minor-fixes, line 14: nit: This name looks like a typo of `invocation` - can it be `invocationQueue` or at least `invocationQ`, which are at least visibly distinct from `invocation`?

flextape/service/queue.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (iq *invocationq) Enqueue(x *invocation) Position {
+	position := Position(len(*iq))
+	offset := QueueID(1)
+	if len(*iq) > 0 {
+		offset = (*iq)[0].QueueID
+	}
+
+	x.QueueID = offset + QueueID(position)
+	(*iq) = append(*iq, x)
```
minor-fixes, line 24: Same question here about locking

flextape/service/queue.go
```diff
@@ -0,0 +1,137 @@
...
+	return position
+}
+
+func (iq *invocationq) Dequeue() *invocation {
+	if len(*iq) <= 0 {
+		return nil
```
minor-fixes, line 30: Returning an error here makes the API a bit harder, but reduces the odds of someone forgetting a nil check

flextape/service/queue.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+
+		inv.QueueID -= QueueID(count)
+		newQueue = append(newQueue, inv)
+	}
+	(*iq) = newQueue
+	return count
+}
+
+func (iq *invocationq) Walk(walker Walker) (*invocation, Position) {
```
minor-fixes, line 76: Public functions need docstrings (I got confused between this and `Walker` above)

flextape/service/service_test.go
```diff
@@ -1090,3 +1122,487 @@ func TestJanitor(t *testing.T) {
...
+
+	// Dear donnie at at next poll will find out that his allocations were bumped.
+	for ix, id := range dqids[4:] {
+		req.Invocation.Id = id
+		resp, err := server.Allocate(ctx, req)
+		assert.Nil(t, err, "error %s", err)
+
+		queued, converted := resp.ResponseType.(*fpb.AllocateResponse_Queued)
+		assert.True(t, converted)
+		// Note the +2 here, comes from the # of allocations from joe.
```
minor-fixes, line 1541: +3, not +2?

flextape/service/prioritizer.go
```diff
@@ -0,0 +1,154 @@
+package service
+
+// Prioritizer is an object capable of sorting a queue in order of priority.
+//
+// In order for the prioritizer to have enough data to make the prioritization
+// decision, it must be invoked every time an entry is queued and dequeued, and
+// every time a license is allocated and released.
+type Prioritizer interface {
+	// Enqueued is called every time an invocation is queued.
+	Enqueued(inv *invocation)
```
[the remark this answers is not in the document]
ccontavalli: done

flextape/service/prioritizer.go
```diff
@@ -0,0 +1,154 @@
...
+//
+// Time is not part of the equation: if user A has 10 licenses allocated
+// and an user B comes by, it will prioritize B over A until there is a
+// 5 ~ 5 distribution in number of licenses.
+type EvenOwnersPrioritizer struct {
+	// Key: invocation.Owner, value represents # of requests queued,
+	// monothonically increasing.
+	enqueued map[string]uint64
+	// Key: invocation.Owner, value represents # of requests dequeued,
+	// monothonically increasing.
```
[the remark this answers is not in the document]
ccontavalli: done

flextape/service/queue.go
ccontavalli: done

flextape/service/prioritizer.go
ccontavalli: I'd rather not add locking here, reasons:
1. the `license` object has no locking, it is not thread safe by contract. It is made thread safe by the caller, service, which has a mutex protecting it. license uses different data structures that are not thread safe, this would be just another one of them, and having a top level caller ensure thread safety is a common pattern in go. Having just this one thread safe would make the code harder to read/update (when reading the code: "why is this data structure thread safe and the others are not? let me find where in the code it is run from multiple threads!" - a bit like using sync.Map or atomic in a class that otherwise does not look like it is used by multiple threads).
2. the `Sorter()` method has to access the content of the invocation. Is it safe to access? does it need a lock on invocation? or is each invocation already locked by the caller? Can this cause lock ordering deadlock? ... for the prioritizer to be thread safe / have useful locking, invocation has to have some thread safety guarantee - probably a lock per invocation? This change would be necessary, but imho out of scope for the PR. IF we say the caller guarantees invocation safety, then I think it would be best to have all methods expect a similar contract?

flextape/service/queue.go
ccontavalli: Similar answer to above. The lock ordering fears are worse here as `Walk` and `Filter` take callbacks, it would be easy to introduce a deadlock accidentally. Eg, add more risks than benefits?

flextape/service/service.go
```diff
@@ -100,13 +110,38 @@ func New(config *fpb.Config) *Service {
...
 // invocation maps to a particular command invocation that has requested a
 // license, and its associated metadata.
 type invocation struct {
 	ID          string    // Server-generated unique ID
 	Owner       string    // Client-provided owner
 	BuildTag    string    // Client-provided build tag. May not be unique across invocations
 	LastCheckin time.Time // Time the invocation last had its queue position/allocation refreshed.
+
+	QueueID  QueueID     // Position in the queue. 0 means the invocation has not been queued yet.
+	Priority interface{} // Priority of the item. Type and value depends on Prioritizer.
```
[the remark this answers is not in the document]
ccontavalli: removed the field.

It was there for 3 reasons mainly:
1. how the code came to be - my initial prioritizer implementation was pretty much assigning an integer to each entry, while letting the service just sort by that priority. It was changed when I realized the per-user state would change all the time, and Priority per entry would need to be updated over time by scanning the queue and recomputing it. With that, an interface based on sort seemed to make more sense but priority remained per entry.
2. I expected each prioritizer to keep some state per entry - state that would be accessible/dumped through the prioritizer. I also liked the idea of explicitly showing that the prioritizer had some state per invocation, and to tie it to the invocation, as to guarantee no leaks and proper ownership. But this argument is weak, as there is other state maintained by the prioritizer, so we still need to guarantee proper handling of state.

ccontavalli:
PTAL?

minor-fixes (approved):

PR #459, opened 2022-02-14
bestie: Add ResultMetric to protobuf definition

gpaussaenf:
Defined a ResultMetric message type to support pass/fail metrics
distinct from the existing TestMetric used for performance and
throughput values and the like.

Tested: Used a local instance of the BES Endpoint with some sample
metric data.

Jira: INFRA-563

### DIFF
--- a/bestie/proto/test_metrics.proto
+++ b/bestie/proto/test_metrics.proto
@@ -32,7 +32,8 @@ message ResultMetric {
     enum Result {
         FAIL = 0;
         PASS = 1;
-        INCOMPLETE = 2;
+        SKIP = 2;
+        INCOMPLETE = 3;
     }
     Result result = 2;
     repeated KeyValuePair tags = 3;
--- a/bestie/proto/test_metrics.proto
+++ b/bestie/proto/test_metrics.proto
@@ -5,6 +5,41 @@ package bestie.proto;
 
 option go_package = "github.com/enfabrica/enkit/bestie/proto";
 
+// Each metric tag must be supplied in the form of a key/value string pair.
+message KeyValuePair {
+    string key = 1;
+    string value = 2;
+}
+
+// Specialized BigQuery test result definition compatible with Prometheus remote read adapter.
+//
+// This message is similar to TestMetric, but with added restrictions to make it
+// less ambiguous for the message sender to construct properly. The same BigQuery
+// table holds both TestMetric and ResultMetric rows from a variety of test applications
+// (clients), therefore standardizing on a specific metricname ('testresult') is
+// essential. This name, a corresponding 'result' tag, and a numeric value is inserted
+// by the BES Endpoint service as it processes a ResultMetric message, therefore these are
+// not specifically defined by the message fields.
+//
+// The allowed Result values are PASS or FAIL, with INCOMPLETE indicating a test
+// that was intended to be run, but could not due to some extenuating condition, for
+// example reliance upon a specific DUT or traffic generator that was unavailable.
+message ResultMetric {
+    enum Type {
+        SUMMARY = 0; // default, if field not specified
+    }
+    Type type = 1;
+    enum Result {
+        FAIL = 0;
+        PASS = 1;
+        INCOMPLETE = 2;
+    }
+    Result result = 2;
+    repeated KeyValuePair tags = 3;
+    //double value = 4;  // not used (provided by BES Endpoint)
+    int64 timestamp = 5;  // nanoseconds since epoch (e.g. Python: time.time_ns())
+}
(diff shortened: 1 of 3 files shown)


bestie/proto/test_metrics.proto
```diff
@@ -5,6 +5,42 @@ package bestie.proto;
...
+//
+// This message is similar to TestMetric, but with added restrictions to make it
+// less ambiguous for the message sender to construct properly. The same BigQuery
+// table holds both TestMetric and ResultMetric rows from a variety of test applications
+// (clients), therefore standardizing on a specific metricname ('testresult') is
+// essential. This name, a corresponding 'result' tag, and a numeric value is inserted
+// by the BES Endpoint service as it processes a ResultMetric message, therefore these are
+// not specifically defined by the message fields.
+//
+// The allowed Result values are PASS or FAIL, with INCOMPLETE indicating a test
```
ccontavalli, line 24: pass and fail for a test should already be captured either by the normal BES protocol, or by parsing the test.xml file in the stream that contains pass/fail information for individual tests?

if we keep the result, can we define a ResultMetric as a TestMetric with on optional Result? eiter as:
```
message ResultMetric {
  ...
  Result result;
  TestMetric metric;
}
```

or even just add the `Result` field in the `TestMetric` directly? set to an "UNSET" default value? so a test setting it explicitly marks it as FAIL/PASS/SKIP/INCOMPLETE, a test that does not know about the field leaves the default 0 value of UNSET?

gpaussaenf: The reason I defined ResultMetric separate from TestMetric is that they convey different information. A single test case can produce multiple TestMetrics, but only one ResultMetric (e.g. pass/fail). Since these end up in the same BigQuery 'testmetrics' table, the ResultMetric needs a distinct metricname identifier "testresult" to allow querying. Instead of relying on the test application to spell it right, the ResultMetric protobuf msg allows the BES Endpoint to take care of the metricname for consistency.

I'm not using the test.xml file, as it does not carry enough useful information for creating BigQuery table entries. Hence, I'm relying on test_metrics.proto and the resultant *.metrics.pb file to give us what we need, which we can control.

bestie/proto/test_metrics.proto
```diff
@@ -5,6 +5,42 @@ package bestie.proto;
...
+// that was intended to be run, but could not due to some extenuating condition, for
+// example reliance upon a specific DUT or traffic generator that was unavailable.
+message ResultMetric {
+    enum Type {
+        SUMMARY = 0; // default, if field not specified
+    }
+    Type type = 1;
+    enum Result {
+        FAIL = 0;
```
minor-fixes, line 33: Typically with enums you want an UNKNOWN as the 0 value so that if the client forgets to set it it's very obvious

bestie/proto/test_metrics.proto
```diff
@@ -5,6 +5,42 @@ package bestie.proto;
 
 option go_package = "github.com/enfabrica/enkit/bestie/proto";
 
+// Each metric tag must be supplied in the form of a key/value string pair.
+message KeyValuePair {
```
minor-fixes, line 9: `message Tag` is shorter, and if Tag ever becomes something else, it's not complected with other uses of `KeyValuePair`

bestie/proto/test_metrics.proto
```diff
@@ -51,10 +87,13 @@ message TestMetric {
     int64 timestamp = 4;  // nanoseconds since epoch (e.g. Python: time.time_ns())
 }
 
-// Each metric tag must be supplied in the form of a key/value string pair.
-message KeyValuePair {
-    string key = 1;
-    string value = 2;
+// Define a metric selector message layer to control which kind of metric message
+// is represented.
+message MetricSelector {
```
minor-fixes, line 92: `Metric` isn't descriptive enough?

bestie/proto/test_metrics.proto
minor-fixes: > Instead of relying on the test application to spell it right

...my vote is to rely on the test application to spell it right. The test application should be going through a test metrics client library which emits this file, so really it's actually the client library's responsibility to emit the testresult metric with the proper spelling.

Doing it the way proposed in this PR necessitates:
* client libraries emitting a special result metric anyway
* all this extra schema surface area to understand/code around

Doing it where client libraries know to emit a specially named metric - well, the libraries have to special-case that metric either way, so might as well avoid all this code

gpaussaenf:
Abandoning this PR in favor of using a text.xml file to process the test case pass/fail results by the BES Endpoint.

PR #462, opened 2022-02-15
bkup: add volume id support, filters

jonathan-enf:
bkup: add volume id support, filters

When I had to use bkup to restore an old laptop's file to a new laptop, one
snag I ran into is that bkup uses the docker ID as a UUID to disambiguate
different machines owned by the same user.  But if I want to restore another
machine's data to this one, I need a way to override the generated ID.

This PR solves the problem by dropping a "volume_id.txt" file in the
.config/rclone directory.  If this file doesn't exist, it is created
and populated with the docker image ID.  This file can be overriden
to restore backups that were created on a different machine.

I also added a "lsvolumes" command to list all available volume ids.

Also, I learned that git repositories are too big to realistically
save and restore.  These are *probably* backed up to remote servers,
so I updated bkup's filters to skip them entirely.

TODO(jonathan): prompt the user for a human-readable ID instead.

Tested:

*  Restored a backup created on another machine.
*  Ran "lsvolumes".

### DIFF
--- a/scripts/bkup
+++ b/scripts/bkup
@@ -41,7 +41,7 @@ readonly BUCKET="rclone-devops-284019"
 readonly PAGER="${PAGER:-less}"
 
 LOCAL_VOLUME_ID="$(\
-  cat /proc/1/cgroup | grep -o -E 'docker/([a-f0-9]{12})' | sed 's/docker.//' | head -n 1 )"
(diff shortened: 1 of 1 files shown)


jonathan-enf:

ccontavalli (approved):

PR #464, opened 2022-02-15
marshal: Fix type detection when passed a URL

minor-fixes:
When a YAML config is included in another config today by URL, the
config language detection by extension fails because it is passed a URL
that has a query string suffix instead of a local file path. When this
happens, the default unmarshaler of JSON is used, which causes obtuse
parse errors to be thrown when parsing other formats.

This change first tries to parse the input path as a URL, falling back
to the assumption that the path is a local file path if URL parsing
fails. URL parsing allows us to pick out the path component, ignoring
all other components like query strings that may suffix the string.

Tested: Added unit test to demonstrate issue and verify fix.

Jira: INFRA-487

### DIFF
--- a/lib/config/marshal/select.go
+++ b/lib/config/marshal/select.go
@@ -31,7 +31,9 @@ var Known = []FileMarshaller{
 // Represents a sorted list of marshallers. Lowest index is the most preferred marshaller.
 type FileMarshallers []FileMarshaller
 
-// ByExtension returns the first FileMarshaller based on the extension of the path provided.
+// ByExtension returns the first FileMarshaller based on the extension of the
+// path provided. `path` can be either a local file path or a URL; if `path` is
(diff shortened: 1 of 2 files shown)


lib/config/marshal/select.go
```diff
@@ -31,6 +33,10 @@ type FileMarshallers []FileMarshaller
 
 // ByExtension returns the first FileMarshaller based on the extension of the path provided.
 func (fm FileMarshallers) ByExtension(path string) FileMarshaller {
+	u, err := url.Parse(path)
```
ccontavalli, line 36: update docstring to specify that URLs are supported? Good catch, by the way.

ccontavalli (approved):

minor-fixes:
Merging over known tunnel_test failure

PR #465, opened 2022-02-16
enkit outputs: Assert that user is logged in

minor-fixes:
The command will fail to fetch an included config that has necessary
default flags if the user is not logged in. This change adds a
validation hook on the top-level command that ensures the user is logged
in via fetching the cookie, so that we can:

* ensure that flag defaults are set
* in the case that flag defaults are not set, offer a sane error message
  to the user.

Tested: Removed credentials by deleting `~/.config/enkit` and verified
that error message appeared

Jira: INFRA-487

### DIFF
--- a/enkit/outputs/commands.go
+++ b/enkit/outputs/commands.go
@@ -72,12 +72,21 @@ func NewRoot(base *client.BaseFlags) (*Root, error) {
 	}
 	defaultOutputsRoot := filepath.Join(homeDir, "outputs")
 
+	rc.Command.PersistentPreRunE = rc.validate
 	rc.PersistentFlags().StringVar(&rc.OutputsRoot, "outputs-root", defaultOutputsRoot, "Root dir of mounted outputs")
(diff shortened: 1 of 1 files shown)


enkit/outputs/commands.go
```diff
@@ -80,6 +81,14 @@ func NewRoot(base *client.BaseFlags) (*Root, error) {
 	return rc, nil
 }
 
+func (rc *Root) validate(cmd *cobra.Command, args []string) error {
+	_, _, err := rc.BaseFlags.IdentityCookie()
```
minor-fixes, line 85: I'm not sold on this approach because:

* I don't think this code handles the case when the user has stale credentials (only the case when the user has no credentials)
* This means a not-logged-in user can't use this command, even if they wanted to manually fill out the required flags (though in this specific case that is moot, because probably the command will require login for a tunnel)

Either we do this approach but replace the IdentityCookie check with a stronger indicator of whether the creds actually work (is there an endpoint we can hit?) and/or we ensure that config loading throws errors/warnings appropriately as we fail to fetch includes that require auth.

PR #466, opened 2022-02-16
bkup: prompt user for a volume identifier

jonathan-enf:
bkup: fix filters, prompt user for a volume identifier

Auto-generating a volume UUID from the docker image is confusing.  Instead,
it's better to just ask the user to name the backup volume they want to save to
or restore from (for example: "chromebook1").  We only prompt once and then
save this cookie away locally.

While working on this PR, I noticed that my file filters weren't working the
way I expected them to.  It turns out that rclone's filter file format doesn't
support comments on the same lines as filters, so I removed them.

Tested:

Filter rules are tested with:

    rclone ls ~ --filter-from ~/.config/rclone/filter-list.txt

and examining the files reported.

UUID prompting was tested by removing my laptop's `volume_id.txt` file and
running `bkup lsvolumes`.

### DIFF
--- a/scripts/bkup
+++ b/scripts/bkup
@@ -239,6 +239,8 @@ EOT
 - /.Trash/                          # do I need to say more?
 - /.local/share/Trash/
 - /.aptitude/                       # cached packages lists
+- /.rustup/                         # local rust installation
+- /.poetry/                         # python poetry installation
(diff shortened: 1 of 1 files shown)


jonathan-enf:

jonathan-enf:
ping.

ccontavalli (approved):

PR #467, opened 2022-02-16
enkit outputs: Change mount paths for various files

minor-fixes:
Prior to this change, all files mentioned in the output stream are
mounted in the scratch directory under the invocation ID using the path
specified in the BuildEvent message verbatim. This change modifies paths
in the following way:

* If the artifact is a test artifact, it is rerooted at
  `bazel-testlogs/$LABEL/$RUN/$ATTEMPT/`, where `$LABEL` is a subdirectory
  generated from the test target's label, `$RUN` is the run number
  (there may be multiple runs for a given test if `--runs_per_test` was
   passed on the bazel commandline) and `$ATTEMPT` is the attempt number
  (if bazel retried a test marked flaky).

* If the artifact is a test artifact, its path is unmangled by replacing
  `__` with `/`, which allows paths like `test.outputs__outputs.zip` to
  be the more intelligible `test.outputs/outputs.zip`.

* If the artifact is a build artifact (has a path prefix of
  `bazel-out/$CONFIG/bin` for any `$CONFIG`, it is rerooted under a
  `bazel-bin` subdirectory.

This change also modifies the unit tests to follow conventions:

* Unit tests are part of the same package as the actual library
* Expected Hardlink list is constructed directly and tested with
  `assert.ElementsMatch`, which removes a nil panic from the test when
  the test fails.

Tested:

* Mounted last HW regression to see all testlogs: `enkit outputs
  mount --invocation-id 0c7c3857-c823-4092-bb0d-8e7ec7c80a73`
* Modified unit tests to ensure `subdir` is added to dest path correctly

Jira: INFRA-546

### DIFF
--- a/lib/kbuildbarn/BUILD.bazel
+++ b/lib/kbuildbarn/BUILD.bazel
@@ -24,8 +24,8 @@ go_test(
         "protoparse_test.go",
         "urls_test.go",
     ],
+    embed = [":go_default_library"],
     deps = [
-        ":go_default_library",
         "//lib/bes:go_default_library",
         "//third_party/bazel/src/main/java/com/google/devtools/build/lib/buildeventstream/proto:build_event_stream_go_proto",
         "//third_party/buildbuddy/proto:buildbuddy_go_proto",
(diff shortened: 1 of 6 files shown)


ccontavalli (approved):
Looks reasonable enough for what it is doing :/

PR #469, opened 2022-02-17
machinist:  remove atomic integer count

aaahrens:
Problem:

atomic integers pin a cpu if dmesg is flooded, this removes the atomic integer.  

This is technically a NoOp since we don't have any grafana metrics currently using these machinist metrics.

Tested:
Ran 
```
bazelisk run //machinist/cmd:cmd -- controlplane --port=8080 --domains=hello.world --bind-net=127.0.0.1
```
```
bazelisk run //machinist/cmd:cmd -- node poll --auth-server=https://auth.corp.enfabrica.net --control-port=8080 --control-host=127.0.0.1 --ips=99.99.99.99 --name=hello
```
Flooded Dmesg via iptables:
```
sudo iptables -A INPUT -j LOG --log-level 3
```
And watched process usage.

### DIFF
--- a/machinist/polling/metrics.go
+++ b/machinist/polling/metrics.go
@@ -24,6 +24,11 @@ var (
 		Name: "machinist_keepalive_fail",
 		Help: "The amount of times keepalive has failed",
 	})
+	dmesgErrors = promauto.NewGauge(prometheus.GaugeOpts{
+		Name:      "dmesg_errors",
+		Namespace: "machinist",
+		Help:      "Logs from dmesg",
+	})
 )
 
 // SendMetricsRequest polls the controlplane for metrics as well as spin up prometheus' node exporter.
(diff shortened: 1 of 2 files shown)


machinist/polling/metrics.go
```diff
@@ -43,14 +41,6 @@ func SendMetricsRequest(ctx context.Context, c *config.Node) error {
 		}, func() float64 {
 			return numErr
 		})
-		averageErr := atomic.NewFloat64(0.0)
-		promauto.NewGaugeFunc(prometheus.GaugeOpts{
-			Name:      "dmesg_errors_per_second",
```
minor-fixes, line 48: This is kinda goofy - usually one exports a counter, and does the rate via prometheus queries.

How about replacing this with a simple counter, and incrementing the counter inline below? Prometheus counters should be threadsafe with no other locking/atomics. Would that resolve the performance issue while preserving the metric?

minor-fixes (approved):

machinist/polling/metrics.go
aaahrens: You're right is goofy, I made it look like the other ones. Still going to nix the average err metric because prometheus should be calculating that.

PR #472, opened 2022-02-17
enkit: Import gRPC to fix py_proto_library issue

minor-fixes:
This change imports gRPC to fix an issue where `py_proto_library`
targets work in workspaces that import enkit but not in enkit
standalone.

The crux of the issue is that other workspaces rely on gRPC's workspace
rules to initialize various binds, including one that maps local Python
headers to `//external:python_headers`. Since enkit doesn't import grpc,
it doesn't make this bind, which breaks the build graph.

This change moves the gRPC import configuration from
`enfabrica/internal` to this repository; once we update the enkit
reference in internal, the grpc initialization code there can be
dropped.

Tested: `bazel query 'deps(//...)'` was broken with #471 but is working
with #471 and this change.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -26,6 +26,27 @@ enkit_init()
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
+    name = "com_github_grpc_grpc",
+    patch_args = ["-p1"],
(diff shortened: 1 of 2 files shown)


gpaussaenf (approved):
Thanks for noticing this and making the fix.

PR #473, opened 2022-02-17
gee docs

jonathan-enf:
- first cut at generating documentation
- added style sheet


scripts/gee
```diff
@@ -124,6 +126,65 @@ if [[ -n "${DEBUG}" ]]; then
   set -x
 fi
 
+function _markdown_to_html() {
```
ccontavalli, line 129: This is brave. I would have personally relied on some external tool, just supported one format.
Anything from the `man` suite (eg, similar to what git does internally to render man pages), which then has tools to convert to pretty much any format.

A while back, I used to add "perldoc" style documentation to scripts, trivially simple, and relied on perldoc on the system (which was everywhere back then). For example:
https://github.com/ccontavalli/sys-scripts/blob/master/ac-system-boot

(perldoc path to script to see the man page)

That said, looks good, approving.

ccontavalli (approved):

PR #475, opened 2022-02-19
bestie: Support text.xml file for summary metrics

gpaussaenf:
Added support for reading individual test case pass/fail/skip
summary results from the output test.xml file generated by
a 'bazel test' command. Supports both fully structured XML
output produced by the --junitxml flag (pytest) and unstructured
XML, where test case results are scraped from the <system-out>
console output text.

Refactored the test_result.go file to process both the text.xml
and outputs.zip files, the latter still used for test result
data metrics contained in *.metrics.pb files within.

Tested: Ran local instance of BES Endpoint using sample metrics
and //systest/src/benchmark and //systest/src/regression tests.
Forced some test cases to fail and to be skipped to verify handling
of these values in the test.xml file.

Jira: INFRA-594


msgoldflam (approved):
I don't know Go, so I will not catch every possible error but the logic looks good throughout for the XML path so I am approving.

bestie/server/test_result.go
```diff
@@ -56,6 +62,34 @@ func identifyStream(bazelBuildEvent bes.BuildEvent, streamId *build.StreamId) *b
 	return &stream
 }
 
+// Open an output file for reading.
+func openOutputFile(fileName, fileUri string) (io.ReadCloser, error) {
+	urlParts := strings.Split(fileUri, "://")
```
minor-fixes, line 67: Look at url.Parse: https://gobyexample.com/url-parsing

bestie/server/test_result.go
```diff
@@ -105,58 +139,50 @@ func handleTestResultEvent(bazelBuildEvent bes.BuildEvent, streamId *build.Strea
...
+			continue
+		}
+		defer fileCloser.Close()
 
-	// Process test metrics output file(s).
-	// Each output file contains a single (potentially large) protobuf message.
-	// For now, it's up to the client to split large metric datasets into multiple .metrics.pb files.
-	if err := processZip(stream, zipFileCloser); err != nil {
-		return fmt.Errorf("Error processing %s file: %w", outFileName, err)
+		// Found an output file to process. It is one whose file suffix
```
minor-fixes, line 161: I think the above would be better off as a switch or if/else ladder like:

```
switch {
case strings.HasSuffix(fname, outputsZipFileName):
  // open file
  // pass to processZipMetrics
case strings.HasSuffix(fname, testXmlSuffix):
  // open file
  // pass to process xml
default:
  continue
}
```

... so that it's more clear where to drop in additional handling for more file types

bestie/server/test_result.go
```diff
@@ -172,30 +198,38 @@ func processZip(stream *bazelStream, zipFile io.Reader) error {
...
 		if err != nil {
 			errs = append(errs, fmt.Errorf("Error reading output file %q: %w", fileName, err))
 			continue
 		}
 		debugPrintf("Read output file to process: %s\n", fileName)
 
-		// Extract all metrics from the file contents.
-		if err := processMetricsProtobufFile(stream, pbCompressedData[:]); err != nil {
+		// Extract all metrics from the protobuf file contents.
+		pResult, err := getTestMetricsFromProtobufData(fileData[:])
```
minor-fixes, line 225: Don't need `[:]` here

bestie/server/xml_result.go
```diff
@@ -0,0 +1,279 @@
...
+	// Also look for a single line containing the SKIPPEDSESSIONFINISH: token designating
+	// a test case that contains a pytest.skip() call during its execution:
+	//
+	//   systest/trial/test_app.py::TestHello::test_app_metrics_file[test.metrics.pb] SKIPPEDSESSIONFINISH: ...
+	//
+	// Extract the full test case ID from the first entry and the passed/failed/skipped result
+	// from the second.
+	matches := make(map[string]string)
+	var key string
+	for _, line := range strings.Split(strings.TrimSuffix(ts.SystemOut, "\n"), "\n") {
```
minor-fixes, line 190: strings.TrimSpace instead of strings.TrimSuffix

bestie/server/test_result.go
gpaussaenf: Fixed. Using url.Parse() is a cleaner solution.

bestie/server/test_result.go
gpaussaenf: OK, changed code to a single switch statement.

bestie/server/xml_result.go
gpaussaenf: Fixed.

bestie/server/test_result.go
gpaussaenf: Fixed.

bestie/server/xml_result.go
```diff
@@ -197,11 +200,20 @@ func parseUnstructuredXml(ts *TestSuite) ([]*xmlResult, error) {
...
+		} else {
+			s := strings.Split(line, " ")
+			if len(s) == 2 {
+				if strings.Contains("FAILED PASSED SKIPPED", s[1]) {
+					// e.g.: systest/trial/test_app.py::TestHello::test_app_metrics_nofile PASSED
+					key = strings.Split(s[0], "[")[0]
+					matches[key] = strings.ToLower(s[1])
+					key = ""
+				}
+			}
```
gpaussaenf, line 216: I found another format being used when --junitxml was not specified, meaning unstructured console output needed to be parsed manually. Pytest printed a summary of each test case and its result, one per line. This is what I've referred to as an abbreviated format, previously seen only with skipped test cases. Turns out, in some situations, this is the only indication of failed/passed/skipped test case results.

minor-fixes: > when --junitxml was not specified

Let's just explicitly not support this case. This happens when the test binary doesn't write a test.xml, and so bazel takes  the test's stdout and writes its own basic test.xml. I really don't want to be in the business of parsing this - it can't possibly be correct for all test frameworks. Let's just make sure the test frameworks emit structured data if its intended to be parsed.

minor-fixes (commented):
I think removing the parsing of stdout simplifies things greatly and we won't miss it, or am I missing something?

bestie/server/xml_result.go
gpaussaenf: I agree that it's going to be an ongoing pain to parse unstructured console output. My intent was to be able to produce pass/fail metrics for all test apps, even ones that do not have the ability to create the junitxml structured output. It'll be simpler (and less error-prone) to just not support unstructured XML. I'll probably archive this offline for future reference, in case the hardware-related tests cannot provide structured XML results.

gpaussaenf: Removed support for an unstructured test.xml file from the latest commit. Now incrementing a service counter and outputting a debug log message (when --debug flag is specified on the command line) when such .xml files are seen.

gpaussaenf:
Agreed, we shouldn't need to support unstructured test.xml files.

PR #477, opened 2022-02-21
faketree: support mount options.

ccontavalli:
Background:
So far, faketree only supported bind mounts, eg, mounting one directory
on top of another.

This PR:
- implements a TODO in code, completing support for mounting arbitrary
  directories.

This is partially in preparation for the next PR, where more of this
feature will be needed, and is a step toward fixing INFRA-256 and 
INFRA-535.

Tested: added unit tests, passing.


minor-fixes (approved):
Jira in PR desc?

PR #481, opened 2022-02-23
faketree: change how the pid namespace is handled.

ccontavalli:
Before this PR:

The PID namespace was created *LAST*, just before execing the
desired command. This caused "a few problems":

1) The spawned command was effectively PID 1 in the namespace.
   PID 1 is always a bit of an oddball, as the kernel treats it
   differently from other processes. See "man pid_namespaces"
   for details.

   Specifically, if PID 1 dies all its children are SIGKILLed
   automatically by the kernel, which in turn means that child
   management has to be done correctly in the spawned command
   (makes next PR not working/impossible to implement).

2) If --mount of /proc was specified, the mount would either
   not work, or point to /proc outside of the namespace.

   Mounting it correctly *required* doing it after the PID
   namespace is mounted. Further, golang libraries fail
   to properly create a uid and gid namespace unless
   /proc is re-mounted (see comment in code).

Goal of this PR is to make sure a) faketree remains the PID#1,
b) /proc is mounted correctly. Both a) and b) can only be done
together.

Additionally:
- clarified help string and messages, now that the API has settled
  a little bit, and mounts are working as expected.

Tested:
- unit tests are still passing - note that the different PID
  mechanics was causing most tests to fail at first attempt
  (no correct /proc handling).
- added a few trivial tests.

This is another step toward a workaround for INFRA-256 and INFRA-535.

### DIFF
--- a/faketree/faketree.go
+++ b/faketree/faketree.go
@@ -127,6 +127,14 @@ func (mo MountOptions) Parse(options string) (uintptr, string, string, error) {
 	return fsflags, fstype, fsdata, multierror.New(errs)
 }
 
+func (mo MountOptions) List() []string {
+	list := []string{}
+	for _, option := range mo {
+		list = append(list, option.Name)
+	}
+	return list
+}
+
 var KnownOptions = MountOptions{
 	{"dirsync", syscall.MS_DIRSYNC},
 	{"mandlock", syscall.MS_MANDLOCK},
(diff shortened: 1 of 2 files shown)


minor-fixes (approved):

minor-fixes:
PR desc needs Jira ticket reference

PR #482, opened 2022-02-23
faketree: wait for all child processes to die before returning.

ccontavalli:
Before this PR:
faketree would wait for its direct child to die, and then exit with its exit
status. If this child process spawned other children, it would not wait for
them to complete.

This is a problem when the parent, for example, is sending SIGTERM to the
entire process group - terminating the direct children possibly sooner than
indirect children, and returning/completing status before all are killed.

In this PR:
Add code so faketree _waits_ for its own child, and any indirect child
spawned within the namespace.

This is protected around the --wait= option, turned on by default, but
can easily be disabled with --wait=false.

Tested: simple test added in faketree_test.sh. Manually tested with
commands like:
	faketree -- sh -c "(sleep 5 & (sleep 2 & sleep 5) & sleep 7) &"

To verify that all commands are waited for.

### DIFF
--- a/faketree/faketree.go
+++ b/faketree/faketree.go
@@ -47,6 +47,81 @@ func (mf *MountFlags) Normalize() (*MountFlags, error) {
 	return &retval, nil
 }
 
+// ExitStatus wraps an exit code into an error.
+//
+// Workaround as there is no accessible constructor to create an ExitError.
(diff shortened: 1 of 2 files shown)


minor-fixes:
Jira ticket ref?

PR #483, opened 2022-02-23
faketree: handle signals in a different way.

ccontavalli:
Background:

What do we do when faketree receives SIGTERM? or SIGKILL? or one
of its children does? So far, faketree has had the default behavior,
whatever it was.

Problems:
- when invoked from within bazel, test-setup.sh creates a process
  group, and then sends SIGTERM to _all_ processes in the process
  group to handle timeout.

  But as soon as the first direct child terminates, it sends SIGKILL to _all_
  processes, effectively terminating them immediately.

  With no signal handling in faketree, SIGTERM causes faketree to
  exit immediately, effectively triggering the SIGKILL.

  This pgroup handling is also imprecise: if a child process creates its own
  process group, it won't get signaled at all.

- when invoked from buildbarn - instead - an action is terminated by sending a
  signal to the direct child only, no pgroup that I could see.

  With no signal handling in faketree, this would cause the opposite
  problem (outside of test? thanks to test-setup.sh? TBD): faketree
  terminates immediately, leaving zombie and orphan processes around
  continuing their work.

In this PR:
- when faketree receives a signal, it 1) ignores the default action of
  the signal, 2) propagates the signal down.

- when the faketree parent effectively dies, it commits karakiri by sending
  SIGTERM or SIGKILL to itself, regardless of the job management system.

Thanks to the use of a PID namespace, and thanks to the previous PR
waiting for _all_ children to terminate, it guarantees that all children
are dead before returning (even with -9, as the namespace is torn down,
and the kernel will kill all processes in it).

This mitigates the problems above, but leaves a few behind:
- signals are multiplied - there is no way to tell a signal sent
  to faketree from a signal sent to the entire group. So by "propagating"
  the signal, when job control uses process groups, effectively the
  signal is sent multiple times (and this is very visible from strace).
- kill -9 of the topmost process will always result in immediate kill
  -9 of the entire namespace.
- if one of the child processes does not terminate on its own,
  faketree will sit there waiting, until an external process sends
  SIGKILL.

A better approach would be to:
- create a new process group from faketree, right after "the first fork",
  so that signals to the process group are not propagated.
- have a bit of code in the "first child" so if the parent dies, it
  tries to do a clean TERM/KILL.
- have a timeout before issuing SIGKILL internally.

For 1 and 2, however, creating a pgrp requires setting up controlling
terminals correctly which is non-trivial, and probably more error
prone than the benefit.

3 May come in a later/separate PR, and should be very easy to implement.

Tested: see faketree_test.sh, also deployed in /tmp/ on a hardware
machine, edited defs.bzl files to point to the new faketree, observed
with the tools described in INFRA-535.

### DIFF
--- a/faketree/faketree.go
+++ b/faketree/faketree.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"os/signal"
 	"os/user"
 	"syscall"
 
@@ -302,14 +303,15 @@ func (mf *MountFlags) MakeTarget(perms os.FileMode) error {
 }
 
 type Flags struct {
-	Fail     bool
-	Root     bool
-	Hostname string
-	Chdir    string
-	Faketree string
-	Perms    uint32
-	Proc     bool
-	Wait     bool
+	Fail      bool
+	Root      bool
+	Hostname  string
+	Chdir     string
+	Faketree  string
+	Perms     uint32
+	Proc      bool
+	Wait      bool
+	Propagate bool
 
 	Uid, Gid int
 	Mount    []MountFlags
@@ -349,6 +351,9 @@ func (opts *Flags) Args() []string {
 	}
 	if !opts.Wait {
 		args = append(args, "--wait=false")
+	}
+	if !opts.Propagate {
+		args = append(args, "--propagate=false")
 	}
 
 	for _, mount := range opts.Mount {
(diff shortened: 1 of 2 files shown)


minor-fixes (approved):

minor-fixes:
PR desc needs Jira reference

ccontavalli:
> PR desc needs Jira reference

this is the only PR that had it :D

PR #485, opened 2022-02-23
faketree: add defense in depth for --wait.

ccontavalli:
Background:
when --wait=false, faketree waits for the command it was asked to run
to complete before returning.

when --wait=true, faketree waits for the command it was asked to run
AND ANY OF ITS CHILDREN to complete before returning.

Problem:
This means that if some of the children never return, faketree will
also never return. This is believed to be OK and actually strongly
desired in the job management systems we use: bazel local will
eventually timeout and send SIGKILL (but not SIGKILL immediately! which is a
problem), while the remote buildbarn cluster will likely do the same.

However, this --wait behavior is still "different" from what one would expect
from a traditional unix command. I'm worried that faketree instances may be
left running forever on the assumption that its direct child terminated/
was terminated.

In this PR:

Add a --timeout option, default it to a long 5 minutes timeout.

Note that this timeout does NOT limit how long a process run, neither
it limits how long to wait before SIGKILL after a SIGTERM. Those
are still left to the job management system.

Rather, it limits HOW LONG TO WAIT AT MOST for OTHER children (eg, not
the one faketree spawned directly) at most, in case those other
children are holding the faketree termination.


faketree/faketree.go
```diff
@@ -476,6 +504,9 @@ func (opts *Flags) Parse(argv []string) ([]string, error) {
 			"specify non standard options. Do so at your own risk, as faketree may no longer work.")
 	fs.BoolVar(&opts.Wait, "wait", opts.Wait, "Wait for all children of this process to die before returning.")
 	fs.BoolVar(&opts.Propagate, "propagate", opts.Propagate, "Take control of signal propagation - see help screen for more details.")
+	fs.DurationVar(&opts.Timeout, "timeout", opts.Timeout,
```
minor-fixes, line 507: Love the idea - recommend naming the flag more specifically than `timeout`, to address the confusion that could arise alluded to in your PR description. Maybe `termination-wait-timeout`?

minor-fixes (approved):

ccontavalli:
The presubmit failed due to the "can't astore download from presubmit", PR looks otherwise good, no change process wise, assuming I have not introduce any regression. Ping?

PR #486, opened 2022-02-24
go deps: Add prometheus_remote_storage_adapter

minor-fixes:
This change adds
https://github.com/KohlsTechnology/prometheus_bigquery_remote_storage_adapter
and its dependencies to the Go repositories fetched by bazel so that we
can build that project from source.

The binary will be included in a container we build and deploy on Cloud
Run; we can't use third-party prebuilt containers because they don't
respect the `$PORT` environment variable that Cloud Run sets, and its
infeasible to patch these containers to do so.

Dependency updates caused conflicts between libraries built from pre-generated
proto code and normally-generated proto code for the google APIs; to fix this:
* rules_go is updated to v0.30 to pick up a version of `@go_googleapis` that is new
  enough
* bazel is updated to 4.2.1 because rules_go requires it
* gazelle build directives are added to some google packages to get things to build

Tested:

```
bazel build @com_github_kohlstechnology_prometheus_bigquery_remote_storage_adapter//:prometheus_bigquery_remote_storage_adapter
bazel test //...
```

succeeds.

Jira: INFRA-609


ccontavalli (approved):
This approach look reasonable, but have you considered instead layering a new image on top of theirs?

This would be very simple to do:
1. Pull in their docker image from WORKSPACE file.
2. Add a create_image rule using the image you just added as base = "..."
3. Overwrite the entrypoint with a .sh script that reads the cloudrun environment variable, exports the PORT/whatever variable the container needs, invokes the original entrypoint.

This .sh script could look something like:
```
#!/bin/bash
export PORT=$CLOUDRUN_PORT
exec /path/to/old/entrypoint.sh
```

gpaussaenf (approved):

minor-fixes:
> This approach look reasonable, but have you considered instead layering a new image on top of theirs?
> 
> This would be very simple to do:
> 
> 1. Pull in their docker image from WORKSPACE file.
> 2. Add a create_image rule using the image you just added as base = "..."
> 3. Overwrite the entrypoint with a .sh script that reads the cloudrun environment variable, exports the PORT/whatever variable the container needs, invokes the original entrypoint.
> 
> This .sh script could look something like:
> 
> ```
> #!/bin/bash
> export PORT=$CLOUDRUN_PORT
> exec /path/to/old/entrypoint.sh
> ```

I did try this (for way too long); the original container is pretty barebones and doesn't seem to be able to run scripts, or even resolve environment variables set in entrypoint.

PR #489, opened 2022-03-03
infra: create buildbuddy shim

aaahrens:
This is a shim designed to be routed at L7, and captures any /file/download url and rewrites it to have the bytestream url be the intra-cluster url name of the storage section of buildbarn. It then forwards to any url of choice (usually the existing buildbuddy service)

As far as integration tests go, this is installed on staging currently. Seems to work nicely. 

jira: INFRA-683


gpaussaenf (approved):
Nice!

PR #492, opened 2022-03-05
bazel: Update to 5.1.0

minor-fixes:
This change updates the bazel version used to 5.1.0.

Tested: `bazel build //...; bazel test //...` passes locally

Jira: INFRA-572


minor-fixes:
Can't move this from draft until internal is confirmed to work with bazel 5.0, since merging this would block us from updating enkit in internal until bazel is updated in internal.

pete-enf (approved):

PR #495, opened 2022-03-11
affected targets: Pass if build graph is fixed between start and end

minor-fixes:
Presubmit failing with PRs that fix a build graph breakage are
confusing and cause delays in landing the fix PRs. At the same time,
there is no sensible set of affected targets to build if the build graph
cannot be queried at both points, which it can't if it was broken at one
of them.

This PR adds a special-case for runs where the build graph was broken at
the start commit (as determined by the entire query failing for any
reason) and fixed at the end commit - the library returns no changed
targets and no error. This should cause the presubmit to pass with a
no-op build in this case.

This is still not a very good presubmit because it doesn't build
anything, and a warning is emitted to this effect. However, it will at
least reduce confusion with fix PRs.

Tested: not yet

Jira: INFRA-174

### DIFF
--- a/lib/bazel/affected_targets.go
+++ b/lib/bazel/affected_targets.go
@@ -86,13 +86,16 @@ func GetAffectedTargets(start string, end string, config *ppb.PresubmitConfig, l
 	// Get all target info for both VCS time points.
 	var startResults *QueryResult
 	var endResults *QueryResult
+	var startQueryErr error
+	var endQueryErr error
 	errs := goroutine.WaitAll(
 		func() error {
 			log.Infof("Querying dependency graph for 'before' workspace...")
 			var err error
 			startResults, err = startWorkspace.Query("deps(//...)", WithUnorderedOutput(), workspaceLogStart)
 			if err != nil {
-				return fmt.Errorf("failed to query deps for start point: %w", err)
+				startQueryErr = fmt.Errorf("failed to query deps for start point: %w", err)
+				return startQueryErr
 			}
 			log.Infof("Queried info for %d targets from 'before' workspace", len(startResults.Targets))
 			return nil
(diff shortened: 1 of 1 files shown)


lib/bazel/affected_targets.go
```diff
@@ -102,13 +105,24 @@ func GetAffectedTargets(start string, end string, config *ppb.PresubmitConfig, l
...
 	)
 	if errs != nil {
+		if startQueryErr != nil && endQueryErr == nil {
+			// We are calculating targets over a change that fixes the build graph
+			// (broken before, working after). In a presubmit context, we want this
+			// step to succeed, but there is no sensible list of targets that the
+			// change affects since the build graph was broken in one stage.
+			//
+			// Pass here but emit a warning.
+			log.Warnf("Broken build graph detected at start point; this change fixes the build graph, but no targets will be tested. This change must be tested manually.")
```
ccontavalli, line 123: I think this is a good improvement, let's merge it in as is so that we have a fix for this problem.

However, my recommendation here would be:
1) Don't emit a warning, rather return a specific error type - so that the caller can tell that the problem was that the source graph could not be built, but the dest graph could.
2) Have the caller decide if the error is worth dying for, or just emit warning. Personally, I would a) emit warning, BUT b) exit with a specific status code, eg, 120.

The tool invoking the diff can then see that the problem is the graph being restored, and implement a different strategy to test the PR. For example: just look at the files modified, and run the tests corresponding to the modified files. Or ignore the problem for now.

Reason: I'm worried about rollbacks or PRs fixing the graph, but also touching other files. Again, this is a relatively little gain compared to the overall gain of this PR, so feel free to merge as is (or maybe just have the different error type? let the caller decide).

ccontavalli (approved):

PR #499, opened 2022-03-18
enproxy: add basic prometheus metrics and plumbing.

ccontavalli:
Background on some design decisions:
1) a separate http server is started on a different port just
   to export metrics. This simplifies authentication, https, and
   simplifies the main mux used for proxying and ssh.

2) some objects that need to export metrics are instantiated
   multiple times. This makes it hard to use the simple/default
   globals.

3) some counters already exist in the code, and are used
   internally for the engine to work.

4) some counters are in really hot paths, eg, per read/write
   performed.

The design uses simple per-object counters, paired with
an object implementing the prometheus.Collector() interface
just like recommended in the docs when reading "legacy counters",
counters that already exist in code.

This also allows multiple instances of objects by using
things like prometheus.WrapRegistererWithPrefix().

### DIFF
--- a/proxy/enproxy/enproxy.go
+++ b/proxy/enproxy/enproxy.go
@@ -106,9 +106,9 @@ func (config *Config) Parse() (utils.PatternList, Warnings, error) {
 
 // Flags represents command line flags necessary to define a proxy.
 type Flags struct {
-	Http  *khttp.Flags
-	Oauth *oauth.RedirectorFlags
-	Nassh *nasshp.Flags
+	Http       *khttp.Flags
+	Oauth      *oauth.RedirectorFlags
+	Nassh      *nasshp.Flags
 	Prometheus *khttp.Flags
 
 	ConfigContent          []byte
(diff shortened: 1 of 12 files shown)


ccontavalli:
I have a couple more counters that I'd like to add, but they require changes that are more involved.
So will send out as a separate PR. This PR has session / thread counters, next PR will have bandwidth, and some info on idle/vs active sessions.

proxy/enproxy/enproxy.go
```diff
@@ -280,13 +322,18 @@ type Enproxy struct {
 
 	mux     http.Handler
 	domains []string
-	starter Starter
+
+	register prometheus.Registerer
+        gatherer prometheus.Gatherer
```
minor-fixes, line 327: nit: gofmt/fix_format

proxy/nasshp/counters.go
```diff
@@ -0,0 +1,358 @@
+package nasshp
+
+import (
+	"github.com/enfabrica/enkit/proxy/utils"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type AllowErrors struct { // DONE
```
minor-fixes, line 8: Can // DONE comments be removed?

minor-fixes (approved):
Presubmit fails probably due to test that can't run in Cloud Build - can you confirm that the affected targets still pass locally before merging?

proxy/enproxy/enproxy.go
ccontavalli: done, could have sworn I had run it.

proxy/nasshp/counters.go
ccontavalli: good catch, done.

ccontavalli:
> Presubmit fails probably due to test that can't run in Cloud Build - can you confirm that the affected targets still pass locally before merging?

yes, all tests passing locally. yes, it's the failing dns test.

PR #502, opened 2022-03-23
machinist: return NXDOMAIN if there are no records for a query.

ccontavalli:
Background:
Without NXDOMAIN, the query is marked successful even without results.
This can cause some clients to stop querying other dns names (when
implementing a resolv.conf 'search' walk, for example).

Fix is very simple:
- return NXDOMAIN status if there are no results.


ccontavalli:
@aaahrens I believe this is the fix, as you had suggested over chat as well (thanks!). Not sure how to test, though.

minor-fixes (approved):

PR #504, opened 2022-03-23
khttp: fix server.go to be more reasonable.

ccontavalli:
Background:
server.go was created out of the proxy in a hurry, turns out it's more
broken than I'd like to admit. In this specific case:
- flags were not prefixed properly
- description were not updated
- semantics were not intuitive

Last point bit me last night: I was assuming the port number to be
a default if not specified in the address, turns out that if an address
is specified at all, the port number is completely ignored.

In this PR:
- fix the prefix
- fix the description
- have the port number behave as a true default port number

Additionally, added small unit test, and go fmt.

### DIFF
--- a/lib/khttp/BUILD.bazel
+++ b/lib/khttp/BUILD.bazel
@@ -20,7 +20,10 @@ go_library(
 
 go_test(
     name = "go_default_test",
-    srcs = ["utils_test.go"],
+    srcs = [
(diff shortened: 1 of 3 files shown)


ccontavalli:
This is the DNS test failing, fwiw, code is otherwise working.

minor-fixes (approved):

PR #508, opened 2022-03-24
enkit: Move command building to library

minor-fixes:
This change moves the command building that happens in `main.go` to a
separate library that returns a struct containing all the pieces with
convenience methods to run the command.

This facilitates writing a binary that emits markdown documentation from
enkit using a built-in cobra library. This documentation can serve as a
reference for users of enkit.

Tested: `bazel run //enkit/doc_gen -- --out-dir=/tmp/docs` works; `bazel
run //enkit -- login` still works.

Jira: INFRA-494

### DIFF
--- a/bazel/dependencies/BUILD.bazel
+++ b/bazel/dependencies/BUILD.bazel
@@ -1,4 +1,4 @@
 exports_files(
     ["*.patch"],
-    visibility=["//visibility:public"],
+    visibility = ["//visibility:public"],
 )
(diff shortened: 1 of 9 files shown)


minor-fixes:
This PR can generate docs from the commandline, which will mean we don't need to keep e.g. an out-of-band google doc up-to-date. Where should we put the generated markdown? Checked into enkit or internal to start?

minor-fixes:
Will check into internal, in case there are flag defaults that shouldn't be made public.

jonathan-enf (approved):

PR #513, opened 2022-03-29
BES Endpoint: Expand test.xml search for "realm" tag

gpaussaenf:
To support multiple test client environments, such as pytest and
gtest, the location of a "realm" tag may not always appear in the
test case properties. This change looks in the attributes of the
<testsuite> and <testsuites> elements of the test.xml file, in
that order, as additional sources of the "realm" tag for the
'testresult' metrics.

Tested: Used variations of `bazel test //model/functional/sfa-model:model-main-test`
to place the "realm" tag in different locations within the test.xml
file. Verified by looking at the console log output of a local bestie
instance that bestie found the tags and correctly prioritized
the one to use in the 'testresult' metric sent to BigQuery.

Jira: INFRA-826

### DIFF
--- a/bestie/server/xml_result.go
+++ b/bestie/server/xml_result.go
@@ -23,6 +23,7 @@ type xmlResult struct {
 // Unmarshalling structs for interpreting XML test results.
 type TestSuites struct {
 	XMLName    xml.Name    `xml:"testsuites"`
+	Realm      string      `xml:"realm,attr"`
 	TestSuites []TestSuite `xml:"testsuite"`
 }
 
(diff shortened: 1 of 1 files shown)


minor-fixes:
Makes sense - I have two main concerns:

* We may want more common tags above and beyond just realm in the future - what is the path for adding such tags? Must we make a bestie change to modify the expected XML schema every time?
* Inheritance behavior can be surprising to users long-term - can we be more strict about where the realm tag (or other "common" tags) must appear in the schema for simplicity's sake? For instance, maybe there's a python client lib the test uses that invokes `RecordProperty()` in the expecteed place, so that users don't have to worry about it?

gpaussaenf:
Closing this PR. Still a valid approach that could be reconsidered in the future, but for now found a way to modify the FE C++ gtests to insert the realm tag using a common fixture. Pytest clients use a conftest hook to insert the realm tag in the test case results.

PR #521, opened 2022-03-31
astore: Remove tag in query when querying by UID

minor-fixes:
This change removes the implicit `latest` tag added to the Datastore query when querying by UID and not tags. This change also makes it an error to query for a package without some sort of version specifier (latest tag or UID).

When querying by only UID, Datastore throws an index error when also sorting by Created. It's not clear why this is (the corresponding index appears to exist) so as a workaround, the ordering is removed in some cases. Since the goal of this query is exactly one version, ordering makes no difference anyway.

The affected queries can be viewed in the unit tests, and corresponding GQL was run manually against Datastore to ensure that there were no other index issues. Each example GQL query is pasted in the testcase as a comment.

Jira: INFRA-830

### DIFF
--- a/astore/server/astore/retrieve.go
+++ b/astore/server/astore/retrieve.go
@@ -65,6 +65,9 @@ func (s *Server) Retrieve(ctx context.Context, req *astore.RetrieveRequest) (*as
 	if req.Uid == "" && req.Path == "" {
 		return nil, status.Errorf(codes.InvalidArgument, "invalid request - no uid and no path")
 	}
+	if req.Uid == "" && (req.Tag == nil || len(req.Tag.Tag) == 0) {
+		return nil, status.Errorf(codes.InvalidArgument, "one of [uid, tags] must be specified to choose a specific version")
+	}
 
 	reqarch := strings.TrimSpace(req.Architecture)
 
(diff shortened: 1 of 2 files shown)


astore/server/astore/retrieve.go
```diff
@@ -76,15 +79,14 @@ func (s *Server) Retrieve(ctx context.Context, req *astore.RetrieveRequest) (*as
 			return nil, status.Errorf(codes.InvalidArgument, "Invalid path - %s", err)
 		}
 	} else {
-		query = datastore.NewQuery(KindArtifact).Order("-Created")
+		query = datastore.NewQuery(KindArtifact)
```
ccontavalli, line 82: uhm .. doesn't this break the order in which artifacts are returned?

astore/server/astore/retrieve.go
```diff
@@ -76,15 +79,14 @@ func (s *Server) Retrieve(ctx context.Context, req *astore.RetrieveRequest) (*as
...
 	}
 	query = query.Limit(1)
 
 	if req.Uid != "" {
 		query = query.Filter("Uid = ", req.Uid)
 	}
-
-	tags := []string{"latest"}
```
ccontavalli, line 87: so, if "no list of tags" is specified at all today, it is assumed to be "latest"

if a list of tags is specified (and this list of tags can be empty?) then the list is used.

Shouldn't the client then specify an empty non-nil list of tags when uid is supplied?

(doesn't removing the lines here affect the default behavior?)

astore/server/astore/retrieve.go
minor-fixes: We expect exactly one artifact in a query with no LIMIT clause, so order must not matter?

astore/server/astore/retrieve.go
minor-fixes: I think well-behaved clients looking to download a package version are specifying the version either by UID or tag; failure to do so should be an error (otherwise people can download by path, rely on default behavior, never think about what version they should be getting, and get broken unexpectedly when `latest` is bad)

If we want to have defaults, I think defaults should be boosted as close to the user as possible (i.e. client can set tag default if appropriate) and the server should be as dumb as possible.

ccontavalli:
this can be closed, right? reopen if you disagree.

PR #522, opened 2022-04-01
astore client: Specify empty tag set when querying by ID

minor-fixes:
Currently, when querying by UID, the client sends a request with no tags
to the server, which the server interprets by defaulting a latest tag
before running the query. This causes the query to return no results
when the package with that UID is not marked `latest`.

With this change, the client will always send a request with an empty
`TagSet` when querying by UID, which the server interprets as "set no 
default". This allows queries for specific UIDs to proceed as expected.

Tested: `bazel run //enkit -- astore download --force-uid rqh8i6qspag3ynxdqdu4k33keymgmxkn --output=/tmp/test_ip.tgz --loglevel-console=debug --overwrite` now downloads the requested file instead of erroring

Jira: INFRA-830

### DIFF
--- a/astore/client/astore/astore.go
+++ b/astore/client/astore/astore.go
@@ -1,6 +1,7 @@
 package astore
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -11,13 +12,13 @@ import (
 	"regexp"
 	"strings"
 
-	"context"
-	"github.com/enfabrica/enkit/astore/rpc/astore"
+	apb "github.com/enfabrica/enkit/astore/rpc/astore"
 	"github.com/enfabrica/enkit/lib/client"
 	"github.com/enfabrica/enkit/lib/client/ccontext"
 	"github.com/enfabrica/enkit/lib/grpcwebclient"
 	"github.com/enfabrica/enkit/lib/kflags"
 	"github.com/enfabrica/enkit/lib/progress"
+
 	"github.com/go-git/go-git/v5"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -26,11 +27,11 @@ import (
 
 type Client struct {
 	conn   grpc.ClientConnInterface
-	client astore.AstoreClient
+	client apb.AstoreClient
 }
 
 func New(conn grpc.ClientConnInterface) *Client {
-	client := astore.NewAstoreClient(conn)
+	client := apb.NewAstoreClient(conn)
 	return &Client{conn: conn, client: client}
 }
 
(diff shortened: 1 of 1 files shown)


minor-fixes:
This doesn't actually work (gets farther, fails due to index error):

```
bminor-> bazel run //enkit -- astore download --force-uid rqh8i6qspag3ynxdqdu4k33keymgmxkn --output=/tmp/test_ip.tgz --loglevel-console=debug
INFO: Invocation ID: e2c0b6f3-febf-463b-9ca8-01877c613de0
INFO: Analyzed target //enkit:enkit (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //enkit:enkit up-to-date:
  bazel-bin/enkit/enkit_/enkit
INFO: Elapsed time: 0.155s, Critical Path: 0.02s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Running command line: bazel-bin/enkit/enkit_/enkit astore download --force-uid rqh8i6qspag3ynxdqdu4k33keymgmxkn '--output=/INFO: Build completed successfully, 1 total action
2022-04-01T15:41:08.771Z        DEBUG   Files to download: [{Remote:rqh8i6qspag3ynxdqdu4k33keymgmxkn RemoteType:uid Local:/tmp/test_ip.tgz Overwrite:false Architecture:[amd64-linux all] Tag:0xc000380340}]
2022-04-01T15:41:08.771Z        INFO    Loading credentials from store 'enkit'
2022-04-01T15:41:08.771Z        INFO    Using credentials of 'scott@enfabrica.net' for requested '(default)'
ERROR: Could not contact the metadata server. Is your connectivity working? Is the server up?
For debugging: rpc error: code = Internal desc = error running query - rpc error: code = FailedPrecondition desc = no matching index found. recommended index is:%0A- kind: Artifact%0A  properties:%0A  - name: Uid%0A  - name: Created%0A    direction: desc%0A
```

I think this is related to https://enfabrica.atlassian.net/browse/INFRA-830?focusedCommentId=15641 - either we remove ordering on this specific query, or fix the index issue somehow?

ccontavalli:
> I think this is related to https://enfabrica.atlassian.net/browse/INFRA-830?focusedCommentId=15641 - either we remove ordering on this specific query, or fix the index issue somehow?

I'd frankly be fine with either solution. Probably removing ordering is better when uid is specified, the constraint is useless in that case (uids are guaranteed unique), and we save on some index space.

Pushing astore is scary as we haven't done it in a while, but it's actually a good reason to do it and write playbook.

This PR looks frankly good, although would have been good to separate the rename, as it's hard to see what changed from a logical standpoint from what changed just because of renaming (no need to change now, will happily approve regardless).

minor-fixes:
Above issue fixed with astore deployment `20220401t212615`

ccontavalli (approved):

PR #527, opened 2022-04-03
enproxy: add automated session expiration.

ccontavalli:
Background:
When a client disconnects, enproxy does not close the session between
the proxy and the backend, so if the client comes back, the session
can be resumed.

This causes connections to build up over time, and pretty much leak
memory until the connections naturally die.

In this PR:
- add a few counters to keep track of orphaned connections, export
  them to prometheus for monitoring.
- use those counters to periodically and automatically expire sessions.

Additionally, a few support classes were needed (and added):
- atomic time, to keep track of time atomically.
- a mock clock implementation, to abstract time.After.

It's a large PR, it could probably have been broken in two,
but it's hard to do after the fact.

Tested: written unit tests, will need to stage the change.

### DIFF
--- a/proxy/nasshp/nassh.go
+++ b/proxy/nasshp/nassh.go
@@ -18,6 +18,7 @@ import (
 	"net/http"
 	"net/url"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -29,6 +30,18 @@ type sessions struct {
 	// Key is a string, the SID. Value is a readWriter pointer.
 	sids sync.Map
 	SessionCounters
+}
+
+func (s *sessions) Iterate(oneach func(rw *readWriter) bool) {
+	s.sids.Range(func(key, value interface{}) bool {
+		rw, converted := value.(*readWriter)
+		if !converted {
+			s.Invalid.Increment()
+			return true
+		}
+
+		return oneach(rw)
+	})
 }
 
 func (s *sessions) Get(sid string) *readWriter {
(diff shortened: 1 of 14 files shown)


proxy/nasshp/nassh.go
```diff
@@ -103,8 +118,135 @@ type options struct {
...
+				counters.ExpireOrphanClosed.Increment()
+			} else {
+				counters.ExpireRuthlessClosed.Increment()
+			}
+
+			counters.ExpireLifetimeTotal.Add(uint64(pausedfor / 1e9))
+			counters.ExpireYoungest.SetIfGreatest(uint64(pausedon))
+
+			rw.browser.Close(fmt.Errorf(
+				"session expired after %s of inactivity", time.Duration(pausedfor)*time.Nanosecond))
```
minor-fixes, line 242: time.Nanosecond is 1, so that can be dropped

proxy/nasshp/nassh.go
```diff
@@ -103,8 +118,135 @@ type options struct {
...
+			if left <= ep.RuthlessThreshold {
+				if pausedfor <= maxorphan {
+					break
+				}
+				counters.ExpireOrphanClosed.Increment()
+			} else {
+				counters.ExpireRuthlessClosed.Increment()
+			}
+
+			counters.ExpireLifetimeTotal.Add(uint64(pausedfor / 1e9))
```
minor-fixes, line 238: `uint64(pausedFor) / uint64(time.Second)` might be clearer?

minor-fixes (approved):

minor-fixes:
#531 to disable test failing the presubmit

proxy/nasshp/nassh.go
ccontavalli: done.

proxy/nasshp/nassh.go
ccontavalli: good suggestion, done.

PR #530, opened 2022-04-04
bazel/utils: update utils with internal changes, publish extra utils.

ccontavalli:
Background:
A few utilities have been duplicated across repositories. A few extra
utilities were added. I have a need to use a couple of those utilities
(mainly, merge_kwargs and some friends) in enkit. Rather than track
the full set of deps, I added most utilities in the enkit repository.

It would be good to have a few more utilities, but for future PRs.

I will follow up this PR with a cleanup of other repositories to
re-use the now published utilities.

### DIFF
--- a/bazel/utils/diff_test.bzl
+++ b/bazel/utils/diff_test.bzl
@@ -77,21 +77,28 @@ def _diff_test_impl(ctx):
         if [[ ! -e "${{exp}}" ]]; then echo "Missing file: ${{exp}}"; exit 1; fi
         diff -u "${{exp}}" "${{act}}" ; RC=$?
         while [ "${{1:-}}" != "" ]; do
-          if [[ ${{RC}} -ne 0 ]] && [[ "$1" == "--update_goldens" ]]; then
-            b="$(readlink -f "${{exp}}")"
-            echo "Updating ${{b}}"
-            cp -vf "${{act}}" "${{b}}"
+          if [[ "$1" == "--update_goldens" ]]; then
+            echo "--update_goldens specified."
+            if [[ ${{RC}} -ne 0 ]]; then
+              b="$(readlink -f "${{exp}}")"
+              echo "Updating ${{b}} from ${{act}}"
+              cp -vf "${{act}}" "${{b}}"
+              RC=0
+            else
+              echo "${{act}} is already up-to-date."
+            fi
           fi
           shift
         done
         if [[ ${{RC}} != 0 ]]; then
           echo "Error: Generated file did not match ${{exp}}."
           echo ""
-          echo "To automatically update your expected data files, run:"
+          echo "To automatically update your expected data file, run:"
           echo ""
-          echo "  ./tools/update_goldens //path/to:your-diff_test  # a target"
-          echo "  ./tools/update_goldens //path/to:all             # a dir"
-          echo "  ./tools/update_goldens //path/to/...             # a tree"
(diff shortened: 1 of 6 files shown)


bazel/utils/diff_test.bzl
```diff
@@ -1,17 +1,72 @@
...
+        ),
+    },
+    test = True,
+)
 
 def _diff_test_impl(ctx):
-    if len(ctx.files.actual) > 1:
-        fail("`actual` must specify a single file.")
-    actual = ctx.files.actual[0].short_path
+    if ctx.attr.output_within_actual:
```
jonathan-enf, line 64: I didn't integrate this functionality when I pushed diff_test public because it's a bit of a mess.

A more bazel-ic implementation would add an optional "deps" attribute to diff_test, that can be used to add a dependency with undeclared outputs.

ccontavalli: I'm personally more concerned about the divergence of the implementations than the cleanness of it :/
What you have works, and if there are improvements or fixes, ... at least we'll benefit from it everywhere?
I can withdraw the PR, or parts of it if you feel strongly, but I do think there's a benefit from sharing this code.

jonathan-enf: I hear you.  OK, let's check in as-is and I'll do a two-stage cleanup later.

bazel/utils/diff_test.bzl
```diff
@@ -22,28 +77,21 @@ def _diff_test_impl(ctx):
         if [[ ! -e "${{exp}}" ]]; then echo "Missing file: ${{exp}}"; exit 1; fi
         diff -u "${{exp}}" "${{act}}" ; RC=$?
         while [ "${{1:-}}" != "" ]; do
-          if [[ "$1" == "--update_goldens" ]]; then
```
jonathan-enf, line 25: Can we revert this change, and instead I'll push a PR that will backport this into internal?  Especially for external users, the improved error messages are useful.

bazel/utils/diff_test.bzl
```diff
@@ -61,32 +109,38 @@ def _diff_test_impl(ctx):
 diff_test = rule(
     doc = """
       A test that compares the contents of two files to ensure they are
-      identical.  Typically, this would be used to compare the output of a
```
jonathan-enf, line 64: the revised docstring is more generic and suitable for external use.  I'll port this back into internal if you don't change it.

bazel/utils/diff_test.bzl
```diff
@@ -61,32 +109,38 @@ def _diff_test_impl(ctx):
...
-      Alternately, consider using the provided `update_goldens` python script
-      as a quick way to identify and regenerate a large number of expected data
-      files at once.
     """,
     implementation = _diff_test_impl,
     attrs = {
         "expected": attr.label(
             doc = "A label indicating the file containing the expected data.",
             allow_files = True,
-            mandatory = True,
```
jonathan-enf, line 82: this mandatory is mandatory

jonathan-enf (approved):
OK, but please revert the improvements I flagged, I'll take care of porting them back to internal later.

minor-fixes (approved):

bazel/utils/diff_test.bzl
ccontavalli: done.

bazel/utils/diff_test.bzl
ccontavalli: makes sense, done.

PR #533, opened 2022-04-07
kunit: Fix BUILD files, imports

minor-fixes:
This change fixes Python import paths and BUILD files for the kunit
binary/library to make it Bazel compliant.

An extra target is also defined to build a Python hermetic zip; this zip
file is ran instead of the script, so that we don't have to worry as
much about runfiles issues. The script template is modified to pass the
zip to python3, as the zip is not itself executable.

Tested: `BAZEL_PROFILE=buildbarn bazel test
//driver/farmem:enf_rma_unit_test
--override_repository=enkit=/home/bminor/dev/enkit` with internal PR
1656 works

### DIFF
--- a/bazel/linux/kunit/BUILD.bazel
+++ b/bazel/linux/kunit/BUILD.bazel
@@ -1,12 +1,33 @@
 py_binary(
     name = "kunit",
     srcs = ["kunit.py"],
-    visibility = ["//visibility:public"],
     deps = [
-        ":kunit_config",
+        ":kunit_json",
         ":kunit_kernel",
         ":kunit_parser",
     ],
+)
+
+py_library(
+    name = "kunit_json",
+    srcs = ["kunit_json.py"],
+    deps = [
+        ":kunit_parser",
+    ],
+)
+
+py_library(
+    name = "kunit_kernel",
+    srcs = ["kunit_kernel.py"],
+    deps = [
+        ":kunit_config",
+        ":kunit_parser",
+    ],
+)
+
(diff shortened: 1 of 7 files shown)


bazel/linux/kunit/BUILD.bazel
```diff
@@ -1,25 +1,58 @@
...
+#    srcs = ["kunit_tool_test.py"],
+#    main = "kunit_tool_test.py",
+#    deps = [
+#        ":kunit_config",
+#        ":kunit_parser",
+#        ":kunit_kernel",
+#        ":kunit_json",
+#        ":kunit",
+#    ],
+#)
```
minor-fixes, line 51: @cbrune This test expects data in a `test_data` directory that doesn't seem to exist in the repo - do you have this testdata? Can we commit it so that the test can run?

bazel/linux/kunit/BUILD.bazel
```diff
@@ -1,25 +1,58 @@
 py_binary(
     name = "kunit",
     srcs = ["kunit.py"],
-    visibility = ["//visibility:public"],
     deps = [
-        ":kunit_config",
```
rajachan, line 6: D'oh, the move to v5.13 missed this.

rajachan (approved):

bazel/linux/kunit/BUILD.bazel
cbrune: I don't know anything about this test.

PR #534, opened 2022-04-07
kunit: Add unit test testdata

minor-fixes:
This change copies testdata that the unit test needs to run from
https://github.com/enfabrica/linux/tree/enf/impish-19.19/tools/testing/kunit/test_data,
and fixes the `py_test` rule to allow access to this data.

Tested: Unit test runs with `bazel test //bazel/linux/kunit/...`


ccontavalli (approved):

cbrune (commented):
LGTM

PR #535, opened 2022-04-08
postsubmit: Generate Python stub

minor-fixes:
This change adds a BUILD rule to generate Python code from
postsubmit.proto, in preparation for a Python conversion tool in
internal.


pete-enf (approved):
LGTM

PR #536, opened 2022-04-08
postsubmit: Add timeout field

minor-fixes:
This change adds a timeout field to the Build spec, which is the only
missing piece to onboard the FPGA nightly regression.

Tested: None

Jira: INFRA-876


pete-enf (approved):
SHIP IT!!!!!

PR #537, opened 2022-04-11
postsubmits: Add more necessary fields

minor-fixes:
This change adds fields to the Build message that are necessary in order
to migrate the two existing HW nightly regressions:
* `fpga-fullchip` passes a bunch of arguments to `bazel test` that we
  need to preserve
* `fpv-nightly` makes use of positive tags in
  `build_tag_filters`/`test_tag_filters`, and currently the proto
  message only handles negative tags.

Jira: INFRA-875


pete-enf (approved):
LGTM

PR #544, opened 2022-04-22
linux/defs.bzl: compile device drivers with -O1 for debug builds

george-enf:
When debugging, it is sometimes more convenient if compile optimizations
are disabled. It is not possible to compile kernel modules with -O0 but
-O1 seems to work. Also, disable function inlining.

Tested:

for i in "" "-c fastbuild" "-c opt" "-c dbg"; do
    echo -ne "$i: "
    bazel build driver/ib/enf_ib-swtest $i 2> /dev/null
    nm bazel-bin/driver/ib/enf-ubuntu-impish-minimal/enf_ib-swtest.ko | grep ' [Tt] ' | wc -l
done

: 229
-c fastbuild: 229
-c opt: 229
-c dbg: 243

### DIFF
--- a/bazel/linux/defs.bzl
+++ b/bazel/linux/defs.bzl
@@ -259,6 +259,16 @@ def _kernel_modules(ctx):
         modules = " ".join(modules),
     )
 
+    compilation_mode = ctx.var["COMPILATION_MODE"]
+    if compilation_mode == "fastbuild":
+        cflags = ""
+    elif compilation_mode == "opt":
+        cflags = ""
+    elif compilation_mode == "dbg":
+        cflags = "-O1 -fno-inline-functions-called-once"
+
+    extra += " BAZEL_CFLAGS+='%s'" % cflags
+
     ctx.actions.run_shell(
         mnemonic = "KernelBuild",
         progress_message = message,
--- a/bazel/linux/defs.bzl
+++ b/bazel/linux/defs.bzl
@@ -261,13 +261,17 @@ def _kernel_modules(ctx):
 
     compilation_mode = ctx.var["COMPILATION_MODE"]
     if compilation_mode == "fastbuild":
-        cflags = ""
+        cflags = "-g"
     elif compilation_mode == "opt":
         cflags = ""
     elif compilation_mode == "dbg":
-        cflags = "-O1 -fno-inline-functions-called-once"
+        cflags = "-g -O1 -fno-inline-functions-called-once"
+    else:
+        fail("compilation mode '{compilation_mode}' not supported".format(
+            compilation_mode=compilation_mode,
+        ))
 
-    extra += " BAZEL_CFLAGS+='%s'" % cflags
+    extra += " EXTRA_CFLAGS='%s'" % cflags
 
     ctx.actions.run_shell(
         mnemonic = "KernelBuild",


george-enf:
The current behavior is to compile the modules for the minimal flavor with `-Os` and for the generic flavor with `-O2`. I assume that this is by design. This PR overrides the optimization only for the debug builds.

bazel/linux/defs.bzl
```diff
@@ -259,6 +259,16 @@ def _kernel_modules(ctx):
         modules = " ".join(modules),
     )
 
+    compilation_mode = ctx.var["COMPILATION_MODE"]
+    if compilation_mode == "fastbuild":
+        cflags = ""
+    elif compilation_mode == "opt":
+        cflags = ""
+    elif compilation_mode == "dbg":
+        cflags = "-O1 -fno-inline-functions-called-once"
```
ccontavalli, line 268: we hard code -g in our Makefile, should -g be provided from here instead? Maybe something like:
"fastbuild" -> -g
"opt" -> no -g
"dbg" -> -ggdb3 or whatever works best, even if more expensive.

bazel/linux/defs.bzl
```diff
@@ -259,6 +259,16 @@ def _kernel_modules(ctx):
...
+
+    extra += " BAZEL_CFLAGS+='%s'" % cflags
```
ccontavalli, line 270: just make this CFLAGS?

bazel/linux/defs.bzl
```diff
@@ -259,6 +259,16 @@ def _kernel_modules(ctx):
...
     )
 
+    compilation_mode = ctx.var["COMPILATION_MODE"]
+    if compilation_mode == "fastbuild":
+        cflags = ""
+    elif compilation_mode == "opt":
+        cflags = ""
+    elif compilation_mode == "dbg":
+        cflags = "-O1 -fno-inline-functions-called-once"
+
```
ccontavalli, line 269: want to add an "else fail("compilation mode not supported?")"

bazel/linux/defs.bzl
```diff
@@ -259,6 +259,16 @@ def _kernel_modules(ctx):
         modules = " ".join(modules),
     )
 
+    compilation_mode = ctx.var["COMPILATION_MODE"]
```
ccontavalli, line 262: nothing to do: the hipster/modern way to do this would be to use //config like you were before, but I'm told the only way to do so from a rule... is to wrap it in a macro, something like:
```
_real = rule(
  attrs = {
    "opt_mode": attr.string(...)
  }..
)

def wrapper(...):
   _real(..., opt_mode = select({"//:is_opt": "opt", "//:is_fastbuild": "fastbuild", "//:is_dbg": "dbg"})
```

where //:is_X is a special target that  is used in select that allows to check if a config parameter is set.

(I'm fine with your current implementation, this is too much work)

bazel/linux/defs.bzl
george-enf: I tried your suggestion but it looks like that `-g` in our Makefile is already redundant. The kernel Makefile will add a `-g` depending on `CONFIG_DEBUG_INFO`: https://github.com/torvalds/linux/blob/master/scripts/Makefile.debug. I don't know if there is a way for a module to opt-out. David is not in this repository so I can't tag him for his input as well.

bazel/linux/defs.bzl
george-enf: Makes sense. Changed to `EXTRA_CFLAGS`.

Interesting trivia: make seems to be able to concatenate variables if asked to (with a +=) in the command line:
```
$ echo -e '$(info $(X))\nall:' | make -f - X=1 X+=2
1 2
```

bazel/linux/defs.bzl
george-enf: Added in case things change in the future. Bazel already enforces it:
```
ERROR: While parsing option -c george: Not a valid compilation mode: 'george' (should be fastbuild, dbg or opt)
```

bazel/linux/defs.bzl
george-enf: In any case, I added `-g` for fastbuild and dbg.

ccontavalli (approved):

PR #545, opened 2022-04-22
flextape: Raise and parameterize timeouts

minor-fixes:
All durations that the service uses are moved to a "server" config
section, so that they are no longer hard-coded. At the same time,
timeouts are greatly raised to reduce the number of expirations that
happen because the client is too slow to refresh an active allocation.

The client is modified to attempt a refresh earlier into the
server-specified duration, but given the raised timeouts its minimum
refresh duration is raised from 1s to 5s.

Jira: INFRA-607

### DIFF
--- a/flextape/proto/config.proto
+++ b/flextape/proto/config.proto
@@ -11,18 +11,18 @@ message Config {
   // One LicenseConfig for each vendor::feature tuple.
   // This should be configured from the license file the vendor provides.
   repeated LicenseConfig license_configs = 1;
+
+  ServerConfig server = 2;
 }
 
 // Default prioritizer. Licenses are allocated in the order they are requested.
-message FIFOPrioritizer {
-}
+message FIFOPrioritizer {}
 
 // Allocates licenses so that they are spread evenly across users.
 //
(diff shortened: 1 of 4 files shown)


ccontavalli (approved):

flextape/proto/config.proto
```diff
@@ -32,7 +32,34 @@ message LicenseConfig {
...
+
+  // Interval on which actions should refresh their allocation while the action
+  // is executing.
+  // Default: 30s
+  uint32 allocation_refresh_duration_seconds = 2;
+
+  // Interval on which to clean up expired/released allocations and queue
+  // entries, and promote queued entries to allocations.
+  // Default: 1s
+  uint32 janitor_interval_seconds = 3;
```
pete-enf, line 55: Why is the default 1 but we are going with 3 yet for other refresh rates we are lowering the time?

pete-enf (approved):

flextape/proto/config.proto
minor-fixes: 3 is the proto tag value, not the default value or actual value here. Each field in a proto message needs a unique number.

PR #556, opened 2022-04-30
linux/defs.bzl: refactor kernel modules support in a backward compatible way.

ccontavalli:
In this PR:

- introduce explicit support for different architectures.
  In short:

  1) a kernel image is now tagged with an architecture (example: "x86_64",
     "um". Magic architecture "host" implies whatever architecture running on the
     host). For a module to be linked with another module, or to run on an image, it
     has to be built for the same architecture.

  2) a kernel module is now tagged with multiple architectures.
     The rule will generate the commands to build for all architectures
     specified, but in a way that the corresponding commands will not be
     run unless the architecture is needed** (uses the same mechanism
     used by C++ library compiled dynamically/statically).

  3) sets the attributes to sane defaults, so that rules using the
     old ARCH="" just keep working.

  4) when running a kernel test, the image used automatically determines
     which module architecture is used.
  5) when linking multiple modules together, any dependent module has
     to be available for the same set of architectures as well.

- changing the above required a singificant refactor of how
  multiple builds of the same module are bundled together, and how
  dependencies are managed and expanded.

  The kernel modules rule now support a kdeps parameter to allow to
  express the dependency of one module to another. This is not a link
  time dependency, like deps, rather a run time dependency, implying
  that one module requires another module to be loaded at run time to
  work.

  Additionally: added a setup parameter, so for each module, a set
  of commands to load it can be explicitly specified, which will be
  carried over when running a test.

- changed the kernel test rule to benefit from the above changes
  (was really hard to do otherwise, given the requirement to load
  kernel modules in a specific order - especially across PRs):

  1) A kernel_test rule is now a macro (for backward compatibility)
     that a) creates a "runtime package": a set of files, and an
     init.sh script, and b) creates a "runner", an abstracted rule
     that allows to run a package in whatever environment.

     This opens the door to creating a vm_test rule, for example,
     or a baremetal_test rule, that given a "runtime package" can
     run it in either environment.

  2) Created a first runner, that runs a package under "uml", just
     like the old test rule.

  To allow for 1 and 2 split, there is now a stricter contract
  with the runner, that expects an init.sh in a specific directory.

- The uml runner was improved to not require a rootfs. It is expected
  that we will stop using a rootfs with those rules.

- Past PRs modified the version of the script in our rootfses
  to obey this new contract.

Tested:
- internal keeps working with --override_repository and no other change
  but the rootfs update (that is required!)
- a PR I have as draft for internal with the new features seems to
  work as expected (adding correct arch and archs parameters, removing
  the rootfs).
- I expect our internal presubmits to be able to catch regressions here
  as a merge PR is sent out automatically.

(I need to repeat some of the tests after minor cleanups before sending
the PR out)


rajachan (approved):
Lots of good enhancements for the kernel* rules. Thanks for taking this on!

PR #561, opened 2022-05-05
gee: disable autostash

jonathan-enf:
gee: disable autostash

Users are having trouble with autostash.  I think it's a better user experience
to simply disable autostash, and require users to commit their work before
rebasing (easier to recover when something goes awry).

Tested: Ran the `up`, `rup`, `update_all` commands in branches both
with and without uncommitted changes.


jonathan-enf:

ccontavalli (approved):

PR #565, opened 2022-05-06
gee pr_submit: disable post-submit diff check

jonathan-enf:
gee pr_submit: disable post-submit diff check

This check was low value (gh pr submit is reliable), and would occasionally
report spurious failures (scenarios where two PRs were submitted at around
the same time, and changed the same file, but neither change conflicted with
the other).  To avoid alarming the user with a useless error, this check
is now disabled.

Tested: Will use gee pr_submit to submit this PR.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -3628,14 +3628,19 @@ function gee__pr_submit() {
   rm "${BODYFILE}"
   _update_main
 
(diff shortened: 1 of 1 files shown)


jonathan-enf:

ccontavalli (approved):

PR #566, opened 2022-05-06
gee: streamline pr_make/pr_submit

jonathan-enf:
This PR addresses two annoyance-level issues:

1. The PR description template was unnecessarily verbose.  The new version is
   shorter and thus easier to edit.

2. Instead of asking the user if they want to edit the PR description, we just
   ask them how (previously, if they decided not to edit the PR description,
   they would win the booby prize of pr_make erroring out).  Additionally, if
   the user specifies the EDITOR environment variable, just use that editor.

3. The common use chase for `pr_submit` is "wait for all presubmits to pass
   and then submit".  Previously, after all presubmits passed, we would wait
   for user confirmation before submitting.  This PR removes that extra
   confirmation, so that pr_submit can be more "fire and forget" after
   waiting for presubmits to succeed.

4. Fixed malformed "git diff" command that was being invoked during PR
   description generation (harmless but cosmetic fix).

Tested: Used this version of gee to run pr_make, and will use it to run pr_submit.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -3322,31 +3322,14 @@ function gee__pr_make() {
   BODYFILE="$(mktemp -t prbody.XXXXXX.txt)"
   (
     cat <<'EndOfPrTemplate'
-# Pull request template
-#
-# This template is designed to help you write a good PR description.  This
-# description will be used as the first comment for your pull request, and will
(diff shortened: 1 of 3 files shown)


jonathan-enf:

ccontavalli (approved):

PR #567, opened 2022-05-06
bazel/linux: split uml test in kunit checker and runner.

ccontavalli:
Background:
Today running a kernel_test implies running uml, loading a bunch
of kunit tests, and checking with the kunit python helper.

This PR:

1) Adds support for a "runtime package" to have a "check", something
   that verifies that the runtime package worked correctly.

2) Breaks the uml test in two parts:

   a) A part that just runs the package in UML.

   b) A part that checks that the output represents a passing
      test from a kunit perspective.

3) Moves the kunit checking to be part of the "runtime package"
   generated by "kernel_test_dir".

This is in preparation for next PR: adding ability to run qemu
instead of uml to execute kunit tests.

Note that the rules are grossly missnamed at this point. The
goal is to send a cleanup PR as more of the transformation is
performed.

Tested:
bazel test --override_repository= ... on internal all pass.
Further, I manually broke one test, and the error was reported
correctly.


rajachan (approved):
Makes sense, LGTM.

PR #569, opened 2022-05-09
enkit ssh: Add --proxy-map arg

minor-fixes:
This change adds a `--proxy-map` argument to `enkit ssh`, allowing one
to map bits of a SSH target address to its corresponding proxy. This
will allow us to configure e.g.:

```
--proxy-map=.foo=https://foo.example.com,.bar=https://bar.example.com
```

in the enkit kconfig, so that a bare `enkit ssh user@machine-1.bar` will
resolve to the (potentially non-default) proxy https://bar.example.com.

This is done by matching all of the non-flag SSH args against all
entries in `--proxy-map` via substring lookup until a match is found; if
no match is found, the proxy named by `--proxy` is used as a fallback.

This logic is likely to work most of the time but fall on its face for
unusual invocations of `enkit ssh`, so proxy decisions are logged at at
least the INFO level.

Tested:
* Added unit tests
* Pushed https://github.com/enfabrica/internal/pull/4917 as staging config, tested machines
   in the three different areas listed

Jira: INFRA-868

### DIFF
--- a/proxy/ptunnel/commands/ssh.go
+++ b/proxy/ptunnel/commands/ssh.go
@@ -12,6 +12,11 @@ import (
 	"strings"
 )
 
+type proxyMapping struct {
+	substring string
+	proxy     string
+}
+
 type SSH struct {
 	*cobra.Command
 	*client.BaseFlags
@@ -22,7 +27,7 @@ type SSH struct {
 
 	Proxy            string
 	proxyList        []string
-	ProxyMap         map[string]string
+	ProxyMap         []proxyMapping
 	BufferSize       int
 	SSH              string
 	UseInternalAgent bool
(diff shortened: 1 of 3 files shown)


proxy/ptunnel/commands/ssh.go
```diff
@@ -135,18 +172,29 @@ func NewSSH(base *client.BaseFlags) *SSH {
...
 	root.RunE = root.Run
 
 	root.Command.Flags().IntVar(&root.BufferSize, "buffer-size", 0, "Default read and write buffer size for window management. If 0, leave the default used by the tunnel")
 	root.Command.Flags().StringVarP(&root.Proxy, "proxy", "p", "", "Full url of the proxy to connect to, must be specified")
-
+	root.Command.Flags().StringSliceVar(
+		&root.proxyList,
+		"proxy-map",
+		nil,
+		"Map of suffix=gateway pairs to use for choosing proxy. Overrides --proxy when an entry matches any of the targets. If set, --proxy is used as the default fallback when no entries match. Entries should be non-overlapping/able to be applied in any order.",
```
ccontavalli, line 196: "Entries should be non-overlapping/able to be applied in any order."

is easy to miss, and the use of `map` above can cause problems for a subset of users that are hard to debug / pinpoint, as it will fail sometimes but work others (most likely would show up as a transient "ACL" / DNS resolution error?).

I have a slight preference to use a [] of struct above, and at least have consistent result/failure. Rest of the change looks great.

minor-fixes: Agreed; fixed.

ccontavalli (approved):

PR #570, opened 2022-05-09
bazel/linux: add initial qemu support.

ccontavalli:
In this PR:
- add an initial rule so we can use "qemu" instead of uml
  as a test runner.
- break down the "uml" runner so to have the ability to share
  code with the qemu target.
- minor improvements here and there to make the output
  clearer and easier to work with.

Tested:
In a modified internal client, I can see kunit tests run in
qemu instead of uml, and I can see the output parsed correctly.
Will test briefly with new kernel David's just merged, to
see tests actually passing/failing (had to comment out code
as the lack of kunit in non-uml kernel caused the build to
actually fail).  This uses 9p file system.

All uml tests that were passing before are still passing.

There's a few TODOs still to address:

1) Not all parameters are passed in correctly when a rootfs
   is - would need a sub-9p mount with the test in it, need
   the code in the rootfs to understand it.

2) Neither UML nor QEMU mounts any other directory inside
   the qemu image. For debugging and convenience, we want to
   map a home subdir in qemu. Will send a dedicated PR to
   address this both for qemu and uml.

3) Terminal is ... broken when run in INTERACTIVE mode.
   Will send a PR out shortly to address this separately.

4) Installing modules or providing a directory that works
   with modprobe will be a separate PR.


bazel/linux/qemu.bzl
```diff
@@ -0,0 +1,120 @@
...
+    qemu_flags = ctx.attr.qemu_defaults + ctx.attr.qemu_flags
+
+    kernel_flags = ctx.attr.kernel_defaults + ctx.attr.kernel_flags
+    return create_runner(ctx, ctx.attr.archs, code, runfiles = runfiles, extra = {
+        "qemu_search": shell.array_literal(qemu_search),
+        "qemu_flags": shell.array_literal(qemu_flags),
+        "kernel_flags": shell.array_literal(kernel_flags),
+    })
+
+kernel_qemu_test = rule(
```
dsahern, line 78: kunit_qemu_test given the specific use case

ccontavalli: Let me update the comments, copied over from other code, good catch.

Tl;Dr: this is not tied with kunit at all: this just runs code in qemu, whatever that is, and saves the result.
An optional checker is used by the caller to verify the output.

Additionally ... with @george-enf we discussed renaming all those rules, as the names don't really add up anymore. But would rather do a separate cleanup sweep across the directory - so keeping the existing naming convention for now.

ccontavalli: added commit, marking as resolved.

bazel/linux/qemu.bzl
```diff
@@ -0,0 +1,121 @@
+load("//bazel/linux:runner.bzl", "CREATE_RUNNER_ATTRS", "create_runner")
+load("@bazel_skylib//lib:shell.bzl", "shell")
+
+DEFAULT_QEMU_FLAGS = [
+    "-enable-kvm",
+    "-cpu",
+    "host",
+    "-machine",
+    "pc,accel=kvm,usb=off,dump-guest-core=off",
```
george-enf, line 9: Minor but good to know that this will fail on presubmit. There is no KVM there. I was doing `accel=kvm:tcg`.

george-enf (approved):

bazel/linux/qemu.bzl
```diff
@@ -0,0 +1,121 @@
...
+                doc = "Architectures supported by this test",
+            ),
+            "kernel_defaults": attr.string_list(
+                default = DEFAULT_KERNEL_FLAGS,
+                doc = "Default parameters passed on the kernel command line",
+            ),
+            "kernel_flags": attr.string_list(
+                doc = "Additional flags to pass to the kernel. Appended to kernel_defaults",
+            ),
+            "qemu_binary": attr.label(
```
rajachan, line 101: So this could be the one we build for FE, for instance?

rajachan (approved):
This is great!

bazel/linux/qemu.bzl
ccontavalli: yep, absolutely. It could also be a shell script or binary that takes qemu options, for example, and then starts the model, waits, and forwards "$@" to qemu.

bazel/linux/qemu.bzl
ccontavalli: if that's ok to you, will change the default in a follow up PR as I see this failing, or add some shell hackery to pick the best one depending one environment.

PR #578, opened 2022-05-10
bazel/linux: turn {uml|qemu} test rules into runnable rules.

ccontavalli:
Background:
So far, all the uml_|qemu_ virtualization rules were for testing
only. Now we want to use them just to run an emulator.

This PR changes the rules from "test = True" to "executable = True",
with all that it implies.

Specifically:
- refactor the .sh scripts - they can no longer rely on TEST_
  environment variables being set.

- introduce an exec_test.bzl utility - that can turn any executable
  rule into a test rule.

- use said rule in existing definitions so to maintain backward
  compatibility with the existing ruleset.

- have the shell scripts in the rules be more useful. Specifically:
  - it is now possible to override emulator/kernel/rootfs parameters
    directly from the command line.

  - paths and runtime info is clearly displayed on the screen, removing
    sandboxing parameters, so it's very easy to debug those scripts
    manually.

Tested:
Existing regression tests are still passing in internal/. An example
PR in qemu is failing exactly as it was before :) needs the module
loading plumbing to be useful.

### DIFF
--- a/bazel/linux/qemu.bzl
+++ b/bazel/linux/qemu.bzl
@@ -56,7 +56,7 @@ QEMU_FLAGS+=("${{EMULATOR_OPTS[@]}}")
 
 echo Running qemu: "$QEMU_BINARY" "${{QEMU_FLAGS[@]}}"
 if [ -z "$INTERACTIVE" ]; then
-    "$QEMU_BINARY" "${{QEMU_FLAGS[@]}}" | tee "$OUTPUT_FILE"
+    "$QEMU_BINARY" "${{QEMU_FLAGS[@]}}" </dev/null | tee "$OUTPUT_FILE"
 else
     "$QEMU_BINARY" "${{QEMU_FLAGS[@]}}"
 fi
--- a/bazel/linux/qemu.bzl
+++ b/bazel/linux/qemu.bzl
@@ -25,7 +25,7 @@ DEFAULT_KERNEL_FLAGS = [
     "panic=-1",  # reboot (exit qemu) immediately on panic.
 ]
 
-def _kernel_qemu_test(ctx):
+def _kernel_qemu_run(ctx):
     code = """
 QEMU_FLAGS={qemu_flags}
 QEMU_SEARCH={qemu_search}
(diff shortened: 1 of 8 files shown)


bazel/utils/exec_test.bzl
```diff
@@ -0,0 +1,59 @@
+load("@bazel_skylib//lib:shell.bzl", "shell")
+
+def _exec_test(ctx):
+    # A test rule must return an executable file, cannot reuse
+    # the executable returned by another rule.
+    template = """#!/bin/sh
+ARGS={argv}; ARGS+=("$@")
+exec {script} "$ARGS[@]"
```
cbrune, line 8: Are you sure this syntax works with `/bin/sh` ?  This looks like `/bin/bash` syntax.

cbrune (approved):
Other than the question of `/bin/sh` v. `/bin/bash` this all looks good to me.

bazel/utils/exec_test.bzl
ccontavalli: good catch, fixing it.

bazel/linux/qemu.bzl
```diff
@@ -51,11 +51,12 @@ fi
 test -z "$KERNEL" || QEMU_FLAGS+=("-kernel" "$KERNEL")
 test -z "$INTERACTIVE" || KERNEL_FLAGS+=("init=/bin/sh")
 
-QEMU_FLAGS+=("-append" "${{KERNEL_FLAGS[*]}}")
+QEMU_FLAGS+=("-append" "${{KERNEL_FLAGS[*]}} ${{KERNEL_OPTS[*]}}")
+QEMU_FLAGS+=("${{EMULATOR_OPTS[@]}}")
 
 echo Running qemu: "$QEMU_BINARY" "${{QEMU_FLAGS[@]}}"
 if [ -z "$INTERACTIVE" ]; then
-    "$QEMU_BINARY" "${{QEMU_FLAGS[@]}}" | tee "$OUTPUTFILE"
```
george-enf, line 58: Does this work for you?

I have to change it to:
```
"$QEMU_BINARY" "${{QEMU_FLAGS[@]}}" < /dev/null | tee "$OUTPUT_FILE"
```

With strace, qemu is stuck at: `ioctl(0</dev/pts/1>, SNDCTL_TMR_START or TCSETS, {B38400 opost -isig -icanon -echo ...}`

With gdb, qemu receives a signal `Thread 1 "qemu-system-x86" received signal SIGTTOU, Stopped (tty output).`

PR #579, opened 2022-05-10
linux/defs.bzl: Use enkit auth for kernel_tree_version

rajachan:
While ctx.download_and_extract() does support a basic auth scheme and
one via .netrc, neither of those models match enkit proxy's auth model.
It is desirable to avoid the need for public URLs for artifacts that are
pushed to astore, so this commit uses the enkit CLI to directly fetch
the artifacts using enkit auth.

Testing: Without proper authentication, I get the following error when
trying to build a kernel driver target (while overriding enkit to slurp
in this change).

```
Stderr:
ERROR: Attempting to use your credentials failed with:
    (for debug only) rpc error: code = Unauthenticated desc = invalid credentials - signature expired

This probably means that you just need to log in again with:
        enkit login raghu@enfabrica.net
```

The same build succeeds once I authenticate with my credentials.

Jira: INFRA-1027


bazel/linux/defs.bzl
```diff
@@ -94,8 +137,12 @@ To create a .tar.gz suitable for this rule, you can use the kbuild tool, availab
...
             mandatory = True,
         ),
-        "url": attr.string(
-            doc = "The url to download the package from.",
+        "path": attr.string(
+            doc = "The astore path to download the package from.",
+            mandatory = True,
+        ),
+        "uid": attr.string(
+            doc = "The astore UID for this package.",
```
rajachan, line 145: Comment for L#151 further down (Github is lame, doesn't let you comment on unmodified lines).... 

I'd like to drop the `auth` param too, now that it isn't wired up to anything. Will do that in the next rev.

bazel/linux/defs.bzl
```diff
@@ -94,8 +137,12 @@ To create a .tar.gz suitable for this rule, you can use the kbuild tool, availab
             doc = "The name of the downloaded kernel. Format is 'distribution-kernel-version-arch', like debian-5.9.0-rc6-rt-amd64.",
             mandatory = True,
         ),
-        "url": attr.string(
-            doc = "The url to download the package from.",
+        "path": attr.string(
```
ccontavalli, line 140: I'd like to retain the ability to fetch by url. Retaining that ability will also allow us a smoother transition, as we can merge your PR in enkit, then update internal, then update the rules to use astore without ever needing atomic commits, and with ease of roll back.

Suggestion:
1) Keep the url attribute here.
2) In the implementation above, just do something like 'if not ctx.attr.url: ...'

(you can use fail() to error out if both astore path and url are provided, or neither of them)

bazel/linux/defs.bzl
```diff
@@ -7,8 +7,51 @@ load("@bazel_skylib//lib:shell.bzl", "shell")
 
 def _kernel_tree_version(ctx):
     distro, version = ctx.attr.package.split("-", 1)
+    astore_path = ctx.path(ctx.attr.path.split("/")[-1])
```
ccontavalli, line 10: this is duplicating a lot of the code we have in astore_package rule.

Suggestion:
1) Add astore_package rule to enkit as is (just the astore_package).
2) Make the implementation a callable function. I would probably recommend doing something like here: https://github.com/enfabrica/enkit/blob/master/bazel/linux/runner.bzl#L33 (define the attributes needed in a dict, have a callable function that is not a rule). This way, we can add astore  download support to the kernel_image or ... kernel_whatever rules as well in 2-3 lines of code.
3) call the function from here, rather than duplicating the code.

ccontavalli (commented):
This is a great starting point! Thanks for doing this.

rajachan:
Moving to draft to save watchers from some noise as I prepare to push up changes and watch presubmits fail:)

bazel/astore/defs.bzl
```diff
@@ -118,3 +118,53 @@ astore_download = rule(
...
+        ))
+
+    ctx.extract(
+        archive = f,
+        output = ".",
+        stripPrefix = "",
+    )
+    ctx.delete(f)
+
+    return None
```
ccontavalli, line 170: return None is the default in python, remove.

bazel/linux/defs.bzl
```diff
@@ -95,8 +103,16 @@ To create a .tar.gz suitable for this rule, you can use the kbuild tool, availab
...
+            doc = "The URL to download the package from. This is mutually exclusive with the astore path/uid arguments.",
+            mandatory = False,
+        ),
+        "path": attr.string(
+            doc = "The astore path to download the package from.",
+            mandatory = False,
+        ),
+        "uid": attr.string(
+            doc = "The astore UID for this package.",
+            mandatory = False,
```
ccontavalli, line 115: mandatory = False is the default, remove here and above?

bazel/astore/defs.bzl
```diff
@@ -118,3 +118,53 @@ astore_download = rule(
 With this rule, you can easily download 
 files from an artifact store.""",
 )
+
+"""Fetches an astore package, validates digest, and extracts contents."""
```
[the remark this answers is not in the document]
ccontavalli: I would also highlight that this is meant to be used from "repository rules" (reason: "repository rules" are very different from normal "rules": they don't work off a dependency graph, and ctx is actually a different kind of object).

ccontavalli (commented):
almost there, minor nits, and then happy to approve.

rajachan:
Addressed comments. Also added the repo rule that uses the function. Once this merges, I'll send PRs to:
- Remove the rule from our internal repos and cut over to the enkit rule
- Update the targets using the unauthenticated downloads to astore instead.

bazel/astore/defs.bzl
```diff
@@ -113,8 +113,78 @@ astore_download = rule(
...
+        archive = f,
+        output = ".",
+        stripPrefix = "",
+    )
+    ctx.delete(f)
+
+astore_package = repository_rule(
+    implementation = astore_download_and_extract,
+    attrs = {
+        "path": attr.string(
```
ccontavalli, line 180: `doc =` for the attributes here?

rajachan: Added.

PR #587, opened 2022-05-11
wip: bundle a binary with its dynamic libraries for deployment in a VM

george-enf:
Test:

Add the following in `ib/util/BUILD.bazel`:
```
load("@enkit//bazel/linux:defs.bzl", "bundle")
bundle(
    name = "bundle-sf_111",
    deps = [
        "//ib/apps:sf_111",
    ]
)
```

And run:
```
$ bazel build ib/util:bundle-sf_111 --override_repository=enkit=$(realpath ~/enkit)
$ (cd bazel-bin/ib/util/bundle-sf_111-root && find -type f)
...
./_solib_k8/_U@rdma-core_S_S_Crdma-core___Uexternal_Srdma-core_Srdma-core_Slib/libmlx5.so.1
./_solib_k8/_U@rdma-core_S_S_Crdma-core___Uexternal_Srdma-core_Srdma-core_Slib/libibverbs.so.1
./_solib_k8/_U@rdma-core_S_S_Crdma-core___Uexternal_Srdma-core_Srdma-core_Slib/libmlx4.so.1
./ib/apps/sf_111
```

### DIFF
--- a/bazel/linux/defs.bzl
+++ b/bazel/linux/defs.bzl
@@ -991,3 +991,31 @@ def kernel_test(name, kernel_image, module, rootfs_image = None, test_dir_cfg = 
     if rootfs_image:
         cfg = mconfig(cfg, rootfs_image = rootfs_image)
     name_runner = mcreate_rule(name, runner, "", runner_cfg, kwargs, cfg)
+
+def _bundle(ctx):
+    out = []
+
+    for dep in ctx.attr.deps:
+        for f in dep.default_runfiles.files.to_list():
+            out.append(f)
+
+    d = files_to_dir(
+        ctx,
+        ctx.attr.name + "-root",
+        out,
+    )
+
+    return [
+        DefaultInfo(files = depset([d]))
+    ]
+
+bundle = rule(
+    implementation = _bundle,
+    attrs = {
+        "deps": attr.label_list(
+            mandatory = True,
+            providers = [],
+            doc = "",
+        ),
+    },
+)


bazel/linux/defs.bzl
```diff
@@ -991,3 +991,31 @@ def kernel_test(name, kernel_image, module, rootfs_image = None, test_dir_cfg =
...
+        ctx,
+        ctx.attr.name + "-root",
+        out,
+    )
+
+    return [
+        DefaultInfo(files = depset([d]))
+    ]
+
+bundle = rule(
```
ccontavalli, line 1012: doc = attribute to explain the goal of the rule.

bazel/linux/defs.bzl
```diff
@@ -991,3 +991,31 @@ def kernel_test(name, kernel_image, module, rootfs_image = None, test_dir_cfg =
...
+    implementation = _bundle,
+    attrs = {
+        "deps": attr.label_list(
+            mandatory = True,
+            providers = [],
```
ccontavalli, line 1017: can be omitted - effectively the rule accepts any kind of dependencies?

Well, do they have to be executable? If yes, you probably want `executable = True` or similar.

bazel/linux/defs.bzl
```diff
@@ -991,3 +991,31 @@ def kernel_test(name, kernel_image, module, rootfs_image = None, test_dir_cfg =
...
+            out.append(f)
+
+    d = files_to_dir(
+        ctx,
+        ctx.attr.name + "-root",
+        out,
+    )
+
+    return [
+        DefaultInfo(files = depset([d]))
```
ccontavalli, line 1009: here, in addition to DefaultInfo, you must return a KernelRuntimeInfo for the bundle to be usable with the other qemu/uml rules I added so far.

To return a KernelRuntimeInfo, you're half way there: you are using files_to_dir to shove all the files in a directory (good!), the part that is missing is generating an init script to run? to execute the bundle.

Suggestions to get there:
- each dep above contains "an executable" to run, the way I'd access it is something like `dep[DefaultInfo]'.files_to_run.executable` and then `.path` (or `.short_path`, never get it right).
- I'd use `ctx.actions.declare_file` and `ctx.actions.write` to generate a shell script (our init) that just runs each executable one after the other with some helpful debug messages, and propagates the exit status. 
- I'd pass the `file object` returned by `ctx.actions.declare_file` to `files_to_dir` to make sure the init is actually copied into the output dir.

bazel/linux/defs.bzl
```diff
@@ -991,3 +991,31 @@ def kernel_test(name, kernel_image, module, rootfs_image = None, test_dir_cfg =
     if rootfs_image:
         cfg = mconfig(cfg, rootfs_image = rootfs_image)
     name_runner = mcreate_rule(name, runner, "", runner_cfg, kwargs, cfg)
+
+def _bundle(ctx):
```
ccontavalli, line 995: minor nit: the current naming convention in the file is unfortunate, a bundle is something else (a kernel module compiled against different archs/kernel - a bundle of the same kernel module).

Right now, bundles are called "runtimes", I'd like to move them (this, and the only other one defined already) into a runtimes.bzl, and probably do a cleanup sweep to rename all rules to decouple from kernel / have more sensible names.

(really: some of the stuff we're adding now has to do with running in an emulator, not much with kernel builds)

PR #590, opened 2022-05-11
kbuild: Publish vmlinuz and modules.tar.gz

cbrune:
This PR updates the kbuild scripts to publish the kernel image and a
tarball of the kernel modules for each kernel variant.

The new files in astore are:

```
  kernel/<branch>/<flavour>/vmlinuz
  kernel/<branch>/<flavour>/modules.tar.gz
```
  
Where flavour is one of generic, minimal, and test (UML).

Also this PR adds an astore tag to each kernel artifact of the form
`kernel=<kernel_version>`.

You inspect samples of this build update in astore here:
```
  $ enkit astore ls -l home/curt/scratch/kernel/enf/impish-19.19/minimal
  $ enkit astore ls -l home/curt/scratch/kernel/enf/impish-19.19/test
```

Fixes: INFRA-1041
Fixes: INFRA-1046


kbuild/v2/scripts/lib.sh
```diff
@@ -33,3 +33,30 @@ get_deb_version() {
...
+    local tag="$4"
+    local public="$5"
+
+    if [ ! -r "$archive" ] ; then
+        echo "ERROR: unable to find archive: $archive"
+        exit 1
+    fi
+
+    # upload archive to astore
+    enkit astore upload "${archive}@${astore_path}" -a $arch -t "$tag"
```
ccontavalli, line 50: temporarily disable tagging, until 1047 is addressed? It'll create packages that are not downloadable otherwise.

kbuild/v2/scripts/lib.sh
```diff
@@ -33,3 +33,30 @@ get_deb_version() {
...
+
+    if [ "$public" = "private" ] ; then
```
ccontavalli, line 52: public / private business can go away once @rajachan changes are in?

ccontavalli (approved):

kbuild/v2/scripts/lib.sh
cbrune: ok

kbuild/v2/scripts/lib.sh
cbrune: It looks like those changes are in.  So everything can be made private.

cbrune: @ccontavalli -- I disabled tagging and the public/private stuff.  PTAL.

PR #592, opened 2022-05-11
astore: Fix bug with astore_download_and_extract implementation

rajachan:
A repository rule's implementation only gets the ctx object as a
parameter, and this commit fixes that with a wrapper function. In a
perfect world, we would pass any attribute needed in via ctx.attr, but
like noted in the inline comment, kernel_tree_version uses "sha256" in
place of "digest", and changing it now would break external consumers of
the rule and a rolling update will not be feasible. Longer-term, this
wrapper should be replaced with a dict of attributes for the various
consumers of the common function.

The commit also introduces a strip_prefix attribute for the general
astore_package rule.

Tested manually with both astore_package consumers (which was missed
with cdf318cf0) and kernel_tree_version consumers (with and without an
authenticated download).

### DIFF
--- a/bazel/astore/defs.bzl
+++ b/bazel/astore/defs.bzl
@@ -170,12 +170,19 @@ def astore_download_and_extract(ctx, digest, stripPrefix):
     ctx.extract(
         archive = f,
         output = ".",
-        stripPrefix = "",
+        stripPrefix = stripPrefix,
     )
     ctx.delete(f)
 
+# This wrapper is in place to allow a rolling upgrade across Enkit and the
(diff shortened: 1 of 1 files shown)


ccontavalli (approved):
How did you test it?

rajachan:
> How did you test it?

Tested manually with both astore_package consumers (which was missed with https://github.com/enfabrica/enkit/commit/cdf318cf090ac29709f2b412e1e0049ae784f08d) and kernel_tree_version consumers (with and without an authenticated download). The specific target which called out this bug also works now. Will get the reporter to confirm this as well.

PR #595, opened 2022-05-12
bazel: Upgrade to 5.1.1

minor-fixes:
We suspect that bazel 5.1.0 has an issue reporting error output when
running remotely that upgrading will fix.

Tested: Internal shows breakage reason when upgraded and pointing at
this enkit

Jira: INFRA-1054


ccontavalli:
did you see that downgrading solves the problem?

minor-fixes:
Yes, and actually there's a patch release to fix this issue. Updated the PR desc

PR #597, opened 2022-05-12
bazel/astore: add support for providing a BUILD.bazel file.

ccontavalli:
astore_package is a repository rule that allows to download a
package directly from astore, after authentication.

In this PR:
- add ability to specify a BUILD.bazel file to copy in the
  extracted location.

Tested:
Added a blob using the feature, run a bunch of builds using it.
You can see PR4981.


rajachan (approved):
LGTM

PR #602, opened 2022-05-13
enkit-astore: add meta-data output command line flags

cbrune:
This patch adds two command line options to the `enkit astore`
commands for formatting the output of artifact meta-data.

```
  -m, --meta-file string       Meta-data output file (default "-")
  -f, --meta-format string     Meta-data output format. One of [text, json] (default "text")
```

These options can be used with any astore command that outputs
artifact meta data, including:

- astore ls
- astore note
- astore download
- astore tag
- astore public
- astore upload

Running the `enkit astore ls` command with JSON output to stdout,
piped through `jq .`, looks like this:

```json
$ bazel run //enkit:enkit -- astore ls kernel/enf/impish-19.19/minimal/build-headers.tar.gz -l -f json | jq .

[                                                                                                                                                   [85/35147]
  {
    "created": 1651805121388725000,
    "creator": "david@enfabrica.net",
    "arch": "amd64",
    "md5": "853102bbcb3a6cf0305d87a11173f9bf",
    "sid": "2h/ny/tq6foqdrtudojw35dmkx4cz846vq",
    "uid": "yqeikizdt6ofyc4uxt46cxbiiwz6gokt",
    "size": 13229515,
    "note": "",
    "tags": [
      "latest"
    ]
  },
  {
    "created": 1651190180548809000,
    "creator": "david@enfabrica.net",
    "arch": "amd64",
    "md5": "430aaeb5e1facead28969bd23a4f4b12",
    "sid": "2n/e2/4orwcghvvnnv6ew3rhfbj8nb8pqg",
    "uid": "osedtx3766kkg6mnx8bh2iwp3i52bayh",
    "size": 13227673,
    "note": "",
    "tags": []
  },
  {
    "created": 1651189530301254000,
    "creator": "david@enfabrica.net",
    "arch": "amd64",
    "md5": "430aaeb5e1facead28969bd23a4f4b12",
    "sid": "st/7w/3wjipb27swx7t8svfux5j6k7rgk5",
    "uid": "gvomb82d278txu6ud8pof47vwjzk44c4",
    "size": 13227673,
    "note": "",
    "tags": []
  }
]
```

Fixes:  INFRA-1039

### DIFF
--- a/astore/client/commands/commands.go
+++ b/astore/client/commands/commands.go
@@ -7,6 +7,7 @@ import (
 	"github.com/enfabrica/enkit/lib/client"
 	"github.com/enfabrica/enkit/lib/config"
 	"github.com/enfabrica/enkit/lib/config/defcon"
+	"github.com/enfabrica/enkit/lib/config/marshal"
 	"github.com/enfabrica/enkit/lib/kflags"
 	"github.com/enfabrica/enkit/lib/kflags/kcobra"
 	"github.com/spf13/cobra"
@@ -21,6 +22,7 @@ type Root struct {
 	*client.BaseFlags
 
 	store *client.ServerFlags
+	outputFile string
 }
 
 func New(base *client.BaseFlags) *Root {
@@ -64,10 +66,29 @@ func NewRoot(base *client.BaseFlags) *Root {
 	}
 
 	rc.store.Register(&kcobra.FlagSet{FlagSet: rc.PersistentFlags()}, "")
+	rc.Command.PersistentFlags().StringVarP(&rc.outputFile, "meta-file", "m", "",
+		fmt.Sprintf("Meta-data output file. Supported formats: %s", marshal.Formats()))
+
 	return rc
 }
 
 func (rc *Root) StoreClient() (*astore.Client, error) {
+	if rc.outputFile != "" {
+		// check output file type is supported
+		marshaller := marshal.ByExtension(rc.outputFile)
+		if marshaller == nil {
+			return nil, fmt.Errorf("Output file extension not supported `%s`.  Supported formats: %s",
+				rc.outputFile, marshal.Formats())
+		}
+
+		// check that the destination is writable
+		file, err := os.Create(rc.outputFile)
+		if err != nil {
+			return nil, fmt.Errorf("Problems creating output file `%s` - %w", rc.outputFile, err)
+		}
+		file.Close()
+	}
+
 	_, cookie, err := rc.IdentityCookie()
 	if err != nil {
 		return nil, err
@@ -82,8 +103,19 @@ func (rc *Root) StoreClient() (*astore.Client, error) {
 }
 
 func (rc *Root) Formatter(mods ...Modifier) astore.Formatter {
-	return NewTableFormatter(mods...)
+	formatterList := NewFormatterList()
+
+	// Always include the stdout table formatter
+	formatterList.Append(NewTableFormatter(mods...))
+
+	if rc.outputFile != "" {
+		// add a marshal-aware formatter
+		formatterList.Append(NewMarshalFormatter(rc.outputFile))
+	}
+
+	return formatterList
 }
+
 func (rc *Root) OutputArtifacts(arts []*arpc.Artifact) {
 	formatter := rc.Formatter(WithNoNesting)
 	for _, art := range arts {
--- a/astore/client/commands/formatter.go
+++ b/astore/client/commands/formatter.go
@@ -2,10 +2,13 @@ package commands
 
 import (
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/dustin/go-humanize"
+	castore "github.com/enfabrica/enkit/astore/client/astore"
 	"github.com/enfabrica/enkit/astore/rpc/astore"
+	"github.com/enfabrica/enkit/lib/config/marshal"
 	"github.com/fatih/color"
 )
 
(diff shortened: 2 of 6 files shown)


astore/client/commands/commands.go
```diff
@@ -21,6 +21,9 @@ type Root struct {
 	*client.BaseFlags
 
 	store *client.ServerFlags
+	outputFormat string
+	outputFile string
+	outputWriter *os.File
```
ccontavalli, line 26: use `io.Writer` instead of `os.File`. os.File is a file on the filesystem, `io.Writer` is anything you can write into. io.Writer is accepted by Fprintf and pretty much the entire standard library.

astore/client/commands/formatter.go
```diff
@@ -85,21 +90,88 @@ func (ff *TableFormatter) Element(el *astore.Element) {
...
+		onePrinted: false,
+		outputFile: outputFile,
+	}
+	return j
+}
+
+func (jf *JsonFormatter) Artifact(af *astore.Artifact) {
+	if !jf.onePrinted {
+		// open array of objects
+		fmt.Fprintf(jf.outputFile, "[")
```
ccontavalli, line 128: proposal: rather than print the json manually from here, how about:
1) accumulate in an array / struct
2) on Flush, use "lib/config/marshal" `MarshalFile("name of the file", pointer to struct)`? MarshalToFile will encode in json/yaml/gob/toml based on the file extension, and serialize using an official library (so all escaping etc is done properly).

astore/client/commands/formatter.go
```diff
@@ -85,21 +90,88 @@ func (ff *TableFormatter) Element(el *astore.Element) {
...
+	} else {
+		// trailing comma after previous row
+		fmt.Fprintf(jf.outputFile, ",\n")
+	}
```
george-enf, line 132: go doesn't have a `json.dumps` equivalent?

astore/client/commands/commands.go
```diff
@@ -64,10 +67,27 @@ func NewRoot(base *client.BaseFlags) *Root {
 	}
 
 	rc.store.Register(&kcobra.FlagSet{FlagSet: rc.PersistentFlags()}, "")
+	rc.Command.PersistentFlags().StringVarP(&rc.outputFormat, "meta-format", "f", "text", "Meta-data output format. One of [text, json]")
+	rc.Command.PersistentFlags().StringVarP(&rc.outputFile, "meta-file", "m", "-", "Meta-data output file")
```
ccontavalli, line 71: suggestion: instead of having format and file, just have "meta-file"? if empty, no meta-file is created. If non-empty, a meta-file is created based on the extension of the file? See also suggestion below.

I'm also wondering if we should always have the output to the terminal (well, it's disabled easily with >/dev/null), so the meta file is something orthogonal one can ask for (or not).

astore/client/commands/formatter.go
cbrune: ok

astore/client/commands/commands.go
cbrune: ok

cbrune:
Update the PR with the suggestions. @ccontavalli @minor-fixes 

For astore paths with artifacts, e.g. `kernel/enf/impish-19.19/minimal/build-headers.tar.gz`, the rendered JSON looks like:
```json
{
  "Artifacts": [
    {
      "sid": "2h/ny/tq6foqdrtudojw35dmkx4cz846vq",
      "uid": "yqeikizdt6ofyc4uxt46cxbiiwz6gokt",
      "tag": [
        "latest"
      ],
      "MD5": "hTECu8s6bPAwXYehEXP5vw==",
      "size": 13229515,
      "creator": "david@enfabrica.net",
      "created": 1651805121388725000,
      "architecture": "amd64"
    },
    {
      "sid": "2n/e2/4orwcghvvnnv6ew3rhfbj8nb8pqg",
      "uid": "osedtx3766kkg6mnx8bh2iwp3i52bayh",
      "MD5": "QwquteH6zq0olpvSOk9LEg==",
      "size": 13227673,
      "creator": "david@enfabrica.net",
      "created": 1651190180548809000,
      "architecture": "amd64"
    },
    {
      "sid": "st/7w/3wjipb27swx7t8svfux5j6k7rgk5",
      "uid": "gvomb82d278txu6ud8pof47vwjzk44c4",
      "MD5": "QwquteH6zq0olpvSOk9LEg==",
      "size": 13227673,
      "creator": "david@enfabrica.net",
      "created": 1651189530301254000,
      "architecture": "amd64"
    },
...
    {
      "sid": "fp/zn/espx3dzzyq5bhrmgi3zwcay4rbpb",
      "uid": "wvb8trrpdsarhkv3yrhruvp8q6cvqeir",
      "MD5": "d0vwU7ylyXQkG4b3MqOKsg==",
      "size": 14048022,
      "creator": "curt@enfabrica.net",
      "created": 1638988818637659000,
      "architecture": "amd64"
    }
  ],
  "Elements": null
}
```

For a path with elements, e.g. `kernel/enf/impish-19.19/minimal`, the rendered JSON looks like
```json
{
  "Artifacts": null,
  "Elements": [
    {
      "name": "vmlinuz-modules.tar.gz",
      "created": 1652374316530206000,
      "creator": "curt@enfabrica.net"
    },
    {
      "name": "modules.tar.gz",
      "created": 1652286028108938000,
      "creator": "curt@enfabrica.net"
    },
    {
      "name": "vmlinuz",
      "created": 1652015810066873000,
      "creator": "carlo@enfabrica.net"
    },
    {
      "name": "deb-artifacts.tar.gz",
      "created": 1638988831995017000,
      "creator": "curt@enfabrica.net"
    },
    {
      "name": "build-headers.tar.gz",
      "created": 1638988818614212000,
      "creator": "curt@enfabrica.net"
    },
    {
      "name": "kernel-artifacts.tar.gz",
      "created": 1638832777401825000,
      "creator": "curt@enfabrica.net"
    }
  ]
}
```

astore/client/commands/formatter.go
cbrune: using the marshal code now.

astore/client/commands/commands.go
cbrune: outputWriter is no longer needed

astore/client/commands/commands.go
cbrune: done

astore/client/commands/commands.go
```diff
@@ -64,10 +66,29 @@ func NewRoot(base *client.BaseFlags) *Root {
...
+	if rc.outputFile != "" {
+		// check output file type is supported
+		marshaller := marshal.ByExtension(rc.outputFile)
+		if marshaller == nil {
+			return nil, fmt.Errorf("Output file extension not supported `%s`.  Supported formats: %s",
+				rc.outputFile, marshal.Formats())
+		}
+
+		// check that the destination is writable
+		file, err := os.Create(rc.outputFile)
```
ccontavalli, line 85: nit: I would personally remove this check, there's no guarantee that the destination will still be writable by the time we get to it, I would just handle the error if it happens there.

(that said, I don't feel strongly)

astore/client/commands/commands.go
```diff
@@ -64,10 +66,29 @@ func NewRoot(base *client.BaseFlags) *Root {
...
+		if err != nil {
+			return nil, fmt.Errorf("Problems creating output file `%s` - %s", rc.outputFile, err)
```
ccontavalli, line 87: use '%w' when printing an error, allows the caller to use Unwrap/Wrap which is a pretty cool feature of golang error handling.

(basically: fmt.Errorf turns an error into a different one, which is more appropriate for the API. But %w and Unwrap allow to access the underlying error - so the caller can be like 'was this a network error? yes, retry. Was this a file system error? ...`)

astore/client/commands/commands.go
cbrune: The error is impossible to handle later.  Here at least we have a chance.

Consider the upload command, what will happen is that the upload to astore happens first, and then the attempt to the write the meta-data file fails.  The data uploaded to astore is immutable -- there is no "handling" that error.   Also the upload is an expensive operation (takes a long time), better to fail early.

ccontavalli (approved):
2 minor comments. I'd love docstrings (comments on top of method names / class names that automatically end up on godoc), and a small unit test. But won't block PR without. Approved.

astore/client/commands/commands.go
ccontavalli: sounds good.

astore/client/commands/commands.go
cbrune: done

cbrune:
add docstrings and a unit test

PR #604, opened 2022-05-13
enkit: remember last known good DNS config endpoint

espeer-enfabrica:
Some enkit commands are dynamically loaded over the network. The config
files and other bits pertaining to these commands are cached locally so
that the commands can still be used if the network is later unavailable.

The location of the configuration file for the domain, which is used for
the subsequent cache lookups, is retrieved via DNS. The result of the
DNS lookup, however, was not cached. The net effect of this is that
even though the commands and their config were indeed cached, there was
no mechanism to find them in the absence of DNS.

Storing a previously good fallback configuration endpoint and using this
when the DNS query fails allows the cached commands to work as expected
when disconnected from the network.

Bug: INFRA-970
Signed-off-by: Edwin Peer <edwin@enfabrica.net>

### DIFF
--- a/lib/kflags/kconfig/config.go
+++ b/lib/kflags/kconfig/config.go
@@ -3,8 +3,6 @@ package kconfig
 import (
 	"fmt"
 	"github.com/enfabrica/enkit/lib/cache"
-	"github.com/enfabrica/enkit/lib/config"
-	"github.com/enfabrica/enkit/lib/config/directory"
 	"github.com/enfabrica/enkit/lib/config/marshal"
 	"github.com/enfabrica/enkit/lib/config/remote"
 	"github.com/enfabrica/enkit/lib/kflags"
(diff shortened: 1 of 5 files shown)


lib/kflags/kconfig/config.go
```diff
@@ -281,7 +284,7 @@ func FromFlags(fl *Flags) Modifier {
 	}
 }
 
-func NewConfigAugmenterFromDNS(cs cache.Store, domain string, binary string, mods ...Modifier) (*ConfigAugmenter, error) {
+func NewConfigAugmenterFromDNS(fallback config.Store, cs cache.Store, domain string, binary string, mods ...Modifier) (*ConfigAugmenter, error) {
```
ccontavalli, line 287: fallback here ... is also a key value store, pretty much, just different API and semantics. For caching the old result of resoultion, I'd probably recommend using `cache.Store`? which also has the benefit of saving you from having to change the contract of the function? And bringing in an additional library?

(eg, cache.Store is passed in explicitly with the purpose of caching results of operations)

lib/kflags/kconfig/config.go
```diff
@@ -292,7 +295,13 @@ func NewConfigAugmenterFromDNS(cs cache.Store, domain string, binary string, mod
 	dns := remote.NewDNS(domain, append([]remote.DNSModifier{remote.WithLogger(options.log)}, options.dnso...)...)
 	eps, errs := dns.GetEndpoints()
 	if len(eps) <= 0 {
-		return nil, multierror.NewOr(errs, fmt.Errorf("no endpoints for domain '%s' could be detected - configure TXT records for %s? No connectivity?", domain, dns.Name()))
+		// try fallback endpoint
+		var ep remote.Endpoint
+		_, err := fallback.Unmarshal(endpoint_key, &ep)
```
ccontavalli, line 300: rather than use a constant, use the domain name as key? + some arbitrary namespace qualifier? like "config-discovery-" + domain?

If you use the Cache, it will hash your key, the important part is that it is unique, it's ok if it contains invalid / dangerous characters.

espeer-enfabrica: Good idea, will push an update shortly.

lib/kflags/kconfig/config.go
espeer-enfabrica: cache.Store works very differently - it does not actually store anything in and of itself. Would need to layer a marshaler on top of it, which the config store already provides.

lib/kflags/kconfig/config.go
espeer-enfabrica: If we continue to use the config store, then the domain would already be scoped to the "fallback" namespace.

lib/kflags/kconfig/config.go
```diff
@@ -281,18 +283,45 @@ func FromFlags(fl *Flags) Modifier {
...
+func FallbackConfig(cs cache.Store) (config.Store, error) {
+	dir, _, err := cs.Get("enkit-config-discovery://")
+	if err != nil {
+		return nil, err
+	}
+	dir, err = cs.Commit(dir)
+	if err != nil {
+		return nil, err
+	}
+	store, err := directory.OpenDir(dir)
```
ccontavalli, line 295: I would just use the marshal/unmarshal function, skipt the config.Store layer entirely, with https://pkg.go.dev/github.com/enfabrica/enkit/lib/config/marshal#MarshalFile?

ccontavalli: constant file name, fallback.yaml or ... with the key changing per domain?

espeer-enfabrica: Have pushed this. It doesn't improve the code much, in fact arguably the manual path manipulation is uglier (and not portable, will push another fix using os.PathSeparator in a minute), but it does remove some imports.

ccontavalli (approved):

PR #605, opened 2022-05-16
bazel/linux: create separate template for kunit tests

george-enf:
Do not reuse the init.template.sh for running kunit tests. This template
powers off the VM on exit and kunit tests are not supposed to do the
same.

### DIFF
--- a/bazel/linux/bundles.bzl
+++ b/bazel/linux/bundles.bzl
@@ -26,14 +26,13 @@ def _kunit_bundle(ctx):
             commands.append("load " + mod.short_path)
             inputs.append(mod)
 
-    init = ctx.actions.declare_file(ctx.attr.name + "-init.sh")
+    init = ctx.actions.declare_file(ctx.attr.name + "-kunit.sh")
     ctx.actions.expand_template(
-        template = ctx.file._template_init,
+        template = ctx.file._template_kunit,
         output = init,
         substitutions = {
             "{target}": package(ctx.label),
             "{message}": "KUNIT TESTS",
-            "{relpath}": init.short_path,
             "{commands}": "\n".join(commands),
         },
         is_executable = True,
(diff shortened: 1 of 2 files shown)


bazel/linux/templates/kunit.template.sh
```diff
@@ -0,0 +1,11 @@
...
+
+echo ========= {message} - {target} ==========
+
+function load {
+	echo "... loading $@."
+	insmod "$@"
+}
+
+set -e
+{commands}
```
ccontavalli, line 11: without the blob I had before with `cd ...` it doesn't work when a rootfs is specified explicitly. Can you add it back?

(note to self: we should really start adding unit tests for those rules, it's to the point where it is too easy to regress)

bazel/linux/bundles.bzl
```diff
@@ -77,9 +76,9 @@ and an init script to run them as a kunit test.""",
             default = 5,
             doc = "Maximum recursive depth when expanding a list of kernel module dependencies.",
         ),
-        "_template_init": attr.label(
+        "_template_kunit": attr.label(
```
ccontavalli, line 79: while at it, it may be worth making this template (and _template_check?) public by removing the starting `_` so one can override them?

(up to you, probably worth a separate PR?)

ccontavalli (approved):
only comment I'd like addressed is adding back the blob for rootfs. Maybe add a comment there as well.

bazel/linux/templates/kunit.template.sh
george-enf: Added the blob back.

bazel/linux/bundles.bzl
george-enf: I will create a separate PR for this.

george-enf: Follow up PR: https://github.com/enfabrica/enkit/pull/609

PR #610, opened 2022-05-17
bazel/linux: add modules_bundle rule

george-enf:
This rule brings the requested module and its dependencies in the VM. It
also executes the shell commands of the setup attribute for each
imported module.


george-enf:
@ccontavalli said that there is work pending to change the way we build modules. Potentially, this PR will not be necessary in that case.

PR #620, opened 2022-05-19
bazel/linux: fix kernel_qemu_run when supplied with qemu_binary

george-enf:


rajachan (approved):
Good catch

ccontavalli (approved):

PR #621, opened 2022-05-19
Increase timeout of running jobs

ahungrynacho:
JIRA: [INFRA-1072](https://enfabrica.atlassian.net/browse/INFRA-1072)

Physical design builds often take 24 hours to up to 7 days to complete. The default timeout for currently running invocations and queued license requests is 2 hours.

Testing:
- Currently, there is no staging flextape and staging flexlm servers to work with the staging cluster. How should I proceed to test?

### DIFF
--- a/flextape/client/flextape_client/main.go
+++ b/flextape/client/flextape_client/main.go
@@ -19,7 +19,7 @@ import (
 )
 
 var (
-	timeout = flag.Duration("timeout", 7200*time.Second, "Max time waiting in license queue")
+	timeout = flag.Duration("timeout", 7*24*60*60*time.Second, "Max time waiting in license queue")
 )
 
 func main() {


minor-fixes:
This is a flag, not a variable, so the value can be changed as a one-off per invocation. This change isn't necessary to change the timeout - just add the flag to whatever calls the flextape client.

However, the point of this timeout is to not waste time if something is horribly wrong with the Flextape server (it is keeping things queued but not allocating, for instance). Making it longer hides these issues; we definitely shouldn't make the default that long, and it's unlikely we should make it that long for any invocation.

This timeout wasn't designed to queue up jobs for literal days - we don't have a technical solution for that yet. I recommend people communicate ~daily and arrange license use separately until we do.

PR #622, opened 2022-05-20
bazel/linux: run the init script even in QEMU interactive mode

george-enf:
The user might want to perform some initialization before starting a
shell in the VM.


george-enf:
Pressed the wrong buttons and deleted my branch by mistake. New PR: https://github.com/enfabrica/enkit/pull/624

PR #623, opened 2022-05-20
INFRA-1088: bestie: enforce 20 MB output file size limit

gpaussaenf:
The BES Endpoint will not attempt to read and process any output
file whose size exceeds 20 MB. This is to prevent out-of-memory
conditions being experienced several times per day in the production
cluster. This does not impact the running 'bazel test' job, but
will preclude updating the BigQuery database with any metrics
contained in the unprocessed files.

Note that the built in default of 20 MB for the file size limit
can be overridden using the --max_file_size command line
parameter.

Tested: Ran a local instance of the bestie server along with
a bazel test //systest/src/demo:test_sample_metrics that sent
its BES events to localhost:6433. Observed the desired behavior
when files exceeded the limit.

Jira: INFRA-1088

### DIFF
--- a/bestie/server/main.go
+++ b/bestie/server/main.go
@@ -25,6 +25,7 @@ import (
 var (
 	logger      = log.New(os.Stdout, "bestie: ", log.Ldate|log.Ltime|log.Lmicroseconds|log.LUTC|log.Lshortfile|log.Lmsgprefix)
 	isDebugMode bool // Set this to true to enable certain debug behaviors (e.g. special log messages).
+	maxFileSize int  = (20 * 1024 * 1024)
 )
 
 func debugPrintf(format string, str ...interface{}) {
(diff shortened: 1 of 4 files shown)


bestie/server/main.go
```diff
@@ -25,6 +25,7 @@ import (
 var (
 	logger      = log.New(os.Stdout, "bestie: ", log.Ldate|log.Ltime|log.Lmicroseconds|log.LUTC|log.Lshortfile|log.Lmsgprefix)
 	isDebugMode bool // Set this to true to enable certain debug behaviors (e.g. special log messages).
+	maxFileSize int  = (20 * 1024 * 1024)
```
minor-fixes, line 28: Given the current RAM limit - maybe 5MB is a good starting place? 4x as many concurrent events before OOM

bestie/server/test_result.go
```diff
@@ -57,6 +57,20 @@ func identifyStream(bazelBuildEvent bes.BuildEvent, streamId *build.StreamId) *b
 	return &stream
 }
 
+// Get the file size.
+func isValidFileSize(fileName string) error {
+	f, err := os.Stat(fileName)
```
minor-fixes, line 62: bytestream URLs aren't files, so these won't be found on-disk. I think a different approach is needed

bestie/server/test_result.go
```diff
@@ -57,6 +57,44 @@ func identifyStream(bazelBuildEvent bes.BuildEvent, streamId *build.StreamId) *b
...
 
+func readFileWithLimit(fileReader io.Reader) ([]byte, error) {
+	// Attempt to read the file contents all at once. If an
+	// EOF is encountered, then the file fit within the max
+	// size buffer, so it is eligible for processing. Otherwise,
+	// return an error indicating the file is too big.
+	// Adding 1 to the limit to know if the file size went over it.
+	var limit int = maxFileSize
+	buf := make([]byte, limit+1)
+	n, err := fileReader.Read(buf)
```
minor-fixes, line 68: I guess this is a bit more complicated than I figured - also forgot that io.Reader.Read can return without filling the whole buffer, so it has to be called in a loop. How about something that:

* creates a LimitedReader based on the io.Reader supplied
* does a ReadAll from the LimitedReader
* checks the underlying io.Reader to see if it has any more bytes, and fails if so

maybe like

```
limitRdr := io.LimitReader(fileReader, limit)
contents, err := io.ReadAll(limitRdr)
if err != nil {
  // failed to read
}

// check for one more byte
b := make([]byte, 1)
_, err := fileReader.Read(b)
if err == nil || !errors.Is(err, io.EOF) {
  // file too big, or some other error
}
// contents is valid + sufficiently small
```

This approach also avoids the buffer allocation of size limit

Limit should be passed in as a function arg with this approach

bestie/server/test_result.go
gpaussaenf: OK, see second commit.

bestie/server/main.go
gpaussaenf: Changed from 20 MB to 5 MB.

bestie/server/test_result.go
gpaussaenf: I'll give this a try (probably not today though).

PR #624, opened 2022-05-20
bazel/linux: introduce interactive mode

george-enf:
Rename the previous interactive mode to single-user mode, which
effectively does init=/bin/sh and is useful for debugging.

The new interactive mode is enabled only programmatically with:

vm_bundle(
    name = "interactive",
    prepare_cmds = [
        "INTERACTIVE=1",
    ],
)

The init script is executed and is expected to invoke the shell.

### DIFF
--- a/bazel/linux/uml.bzl
+++ b/bazel/linux/uml.bzl
@@ -18,16 +18,19 @@ else
 fi
 
 # If debugging is enabled, throw the user in a shell.
-if [ -z "$INTERACTIVE" ]; then
+if [ -z "$INTERACTIVE" -a -z "$SINGLE" ]; then
   OPTIONS=("con0=null,fd:1" "con1=null,fd:1" "${{OPTIONS[@]}}")
 else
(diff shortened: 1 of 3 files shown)


bazel/linux/uml.bzl
```diff
@@ -18,16 +18,18 @@ else
 fi
 
 # If debugging is enabled, throw the user in a shell.
-if [ -z "$INTERACTIVE" ]; then
-  OPTIONS=("con0=null,fd:1" "con1=null,fd:1" "${{OPTIONS[@]}}")
-else
+if [ "$SINGLE" ]; then
```
ccontavalli, line 21: `if [ -z "$SINGLE" -a -z "$INTERACTIVE" ]; then`...

and remove the elif below? also, forgot the `-z`?

ccontavalli (approved):

bazel/linux/uml.bzl
george-enf: Rewrote this piece of code to make it more clear. How does it look now?

ccontavalli (approved):

PR #626, opened 2022-05-20
bazel/linux: handle multiple module build types

george-enf:
Store the name of the kernel_module rule as part of the generated
module's path to allow compiling the same module with different
parameters.

Example:
Before: bazel-bin/.../$KERNEL/$ARCH/$MODULE.ko
After:  bazel-bin/.../$KERNEL/$NAME/$ARCH/$MODULE.ko


ccontavalli (approved):
Reference the SF- jira in the commit message? Looks great regardless.

PR #629, opened 2022-05-20
gee pr_make: treat comments like blank lines

jonathan-enf:
From user feedback, some users write PR descriptions like this:

```

gee pr_make: treat comments like blank lines

From user feedback, some users write PR descriptions like this:
```

This then fails because gee deletes all comment lines, and then fails the check
to ensure there is a blank line between the title and the body.  Rather than
changing users, it's probably better to change the tool, so that now comments
are treated like blank lines instead of just deleted.  So, whereas previously
the aforementioned text would render as:

```
gee pr_make: treat comments like blank lines
From user feedback, some users write PR descriptions like this:
```

gee now renders:

```
gee pr_make: treat comments like blank lines

From user feedback, some users write PR descriptions like this:

```

Tested: Created this PR using the new version of gee.  Manual validation
of the sed recipe.


jonathan-enf:

ccontavalli (approved):

PR #631, opened 2022-05-24
bazel/linux: rework repository rules to all stem from one.

ccontavalli:
Before this change, we had a repository rule to download a kernel image,
a repository rule to download a kernel tree, and a set of rules to use
to download kernel modules.

In this PR:
- created a single rule that can handle a single .tar.gz potentially
  containing all of them (build tree, modules, image) in a single package,
  or only one of them.
- refactored code to maintain backward compatibility, while re-using
  this one single rule to rule them all.

The main benefits are:
- this brings in support for a single .tar.gz containing image and modules,
  and to run depmod at module unpack time.
- this allows _all_ kernel dependencies to come from either http or
  astore in WORKSPACE file.

The main drawback is that to use this rule, targets need to be changed
to use a different path from the external workspace, as there's now a single
namespace containing all targets. I suspect this is inevitable.

Tested:
running builds and tests with --override_repository shows no regression.
showcase PR using the new infrastructure works (will push shortly).

### DIFF
--- a/bazel/astore/defs.bzl
+++ b/bazel/astore/defs.bzl
@@ -119,7 +119,7 @@ With this rule, you can easily download
 files from an artifact store.""",
 )
 
-def astore_download_and_extract(ctx, digest, stripPrefix):
+def astore_download_and_extract(ctx, digest, stripPrefix, path = None, uid = None):
     """Fetch and extract a package from astore.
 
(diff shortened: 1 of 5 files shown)


cbrune (approved):
This looks great.

PR #636, opened 2022-05-26
gee diagnose: add ssh diagnostics

jonathan-enf:
If a user is having trouble connecting over ssh to github, the diagnostics
file now contains the information I should need to root cause the issue.

Tested: ran `gee diagnose`, inspected result.


jonathan-enf:

ccontavalli (approved):

PR #637, opened 2022-05-27
bazel/linux: pass TMPDIR in the emulator

george-enf:
The host and the guest share the filesystem so TMPDIR is needed to
instruct well-behaved guest programs to write their temporary files in
the appropriate location (instead of /tmp).


george-enf:
I am closing this PR. Passing data over `TMPDIR` is a bad idea. We should come up with a better plan.

PR #638, opened 2022-05-27
bazel/linux: propagate the exit status outside the emulator

george-enf:
Success or failure inside the emulator is depicted in the exit status of
bazel run. The init script that executes in the emulator stores its exit
status in a temporary file. The script that executes the emulator
adjusts its own exit status based on this file.

Jira: https://enfabrica.atlassian.net/browse/SF-165

### DIFF
--- a/bazel/linux/templates/runner.template.sh
+++ b/bazel/linux/templates/runner.template.sh
@@ -161,6 +161,11 @@ test "$estatus" == 0 || {
     echo 1>&2 "===== emulator exited with non zero status - check logs above ===="
     exit "$estatus"
 }
+test -f "$OUTPUT_DIR/exit_status_file" && estatus=$(< "$OUTPUT_DIR/exit_status_file")
+test "$estatus" == 0 || {
+    echo 1>&2 "===== a program in the emulator exited with non zero status - check logs above ===="
+    exit "$estatus"
+}
 test -z "$INTERACTIVE" -a -z "$SINGLE" || exit "$estatus"
 
 echo 1>&2 "===== emulator exited - now running checkers (if any) ===="
--- a/bazel/linux/templates/init-uml.template.sh
+++ b/bazel/linux/templates/init-uml.template.sh
@@ -1,7 +1,11 @@
 #!/bin/sh
 
 echo ========= {message} - {target} ==========
-trap "poweroff -f" EXIT
+on_exit() {
+    echo $? > "/tmp/output_dir/exit_status_file" || true
+    poweroff -f
+}
+trap on_exit EXIT
 
 # Find the "root" where the package was mounted.
 path="$0"


bazel/linux/templates/runner.template.sh
```diff
@@ -94,6 +94,10 @@ SINGLE=""
 TMPDIR="${TEST_TMPDIR:-$(mktemp -d)}"
 OUTPUT_DIR=${TEST_UNDECLARED_OUTPUTS_DIR:-$(mktemp -d)}
 
+# The program in the emulator should write its exit status to this file.
+EXIT_STATUS_FILE=$TMPDIR/exit_status_file
+KERNEL_OPTS+=("EXIT_STATUS_FILE=$EXIT_STATUS_FILE")
```
ccontavalli, line 99: if we pass an OUTPUT_DIR, I would just assume a specific path in OUTPUT_DIR.

Side note: we should have a README documenting the contract between VM and runner scripts, and scripts inside the VM somewhere. Happy to start writing one.

bazel/linux/templates/runner.template.sh
```diff
@@ -157,6 +161,7 @@ echo 1>&2 "Running emulator..."
 {code}
 
 estatus="$?"
+test "$estatus" == 0 && estatus=$(cat "$EXIT_STATUS_FILE")
```
ccontavalli, line 164: `$(< "$EXIT_STATUS_FILE")` no need for cat in bash.

bazel/linux/templates/init.template.sh
```diff
@@ -1,7 +1,11 @@
 #!/bin/sh
 
 echo ========= {message} - {target} ==========
-trap "poweroff -f" EXIT
+on_exit() {
+    echo $? > "$EXIT_STATUS_FILE" || true
```
ccontavalli, line 5: when `qemu_run` or `qemu_test` is executed, multiple tests can be specified in a single run.

I think we should save the status of each test, ideally also how long each step took (so we can convert those files easily to junit.xml at the end of the run, and show nice break down by test - assuming also that the same 'test script' will run from different 'test targets').

Proposal:
1) Change the {commands} below to have a `run command ...` that saves the status per command?

Question: should we remove the `set -e` global, let all tests run before shutting down the VM? (each  script can have an internal `set -e`? and can call poweroff -f - or a function exported by this script - to stop the test?)

bazel/linux/templates/runner.template.sh
ccontavalli: also, messages below should be updated: they should clearly specify if the emulator failed, one of the tests inside the emulator (which? ideally!).

bazel/linux/templates/runner.template.sh
george-enf: Changed `TMPDIR` to `OUTPUT_DIR`

bazel/linux/templates/runner.template.sh
george-enf: Fixed in next revision.

bazel/linux/templates/init.template.sh
george-enf: I am not sure how we can do that cleanly with what we've got. Some of the commands are initialization commands (e.g. setup mounts). As you pointed out offline, we should split the `commands` into `setup` and `run` phases.

george-enf: Reworked this PR. It's only job is to pass the exit status of the last command from inside the emulator to the host.

ccontavalli (approved):

PR #639, opened 2022-05-31
bazel/linux: pass command line arguments to QEMU wrapper scripts

george-enf:
Prefix all arguments that are handled by the runner itself with
--runner-. The rest of the arguments are stored in an array ARGS and are
available to a potential QEMU wrapper script.

Jira: https://enfabrica.atlassian.net/browse/SF-194

### DIFF
--- a/bazel/linux/templates/runner.template.sh
+++ b/bazel/linux/templates/runner.template.sh
@@ -115,6 +115,8 @@ while getopts "k:e:r:hsx" opt; do
 done
 shift $((OPTIND - 1))
 
+WRAPPER_OPTS=("$@")
+
 showstate
 
 echo 1>&2 "======================================"
@@ -140,11 +142,12 @@ trap onexit EXIT
 #   - OUTPUT_FILE - console output must be stored in this file
 #                   (kernel boot log, and any shell output).
 #   - OUTPUT_DIR - directory where to store any other output file.
+#   - WRAPPER_OPTS - extra command line arguments to emulator wrapper script.
 #
 # - Additionally, they should check for:
 #   - KERNEL_OPTS - array, may have additional kernel arguments.
 #   - EMULATOR_OPTS - array, may have additional arguments for the emulator.
-for var in TARGET KERNEL INIT ROOTFS RUNTIME TMPDIR INTERACTIVE SINGLE OUTPUT_FILE OUTPUT_DIR KERNEL_OPTS EMULATOR_OPTS; do
+for var in TARGET KERNEL INIT ROOTFS RUNTIME TMPDIR INTERACTIVE SINGLE OUTPUT_FILE OUTPUT_DIR KERNEL_OPTS EMULATOR_OPTS WRAPPER_OPTS; do
     export "$var"
 done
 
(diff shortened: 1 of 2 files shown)


bazel/linux/templates/runner.template.sh
```diff
@@ -22,39 +22,39 @@ function help {
 This script executes your target in an emulator.
 Prefer to use as:
 
-  bazel run $TARGET -- [-k kernel option]... [-e emu option]... [-r rootfs] [-s|-x|-h]
+  bazel run $TARGET -- [--runner-kernel-opts kernel option]... [--runner-emulator-opts emu option]... [--runner-rootfs rootfs] [--runner-single|--runner-show-state|--runner-help]
```
ccontavalli, line 25: you also need to update the message in the `onexit` function, and all the Examples provided below.

Alternatives:
1) Leave the options as they are, support `--` like common? so eg, `bazel run :target -- -runner-opt -- -script-opt`
2) Like we have `-k` kernel option, `-e` emulator option, have a `-a` script in VM argument?

bazel/linux/templates/runner.template.sh
```diff
@@ -22,39 +22,39 @@ function help {
...
   bazel build $TARGET
-  bazel-bin/.../path/to/file [-k option]... [-e option]... [-s|-x|-h]
+  bazel-bin/.../path/to/file [--runner-kernel-opts kernel option]... [--runner-emulator-opts emu option]... [--runner-rootfs rootfs] [--runner-single|--runner-show-state|--runner-help]
 
 Accepted options:
 
-  -s           Configures /bin/sh as init, drops you in a shell.
-               Hint: you can then use the paths shown with -x to manually
-               start the init script used.
+  --runner-single                Configures /bin/sh as init, drops you in a shell.
```
ccontavalli, line 34: Change -x below, it's no longer -x? But see comment above.

bazel/linux/templates/runner.template.sh
```diff
@@ -154,6 +161,8 @@ echo 1>&2 "Running preparations..."
 
 echo 1>&2 "======================================"
 echo 1>&2 "Running emulator..."
+export BASH_ENV="$TMPDIR/env"
```
ccontavalli, line 164: should this be OUTPUT_DIR again? so we can save the parameters passed?

bazel/linux/templates/runner.template.sh
```diff
@@ -115,6 +115,8 @@ while getopts "k:e:r:hsx" opt; do
 done
 shift $((OPTIND - 1))
 
+WRAPPER_OPTS=("$@")
```
ccontavalli, line 118: do we need to add support for wrapper_opts in the vm_bundle as well?

ccontavalli (approved):
looks great. before merging, can you confirm that qemu without wrapper has no problem with the extra `--` being passed at the end of the command line?

(if it is, prepend WRAPPER_OPTS with -- if non empty, and then just expand `"${WRAPPER_OPTS[@]}"` similar as today?)

george-enf:
> looks great. before merging, can you confirm that qemu without wrapper has no problem with the extra `--` being passed at the end of the command line?

Good catch. Fixed.

bazel/linux/templates/runner.template.sh
george-enf: Created Jira https://enfabrica.atlassian.net/browse/SF-264

PR #648, opened 2022-06-07
bazel/linux: add support for running multiple tests in an emulator

george-enf:
The following output is printed inside the emulator:

```
//tests:test1                                                            PASSED in 0.2s
//tests:test2                                                            FAILED in 0.3s
//tests:test3                                                            PASSED in 0.4s
==================== Test output for //tests:test2:
test2 script output
================================================================================
Executed 3 test(s): 2 test(s) pass(es) and 1 fail(s).
```

Jira: https://enfabrica.atlassian.net/browse/SF-147

### DIFF
--- a/bazel/linux/templates/runner.template.sh
+++ b/bazel/linux/templates/runner.template.sh
@@ -161,11 +161,19 @@ test "$estatus" == 0 || {
     echo 1>&2 "===== emulator exited with non zero status - check logs above ===="
     exit "$estatus"
 }
+
+# If XML_OUTPUT_FILE is defined and the program in the emulator created a
+# junit.xml, move it to the expected location.
+test "${XML_OUTPUT_FILE}" -a -f "$OUTPUT_DIR/junit.xml" && {
+  mv "$OUTPUT_DIR/junit.xml" $XML_OUTPUT_FILE
+}
+
 test -f "$OUTPUT_DIR/exit_status_file" && estatus=$(< "$OUTPUT_DIR/exit_status_file")
 test "$estatus" == 0 || {
     echo 1>&2 "===== a program in the emulator exited with non zero status - check logs above ===="
     exit "$estatus"
 }
+
 test -z "$INTERACTIVE" -a -z "$SINGLE" || exit "$estatus"
 
 echo 1>&2 "===== emulator exited - now running checkers (if any) ===="
--- a/bazel/linux/test.bzl
+++ b/bazel/linux/test.bzl
@@ -3,12 +3,118 @@ load("//bazel/linux:qemu.bzl", "kernel_qemu_run")
 load("//bazel/utils:macro.bzl", "mconfig", "mcreate_rule")
 load("//bazel/utils:exec_test.bzl", "exec_test")
 load("//bazel/linux:bundles.bzl", "kunit_bundle")
+load("//bazel/linux:runner.bzl", "commands_and_runtime", "get_prepare_run_check")
+load("//bazel/linux:providers.bzl", "RuntimeBundleInfo", "RuntimeInfo")
+load("@bazel_skylib//lib:shell.bzl", "shell")
 
-def qemu_test(name, kernel_image, run, qemu_binary = None,
+def _test_runner(ctx):
+    script_begin = r"""#!/bin/bash
+set -eu -o pipefail
+GREEN=$(echo -en "\033[32m")
+RED=$(echo -en "\033[31m")
+RESET=$(echo -en "\033[0m")
+failed_outputs=()
+junit=""
+tests=0
+failures=0
+run_test() {
+    local title="$1"
+    local script="$2"
+    shift 2
+    local args=("$@")
+    local failed=1
+    local output
+    local status
+    local start_time_ns
+    local stop_time_ns
+    local duration_ns
+    local duration_s
+    local error_msg=""
+
+    printf "%-73s" "${title}"
+    start_time_ns=$(($(date +%s%N)))
+
+    output=$("${script}" "${args[@]}" 2>&1) && failed=0
+
+    tests=$[tests + 1]
+    if [ $failed -eq 0 ]; then
+        status="${GREEN}PASSED${RESET}"
+    else
+        failures=$[failures + 1]
+        status="${RED}FAILED${RESET}"
+        failed_outputs+="==================== Test output for ${title}:\n${output}\n"
+        error_msg="<failure message=\"failed\"/>"
+    fi
+    stop_time_ns=$(($(date +%s%N)))
+    duration_ns=$(($stop_time_ns - $start_time_ns))
+    duration_s=$(echo "$duration_ns/10^9" | bc -l)
+    printf "%s in %.1fs\n" "$status" "$duration_s"
+    junit+="
+        <testsuite name=\"${title}\" tests=\"1\" failures=\"${failed}\">
+            <testcase name=\"${title}\" status=\"run\" duration=\"${duration_s}\" time=\"${duration_s}\">${error_msg}</testcase>
+        </testsuite>
+"
+}
+"""
+
+    script_end = r"""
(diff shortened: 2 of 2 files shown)


bazel/linux/test.bzl
```diff
@@ -20,9 +97,13 @@ def qemu_test(name, kernel_image, run, qemu_binary = None,
         config: dict, all additional attributes to pass to the
             kernel_qemu_run rule, generally created with mconfig().
     """
+
+    runner_script = mcreate_rule(name, test_runner, "test-runner", [],
```
ccontavalli, line 101: with mcreate_rule, merge in kwargs. Otherwise `visibility = ` or `tag = ` will not propagate correctly to all rules instantiated by the macro.

ccontavalli, line 101: update the docstring of the function: now this macro instantiates 3 rules: {name}-test-runner, {name}-run, and {name}

bazel/linux/test.bzl
```diff
@@ -4,6 +4,83 @@ load("//bazel/utils:macro.bzl", "mconfig", "mcreate_rule")
...
+        binary = test.files_to_run.executable
+        tests.append("run_test {title} {script}".format(
+            title = test.label,
+            script = binary.short_path,
+        ))
+        runfiles = runfiles.merge(ctx.runfiles([binary]))
+
+    script = ctx.actions.declare_file("{}_test_runner.sh".format(ctx.attr.name))
+    ctx.actions.write(script, script_begin + "\n".join(tests) + script_end)
+    print(script_begin + "\n".join(tests) + script_end)
```
ccontavalli, line 69: remove print?

bazel/linux/test.bzl
```diff
@@ -4,6 +4,83 @@ load("//bazel/utils:macro.bzl", "mconfig", "mcreate_rule")
...
+    return [DefaultInfo(runfiles = runfiles, executable = script)]
```
ccontavalli, line 70: so... the `tests` labels passed here... if they are `vm_bundle()` targets, this rule does not propagate the `pre` step and `check` step to the runner below. So a `vm_bundle` passed to a qemu_test target would silently break?

ccontavalli: (it may be ok to do so, but I would be a bit more defensive here, see other comment above when parsing list of targets)

bazel/linux/test.bzl
```diff
@@ -4,6 +4,83 @@ load("//bazel/utils:macro.bzl", "mconfig", "mcreate_rule")
...
+fi
+succeeded=$[tests - failed]
+echo "Executed ${tests} test(s): ${succeeded} test(s) pass(es) and ${failed} fail(s)."
+exit $failed
+"""
+
+    tests = []
+    runfiles = ctx.runfiles()
+    for test in ctx.attr.tests:
+        binary = test.files_to_run.executable
```
ccontavalli, line 60: test here could be any target / label. It may not be executable. If it's not, user will get a stack trace very hard to read.

I'd try to be more defensive here, do something like:
```
if not test.files_to_run.executable:
    fail(location(ctx) + ... "you specified a target that is not executable " + package(test))
```
for example.

ccontavalli: if we do end up breaking the vm_bundle features here, I'd add a test to see if there is a bundle provider, if there is (and it has a pre/setup and check step), throw an error?

bazel/linux/test.bzl
```diff
@@ -4,6 +4,83 @@ load("//bazel/utils:macro.bzl", "mconfig", "mcreate_rule")
...
+    executable = True,
+    attrs = {
+        "tests": attr.label_list(
+            mandatory = True,
+            doc = "List of executable targets to run in the emulator as tests. The duration and the exit status of each test is recorded and reported.",
+        ),
+    },
+)
+
 def qemu_test(name, kernel_image, run, qemu_binary = None,
```
ccontavalli, line 84: I think we said we'd add a separate setup step here? separate PR?

bazel/linux/test.bzl
george-enf: Fixed. I assume it just means I have to do `..., kwargs, mconfig(...`

george-enf: Fixed

bazel/linux/test.bzl
george-enf: Removed.

bazel/linux/test.bzl
george-enf: Added.

bazel/linux/test.bzl
george-enf: Added the error message.

bazel/linux/test.bzl
george-enf: This is pending. I will work on it and update the PR.

george-enf: I've created a 2 refactoring PRs and updated this PR with support for `vm_bundle`.

bazel/linux/templates/runner.template.sh
```diff
@@ -161,6 +161,13 @@ test "$estatus" == 0 || {
     echo 1>&2 "===== emulator exited with non zero status - check logs above ===="
     exit "$estatus"
 }
+
+if [ ${XML_OUTPUT_FILE+x} ]; then
+  # The program in the emulator has created a junit.xml. Move it to the
+  # expected location.
+  mv "$OUTPUT_DIR/junit.xml" $XML_OUTPUT_FILE
```
ccontavalli, line 168: 2 things here:
1) `"$XML_OUTPUT_FILE"` (quotes around the $ variable)
2) logic seems fishy? Should the if instead test to see if "$OUTPUT_DIR/junit.xml" exists, and if it does, copy it in the right place?

bazel/linux/test.bzl
```diff
@@ -3,12 +3,118 @@ load("//bazel/linux:qemu.bzl", "kernel_qemu_run")
...
+        failures=$[failures + 1]
+        status="${RED}FAILED${RESET}"
+        failed_outputs+="==================== Test output for ${title}:\\n${output}\\n"
+        error_msg="<failure message=\\\"failed\\\"/>"
+    fi
+    stop_time_ns=$(($(date +%s%N)))
+    duration_ns=$(($stop_time_ns - $start_time_ns))
+    duration_s=$(echo "$duration_ns/10^9" | bc -l)
+    printf "%s in %.1fs\\n" "$status" "$duration_s"
+    junit+='
```
ccontavalli, line 52: escaping is a bit crazy here, I'd consider using `"` instead of `'`, and then \"? may be more readable?

I think `r"""` is supported in starlark, which tells python to ignore the `\` escaping, so it does not become `\\"`

(up to you)

bazel/linux/test.bzl
```diff
@@ -3,12 +3,118 @@ load("//bazel/linux:qemu.bzl", "kernel_qemu_run")
...
+        <testsuite name="'"${title}"'" tests="1" failures="'"${failed}"'">
```
ccontavalli, line 53: nit, and ok to not consider this in this PR, leave a TODO: but when we run a gunit test or py_test in the VM, those will output their own junit.xml file. We should check that environment variables are such they don't overwrite the file you generate, and ideally they'd be merged in the junit.xml file you are creating.

(I know, I know)

ccontavalli (approved):
Only one comment I'd like you to fix, the `if` with "$OUTPUT_DIR/junit.xml", and running shellcheck. The rest are nits / optional, can put TODOs in place.

bazel/linux/test.bzl
george-enf: Changed to `r"""`. Let me know if escaping looks a bit better now.

bazel/linux/test.bzl
george-enf: Good point. I will have to verify that.

PR #649, opened 2022-06-08
astore_package: Raise sha256sum timeout

minor-fixes:
This change raises the timeout for calculating the sha256 of items
downloaded from enkit. This timeout is currently causing false failures
in presubmits - while an individual sha256 should be relatively fast
compared to the old timeout, it is likely that the Cloud Build machine
is downloading/unpacking other items concurrently and is therefore
slower than expected.

Jira: INFRA-1179

### DIFF
--- a/bazel/astore/defs.bzl
+++ b/bazel/astore/defs.bzl
@@ -151,7 +151,7 @@ def astore_download_and_extract(ctx, digest, stripPrefix, path = None, uid = Non
 
     # Check digest of archive
     checksum_args = ["sha256sum", f]
-    res = ctx.execute(checksum_args, timeout = 10)
+    res = ctx.execute(checksum_args, timeout = 60)
     if res.return_code:
         fail("Failed to calculate checksum\nArgs: {}\nStdout:\n{}\nStderr:\n{}\n".format(
             checksum_args,


bazel/astore/defs.bzl
```diff
@@ -151,7 +151,7 @@ def astore_download_and_extract(ctx, digest, stripPrefix, path = None, uid = Non
 
     # Check digest of archive
     checksum_args = ["sha256sum", f]
-    res = ctx.execute(checksum_args, timeout = 10)
+    res = ctx.execute(checksum_args, timeout = 60)
```
jonathan-enf, line 154: I'd go for the gusto and increase this to 600 seconds.  We'd rather that presubmits succeed slowly than fail unnecessarily, yeah?

jonathan-enf (approved):
This is good but I also think we should probably make presubmits fail if this times out, rather than run no tests and succeed.  Can you please do that in a follow-on PR?

minor-fixes:
> This is good but I also think we should probably make presubmits fail if this times out, rather than run no tests and succeed. Can you please do that in a follow-on PR?

It's not so simple; see https://enfabrica.atlassian.net/browse/INFRA-1179?focusedCommentId=18416. Can we continue to discuss in that issue?

PR #650, opened 2022-06-08
enkit affected tests: made broken build graph an error

jonathan-enf:
This PR causes presubmits where either the start or end build graph is broken
to fail.  Rationale: it's better to ask a user who is submitting a PR that
fixes the build graph to force-submit (override the presubmit), than to run the
risk of a PR that isn't fixing the build graph to accidentally have an
incorrect "pass" result.

The rationale for the original code was that this situation only arises
if the user is fixing a build graph that is broken at top-of-tree.  Unfortunately,
bazel isn't 100% reliable at generating build graphs, and so there are
other reasons why a start build graph might fail.

This PR makes it mandatory that a PR that fixes the build graph manually
test, and the override the presubmit in order to merge changes.

Jira: INFRA-1179

Tested: existing unit tests pass.  I am unable to force bazel to fail on
demand, but this change is pretty straightforward.


jonathan-enf:

PR #653, opened 2022-06-09
enkit tunnel: Add --use-srv-port flag

minor-fixes:
This change adds a boolean flag `--use-srv-port`. When set, this flag
ignores the target port arg and instead queries enproxy for an
appropriate target port based on an SRV record lookup.

Tested: Against current enproxy, fails with 404 as expected

Jira: INFRA-1223

### DIFF
--- a/proxy/ptunnel/tunnel.go
+++ b/proxy/ptunnel/tunnel.go
@@ -2,6 +2,7 @@ package ptunnel
 
 import (
 	"encoding/binary"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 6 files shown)


proxy/ptunnel/tunnel.go
```diff
@@ -127,6 +128,31 @@ func GetSID(proxy *url.URL, host string, port uint16, mods ...GetModifier) (stri
...
+	curl := *proxy
+
+	params := proxy.Query()
+	params.Add("host", host)
+	curl.RawQuery = params.Encode()
+	curl.Path = path.Join(curl.Path, "/get_srv_port")
+
+	var body string
+	// TODO: retries?
+	// TODO: options?
```
minor-fixes, line 141: Should I copy the pattern used above for retries, options, errors, etc?

minor-fixes:
The alternative to this PR is to make the target port optional throughout, and have the proxy try to deduce the target port when it's omitted. This feels cleaner architecturally, but seems to be a more intrusive change - feedback on this approach or others is welcome.

PR #654, opened 2022-06-09
bazel/linux: expose /dev/output_dir

george-enf:
Map the $OUTPUT_DIR directory from the host to the device
/dev/output_dir in the guest.

### DIFF
--- a/bazel/linux/qemu.bzl
+++ b/bazel/linux/qemu.bzl
@@ -48,6 +48,10 @@ else
     KERNEL_FLAGS+=("root=/dev/root" "rootfstype=9p" "init=$INIT")
     KERNEL_FLAGS+=("rootflags=trans=virtio,version=9p2000.L,msize=5000000,cache=mmap,posixacl")
 fi
+
+QEMU_FLAGS+=("-fsdev" "local,security_model=none,multidevs=remap,id=fsdev-fsOutputDir,path=$OUTPUT_DIR")
+QEMU_FLAGS+=("-device" "virtio-9p-pci,fsdev=fsdev-fsOutputDir,mount_tag=/dev/output_dir")
+
 test -z "$KERNEL" || QEMU_FLAGS+=("-kernel" "$KERNEL")
 test -z "$SINGLE" || KERNEL_FLAGS+=("init=/bin/sh")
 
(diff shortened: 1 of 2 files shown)


bazel/linux/templates/init.template.sh
```diff
@@ -10,4 +10,17 @@ dir="${path%%$relpath}"
...
 set -e
+
+# Remount / with the nosuid flag. Otherwise the following mount fails with:
+# mount: only root can use "--options" option (effective UID is 65534)
+python -c 'import ctypes; exit(ctypes.cdll.LoadLibrary("libc.so.6").mount("", "/", "", 2|32, 0))'
+
+mount --types tmpfs tmpfs /tmp
+
+# Mount the output directory. This directory is shared with the host.
+mkdir /tmp/output_dir
```
salil-enf, line 21: How does this work with multiple tests running? (I see you have a separate PR for it; not sure if this is answered in a different PR.)
Will two tests running in parallel end up writing in the same dir?

george-enf: The `/tmp/output_dir` is inside each VM and is not shared between them. The host directory where it maps, I believe, is unique.

ccontavalli: As George says. bazel creates a new temp output directory for each test it runs in parallel. Even if within the VM it has the same name, it is mapped to different directories outside the VM.

bazel/linux/templates/init.template.sh
```diff
@@ -10,4 +10,17 @@ dir="${path%%$relpath}"
 test "$dir" == "$path" || cd "$dir"
 
 set -e
+
```
ccontavalli, line 13: I remember init.template.sh being shared between qemu runs and uml test runs with kunit.

Does this change break the uml test runs? uml has no 9p, although it has a comparable file system.

Maybe instead of modifying the template, those commands should be provided by qemu_run or uml_run? not sure.

bazel/linux/templates/init.template.sh
salil-enf: that makes sense. I was expecting to see here that the mount ties the unique test directory from host to a standard path inside the VM. Is that not available yet?

ccontavalli: standard path inside the VM is /tmp/output_dir, from the look of the code.

bazel/linux/templates/init.template.sh
ccontavalli: please test the change with the uml_run targets, btw, although I suspect it'll be caught on merge in internal.

ccontavalli (approved):

PR #665, opened 2022-06-15
bazel: add missing pipefail to bash scripts

george-enf:
Some bash scripts were failing silently due to missing pipefail.

### DIFF
--- a/bazel/linux/templates/runner.template.sh
+++ b/bazel/linux/templates/runner.template.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -e
+set -o pipefail
 
 echo "===== Running target: {target}"
 echo "===== Path: $(realpath $0)"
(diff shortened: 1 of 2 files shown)


bazel/linux/templates/runner.template.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -e
+set -o pipefail
```
ccontavalli, line 4: nit: just do `set -euo pipefail` like in other scripts?

ccontavalli (approved):

bazel/linux/templates/runner.template.sh
george-enf: I tried that first but apparently the kernel tries everything as 1 argument in shebangs... So you can't pass multiple ones. https://unix.stackexchange.com/questions/533415/invalid-option-name-error-with-shebang-bin-bash-o-pipefail-in-script

PR #667, opened 2022-06-16
kunit: updates for kunit inside of QEMU

cbrune:
```
    kunit: do mot mount sysfs on /sys automatically

    This patch removes the automatic mount of sysfs on /sys for a couple
    of reasons:

    1. the kunit framework now uses 'insmod' instead of 'modprobe', so the
    original reason for mounting sysfs no longer exists.

    2. kunit_test() can be run with a custom runner that can handle any
    needed "pre-test" steps, like mounting sysfs or procfs, etc.

    For the case of using a QEMU runner, we will use a common system setup
    set of run commands that already handles mounting various kernel file
    systems.

    Relates-to: MCU-85
```

```
    kunit: append the kunit command steps to the runner steps

    This patch swaps the order of the 'runner_cfg' and 'cfg' parameters in
    the mcreate_rule() call for kunit tests.  The net effect is the
    commands from 'cfg' are appended to the 'runner_cfg' commands, instead
    of the other way around.

    We want the 'runner_cfg' commands to run first, as their job is to
    setup the test environment for the kunit test.  We do not want to run
    the kunit test and then setup the environment.

    Relates-to: MCU-85
```

### DIFF
--- a/bazel/linux/bundles.bzl
+++ b/bazel/linux/bundles.bzl
@@ -121,11 +121,7 @@ def _kunit_bundle(ctx):
     mods = get_compatible(ctx, ki.arch, ki.package, ctx.attr.module)
     alldeps = expand_deps(ctx, mods, ctx.attr.depth)
 
-    commands = [
-        # modprobe does not work correctly without /sys
-        "mount -t sysfs sysfs /sys || echo 1>&2 'Could not mount sys - modprobe may not work properly'",
-    ]
-
+    commands = []
     inputs = []
     for kmod in alldeps:
         commands += ["", "# module " + package(kmod.label)]
--- a/bazel/linux/test.bzl
+++ b/bazel/linux/test.bzl
@@ -173,13 +173,14 @@ def kunit_test(name, kernel_image, module, rootfs_image = None,
         mconfig(module = module, image = kernel_image),
     )
 
-    cfg = mconfig(
-        run = [runtime],
-        kernel_image = kernel_image,
-        kernel_flags = [
-            "printk.time=0", # printk timestamps breaks kunit result parsing
-        ])
+    cfg = mconfig(run = [runtime], kernel_image = kernel_image)
+
     if rootfs_image:
         cfg = mconfig(cfg, rootfs_image = rootfs_image)
+
+    if runner == kernel_qemu_run:
+        # printk timestamps breaks kunit result parsing in the QEMU runner
+        cfg = mconfig(cfg, kernel_flags = [ "printk.time=0" ])
+
     name_runner = mcreate_rule(name, runner, "emulator", cfg, kwargs, runner_cfg)
     exec_test(name = name, dep = name_runner, **kwargs)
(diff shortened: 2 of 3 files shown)


bazel/linux/bundles.bzl
```diff
@@ -121,11 +121,7 @@ def _kunit_bundle(ctx):
     mods = get_compatible(ctx, ki.arch, ki.package, ctx.attr.module)
     alldeps = expand_deps(ctx, mods, ctx.attr.depth)
 
-    commands = [
-        # modprobe does not work correctly without /sys
-        "mount -t sysfs sysfs /sys || echo 1>&2 'Could not mount sys - modprobe may not work properly'",
-    ]
-
+    commands = []
```
ccontavalli, line 124: removing this without other changes will break existing tests and uses? What is the problem with the extra sysfs mount?

Can we first add the mount in users, and then remove the mount from enkit?

bazel/linux/test.bzl
```diff
@@ -176,5 +176,5 @@ def kunit_test(name, kernel_image, module, rootfs_image = None,
     cfg = mconfig(run = [runtime], kernel_image = kernel_image)
     if rootfs_image:
         cfg = mconfig(cfg, rootfs_image = rootfs_image)
-    name_runner = mcreate_rule(name, runner, "emulator", runner_cfg, kwargs, cfg)
+    name_runner = mcreate_rule(name, runner, "emulator", cfg, kwargs, runner_cfg)
```
ccontavalli, line 179: this part I believe looks good.

cbrune:
I just added 4aaff90, which defaults the `kunit_test()` rule to use the QEMU runner.

cbrune:
I just added ac28ff8 

```
    kernel_test: use kunit_test() with a UML runner

    Previously, kernel_test was a straight alias for kunit_test.  The
    current consumers of kernel_test assume the runner is the UML runner,
    however kunit_test now uses the QEMU runner by default.

    This patch rewrites kernel_test as a wrapper around kunit_test passing
    in a UML runner.  This allows legacy kernel_test consumer to continue
    to work until they are converted to use the kunit_test directly.
```

bazel/linux/bundles.bzl
ccontavalli: this is shared with non kunit targets - eg, all kernel_qemu_run and qemu_test targets use bundles and generated code.

If any of them requires modprobe, this will break them.

ccontavalli (commented):
still worried about removing the sysfs mount.

bazel/linux/bundles.bzl
cbrune: ummm, no.  This function is called `def _kunit_bundle(ctx):` -- it is only for kunit.  why would anything but kunit call this function?

this change does not affect any kunit tests.

ccontavalli: sorry, my bad - should have looked more carefully at the filename.

ccontavalli (approved):

PR #674, opened 2022-06-22
enkit/bazel/utils: fix error in exec_test.

ccontavalli:
Background:
I noticed qemu_test rules failing when manually specifying some command
line parameters. Turns out the problem was in exec_test, that did not
expand a bash array correctly.

In this PR:
- fix the shell argument expansion.
- add a tiny test to verify that exec_test works as expected.
  fool me once, yadda yadda.

Also, trying to add examples for bazel rules tests, and examples of
sh_binary() just invoking and using other targets (hint: it's easy!).

### DIFF
--- a/bazel/utils/echo-argv_test_check.sh
+++ b/bazel/utils/echo-argv_test_check.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+cmp <(./bazel/utils/echo-argv_test_test.sh) - <<'EOF'
+arg1
+arg2 with space
+arg3 with `characters`
+EOF
+test "$?" == 1 || {
+    echo "first comparison was expected to fail - test not working?" 1>&2
+    exit 1
+}
+
+set -eu -o pipefail
+
+cmp <(./bazel/utils/echo-argv_test_test.sh) - <<'EOF'
+arg1
+arg2 with space
+arg3 with $NASTY `characters`
+EOF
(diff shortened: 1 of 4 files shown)


skarat-enf (approved):

bazel/utils/exec_test.bzl
```diff
@@ -5,7 +5,7 @@ def _exec_test(ctx):
     # the executable returned by another rule.
     template = """#!/bin/bash
 ARGS={argv}; ARGS+=("$@")
-exec {script} "$ARGS[@]"
+exec {script} "${{ARGS[@]}}"
```
cbrune, line 8: Why are double braces required here?  I'm not a bash array expansion expert.  The examples I see online are only using a single set of braces.

ccontavalli: this is a string expanded with python format. `{`, for python format, needs to be escaped as `{{`. bash sees a single `{'.

bazel/utils/echo-argv_test_check.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
ccontavalli: if ok with it, I'll file a jira for a fixit item. Eg, pretty much ~all scripts in enkit/ use #!/bin/bash. I'd keep consistency, and fix as one.

bazel/utils/exec_test.bzl
cbrune: ah

cbrune (approved):

PR #677, opened 2022-06-27
bazel/utils: initial set of rules to run an arbitrary target on a remote machine.

ccontavalli:
See the top of remote.bzl file for an explanation of the ruleset added.

Why? Typical development cycle has developers build scripts or tests
locally, and wanting to try them on a remote machine. I expect we'll have
something better and more dynamic with machine allocations in the future,
but this is a starting point for manual testing.

SF-218 for background as well.

This is a first step: qemu run rules will be modified to better play
with the remote rules, and allow baremetal run, and we'll need an
official wrapper to prepare the environment on the remote machine.

### DIFF
--- a/bazel/utils/remote.bzl
+++ b/bazel/utils/remote.bzl
@@ -1,0 +1,300 @@
+"""
+Simple rules to allow running targets on remote machines using rsync + ssh.
+
+Basic usage is trivial. For example, by defining:
+
+    sh_test( 
+        name = "my_test",
+        ...
+    )
+
+    remote_run(
+        name = "test-on-remote-machine",
+        target = ":my_test",
+        machines = ["test-machine-00.corp"],
+    )
+
+You can then run `bazel run :test-on-remote-machine` which will result in:
+  1) Target "my_test" being built, including all its dependencies.
+  2) The target, and all its dependencies, copied on the remote machine using
+     rysnc, in a directory tree that matches the bazel structure.
+  3) The target being run on the remote machine.
+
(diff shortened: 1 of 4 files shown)


bazel/utils/remote.bzl
```diff
@@ -0,0 +1,300 @@
...
+        destdir = shell.quote(attrs.destdir),
+        target = shell.quote(package(attrs.target.label)),
+        target_opts = shell.array_literal(attrs.target_opts),
+        executable = shell.quote(target_exec.short_path),
+        workspace = shell.quote(ctx.workspace_name),
+        rsync_opts = shell.array_literal(attrs.rsync_opts),
+        rsync_bin = shell.quote(attrs.rsync_bin),
+        ssh_opts = shell.array_literal(attrs.ssh_opts),
+        ssh_bin = shell.quote(attrs.ssh_bin),
+        machines = shell.array_literal(attrs.machines),
```
rajachan, line 245: Is there some way we can make this machine list available to the runner target as an environment variable / argument? The first machine takes on the execution role, but it would need access to a host list when running distributed jobs.

ccontavalli: added, great idea.

ccontavalli:
In TODO: 1) I want to add *tests* for these rules, which will allow me to fix some escaping issues (see TODO), but 2) want to try to use it as well, as I fear I'll need to change the API I bit.

Objections to merging this PR as is (already large), and address 1 and 2 in follow up PRs?

george-enf (approved):

rajachan (approved):

PR #679, opened 2022-06-28
bazel/linux: export the env. variable OUTPUT_DIR in the guest

george-enf:
Replace in the guest the hardcoded path /tmp/output_dir with the
environment variable OUTPUT_DIR.

### DIFF
--- a/bazel/linux/templates/init-qemu.template.sh
+++ b/bazel/linux/templates/init-qemu.template.sh
@@ -2,7 +2,7 @@
 
 echo ========= {message} - {target} ==========
 on_exit() {
-    echo $? > "/tmp/output_dir/exit_status_file" || true
+    echo $? > "$OUTPUT_DIR/exit_status_file" || true
     poweroff -f
 }
 trap on_exit EXIT
@@ -22,9 +22,10 @@ python -c 'import ctypes; exit(ctypes.cdll.LoadLibrary("libc.so.6").mount("", "/
 mount --types tmpfs tmpfs /tmp
 
 # Mount the output directory. This directory is shared with the host.
-mkdir /tmp/output_dir
+export OUTPUT_DIR=/tmp/output_dir
+mkdir "$OUTPUT_DIR"
 mount --types 9p \
     --options trans=virtio,version=9p2000.L,msize=5000000,cache=mmap,posixacl \
-    /dev/output_dir /tmp/output_dir
+    /dev/output_dir "$OUTPUT_DIR"
 
 {commands}
--- a/bazel/linux/test.bzl
+++ b/bazel/linux/test.bzl
@@ -70,7 +70,7 @@ echo '<?xml version="1.0" encoding="UTF-8"?>
 <testsuites>
 '"${junit}"'
 </testsuites>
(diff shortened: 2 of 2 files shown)


bazel/linux/templates/init-qemu.template.sh
```diff
@@ -22,9 +22,9 @@ python -c 'import ctypes; exit(ctypes.cdll.LoadLibrary("libc.so.6").mount("", "/
 mount --types tmpfs tmpfs /tmp
 
 # Mount the output directory. This directory is shared with the host.
-mkdir /tmp/output_dir
+mkdir $OUTPUT_DIR
```
ccontavalli, line 25: double quotes around variables? test with shellcheck even if template.

bazel/linux/templates/init-qemu.template.sh
```diff
@@ -22,9 +22,9 @@ python -c 'import ctypes; exit(ctypes.cdll.LoadLibrary("libc.so.6").mount("", "/
...
 mount --types 9p \
     --options trans=virtio,version=9p2000.L,msize=5000000,cache=mmap,posixacl \
-    /dev/output_dir /tmp/output_dir
+    /dev/output_dir $OUTPUT_DIR
```
ccontavalli, line 28: same as above

bazel/linux/qemu.bzl
```diff
@@ -49,6 +49,7 @@ fi
 
 QEMU_FLAGS+=("-fsdev" "local,security_model=none,multidevs=remap,id=fsdev-fsOutputDir,path=$OUTPUT_DIR")
 QEMU_FLAGS+=("-device" "virtio-9p-pci,fsdev=fsdev-fsOutputDir,mount_tag=/dev/output_dir")
+KERNEL_FLAGS+=("OUTPUT_DIR=/tmp/output_dir")
```
ccontavalli, line 52: how does this go from /proc/cmdline to env variable? don't see anything in init-qemu-template.sh?

george-enf: According to https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html:
> parameters with ‘=’ go into init’s environment

bazel/linux/templates/init-qemu.template.sh
george-enf: Fixed

bazel/linux/templates/init-qemu.template.sh
george-enf: Fixed

ccontavalli (approved):

bazel/linux/qemu.bzl
george-enf: After thinking a bit about your earlier comment, I realized that although it sounds interesting to pass the environment variable through the kernel command line, it is not really needed here. The location of the OUTPUT_DIR in the guest can be chosen arbitrarily by the guest. There is no need for the host to know or control the placement.

It's good to know that we can pass environment variables from the host to the guest, but in this case it really doesn't give any benefit.

ccontavalli: makes sense, so remove it? good point.

PR #682, opened 2022-06-30
gee restore_all_branches: make more interactive

jonathan-enf:
The typical user of "restore_all_branches" doesn't really want
to restore all the branches.  This change allows the user to
interactively select which branches to restore, if they have
the `dialog` utilty installed.

Alternately, a list of branches can be specified on the
commandline.

Tested: Works locally for me, also getting @abhienfabrica to help validate
a pre-release version.


jonathan-enf:

george-enf (approved):
My only worry here is if we can handle a branch name with a space. Hopefully, everything is escaped correctly.

PR #686, opened 2022-07-03
bazel/linux: add cleanup attribute to vm_bundle

george-enf:
Clean up operations execute outside of the emulator after the execution
of the emulator in reverse order.

Jira: https://enfabrica.atlassian.net/browse/SF-246

### DIFF
--- a/bazel/linux/bundles.bzl
+++ b/bazel/linux/bundles.bzl
@@ -25,6 +25,7 @@ def _vm_bundle(ctx):
     bundle = {}
     _add_attr_bundle(ctx, bundle, "prepare")
     _add_attr_bundle(ctx, bundle, "run")
+    _add_attr_bundle(ctx, bundle, "cleanup")
     _add_attr_bundle(ctx, bundle, "check")
 
     return RuntimeBundleInfo(**bundle)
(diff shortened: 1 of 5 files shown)


bazel/linux/runner.bzl
```diff
@@ -132,13 +136,14 @@ def create_runner(ctx, archs, code, runfiles = None, extra = {}):
             ),
         )
 
-    prepares, runs, checks = get_prepare_run_check(ctx, ctx.attr.run)
+    prepares, runs, cleanups, checks = get_prepare_run_check(ctx, ctx.attr.run)
```
ccontavalli, line 139: This is also used in bazel/linux/test.bzl, do you need to update it there as well?

bazel/linux/bundles.bzl
```diff
@@ -102,6 +107,17 @@ You can set up the bundle like this:
         "run_args": attr.string(
             doc = "Optional parameters to pass to the run_bin. Can use shell expansion.",
         ),
+        "cleanup_cmds": attr.string_list(
+            doc = "Shell commands to run OUTSIDE the VM AFTER the RUN (in reverse order) to clean up the environment (before cleanup_bin - optional)",
```
ccontavalli, line 111: can you expand the docstring here (or a better place? not sure) to explain the difference between check and cleanup?

On the top of my head:
- cleanups are ALWAYS run
- checks are only run if the tests in the VM complete successfully? and in non-interactive mode?

Otherwise, when would one use cleanup vs check?

george-enf: Fixed

bazel/linux/runner.bzl
george-enf: Good catch. Thanks for that!

ccontavalli (approved):

espeer-enfabrica (commented):
LGTM

PR #687, opened 2022-07-03
bazel/linux: fix kernel_module to define its runfiles

george-enf:
kernel_module should define its runfiles so that the following works:

kernel_module(name = "mod" ...)
sh_binary(name = "test", srcs = ["test.sh"], data = [":mod"])

Jira: https://enfabrica.atlassian.net/browse/SF-219

### DIFF
--- a/bazel/linux/defs.bzl
+++ b/bazel/linux/defs.bzl
@@ -126,7 +126,15 @@ def _kernel_modules(ctx):
             setup = ctx.attr.setup,
         ))
 
-    return [DefaultInfo(files = depset(outputs)), KernelBundleInfo(modules = bundled)]
+    return [
+        DefaultInfo(
+            files = depset(outputs),
+            runfiles = ctx.runfiles(files = outputs),
+        ),
+        KernelBundleInfo(
+            modules = bundled,
+        ),
+    ]
 
 kernel_modules_rule = rule(
     doc = """Builds kernel modules.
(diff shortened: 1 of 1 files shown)


bazel/linux/defs.bzl
```diff
@@ -126,7 +126,15 @@ def _kernel_modules(ctx):
             setup = ctx.attr.setup,
         ))
 
-    return [DefaultInfo(files = depset(outputs)), KernelBundleInfo(modules = bundled)]
+    return [
+        DefaultInfo(
+            files = depset(outputs),
+            runfiles = ctx.runfiles(files = outputs),
```
ccontavalli, line 132: is this a generic rule? Eg, if a binary depends on a target for data, only runfiles are brought in, and must be declared as runfiles? What triggers the requirement for those files to be runfiles?

(eg, should we recommend to do this for every rule we write going forward? Why is it necessary here, but not in other places?)

ccontavalli (approved):
a question below, but fine to merge as is.

bazel/linux/defs.bzl
rajachan: Both outputs and runfiles should be available as data for sh_binary targets from what I've seen. Maybe are outputs are not being enumerated properly with the kernel rules?

george-enf: I created a very simple test case:
BUILD.bazel
```
load("simple_rule.bzl", "simple_rule")
simple_rule(
    name = "simple_rule1",
)
sh_binary(
    name = "test1",
    srcs = ["test.sh"],
    data = [":simple_rule1"],
)
```
test.sh
```
find
```
simple_rule.bzl
```
def _simple_rule(ctx):
    outputs = []
    for i in range(1, 3):
        output = ctx.actions.declare_file("file" + str(i))
        outputs.append(output)
        ctx.actions.write(output, "foo" + str(i))
    return [
        DefaultInfo(
            files = depset(outputs),
            # runfiles = ctx.runfiles(files = outputs),
        ),
    ]
simple_rule = rule(
    implementation = _simple_rule,
)
```
And then, I do `bazel run test1`. I can only see `file1` and `file2` if I uncomment the `runfiles` above. It looks to me that `runfiles` can be a disjoint set to `files`.

ccontavalli (approved):

PR #689, opened 2022-07-05
bazel/utils/remote.bzl: adds test, fix existing features.

ccontavalli:
In this PR:
- add unit tests to the remote rule. This is a lot of lines,
  but actually very little logic.
- fix several problems:
  - pass kwargs from the macro
  - add an help screen
  - rename _bin parameters to _cmd, to indicate those can really be
    bash commands, not just paths to binaries.
  - have wrappers correctly override parameters.


bazel/utils/remote.bzl
```diff
@@ -133,69 +133,93 @@ directory that a wrapper requires by default to be used.
...
+    )),
+    ("noop", attr.label, dict(
         executable = True,
         cfg = "host",
         default = "@enkit//bazel/utils/remote:noop",
         doc = "If this target is used as wrapper or as a target, it is consider a noop operation. Useful when target or wrapper as specified via label_flag()",
-    ),
-    "rsync_bin": attr.string(
+    )),
+    ("rsync_cmd", attr.string, dict(
```
rajachan, line 151: Was thinking about the time it would take to make large copies.. instead of just assuming rsync, we can take a copy_cmd and copy_opts and allow a tool override to use rsync, scp, pdsh cp, mpifilutils, foobar etc later. Perhaps v3 enhancement.

rajachan (approved):
Looking good. Thanks for the comprehensive tests!

bazel/utils/remote.bzl
ccontavalli: let me rename the attributes, I think in practice that's already the case.

ccontavalli: I spoke too soon: the current shell script relies on a *file* listing all the files to be copied. So it passes `--files-from` to rsync. This is easy to change, simple template substitution. But unless you feel strongly, I'd rather to it in a separate PR.

rajachan: I don't, yeah let's do that in a follow up PR

PR #693, opened 2022-07-06
bazel/utils: cc_filter works only for shared libraries

george-enf:


bazel/utils/cc_filter.bzl
```diff
@@ -9,6 +9,8 @@ def _cc_filter(ctx):
         libraries = []
 
         for src_lib in src_linker_input.libraries:
+            if not src_lib.dynamic_library:
```
ccontavalli, line 12: update the docs to mention the constraint? Add a fail() or print() statement to warn about the problem?

PR #695, opened 2022-07-06
bazel/utils/remote: allow local copy, allow non-exec targets.

ccontavalli:
With this PR, remote_run is renamed to export_and_run, and can now:
- export a target to a local directory. This will copy the target
  and all its necessary dependencies. This is useful for example
  when there are generated files involved, and you want to use a
  non bazel build or tooling.
- export (or remote copy) a non-executable target. When a target
  is not executable, files are just copied.
- allow for detinations that contain a full path. So to copy files
  in a specific remote directory, one can use 'machine:/path/to/dir'.

Added corresponding tests.

### DIFF
--- a/bazel/utils/BUILD.bazel
+++ b/bazel/utils/BUILD.bazel
@@ -119,7 +119,7 @@ cc_binary(
 
 #### Actual tests
 remote_run(
-    name = "test-no-machines",
+    name = "test-no-dests",
     rsync_cmd = "./bazel/utils/save-argv test1.rsync",
     rsync_opts = [
         "--test_rsync_opt",
@@ -132,10 +132,10 @@ remote_run(
     tools = [":save-argv"],
 )
 
-# No machines are specified for a remote run - the command should fail!
+# No dests are specified for a remote run - the command should fail!
 exec_test(
     name = "test-no-machine-fails_test",
-    dep = ":test-no-machines",
+    dep = ":test-no-dests",
     must_fail = True,
 )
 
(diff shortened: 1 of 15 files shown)


bazel/utils/remote.bzl
```diff
@@ -236,21 +252,30 @@ def _remote_run_impl(ctx):
...
+          wrapper_exec = ctx.attr.wrapper[DefaultInfo].files_to_run.executable
+          wrapper_runfiles = ctx.attr.wrapper[DefaultInfo].default_runfiles
+          runfiles = runfiles.merge(wrapper_runfiles)
+          runfiles = runfiles.merge(ctx.runfiles(files = [wrapper_exec]))
+
+          target_opts = attrs.wrapper_opts + ["--", target_exec] + target_opts
+          target_exec = wrapper_exec.short_path
+    else:
+      no_execute = True
+      print("FILES", tdi.files.to_list())
```
george-enf, line 277: maybe forgotten?

george-enf (approved):

bazel/utils/BUILD.bazel
```diff
@@ -119,7 +119,7 @@ cc_binary(
 
 #### Actual tests
 remote_run(
-    name = "test-no-machines",
+    name = "test-no-dests",
     rsync_cmd = "./bazel/utils/save-argv test1.rsync",
     rsync_opts = [
         "--test_rsync_opt",
```
rajachan, line 125: We should default to `-q` quiet logs too. My stderr from 2 machines was chatty enough, it will get noisier with more hosts.

rajachan (commented):
Change looks good, but testing a bit with a target I have for this before approving...

rajachan:
Passing rsync args with `bazel run xxx -- "-r'-q'"` did not seem to work for me.

PR #699, opened 2022-07-08
gee: don't be so restrictive about what directory gee is run from

jonathan-enf:
This PR fixes a minor annoyance: previously, gee would refuse to run certain
commands if executed outside of a gee-controlled git repository.  In the
spirit of "do what I mean not what I say", gee will now switch to a valid
gee-controlled git repository before proceeding.

Thus, gee commands can continue to operate under the assumption that
they are running from a gee-controlled git repository, and the user
has one less failure mode.

Tested: Ran "gee lspr" from a variety of directories.

In ~/gee/enkit: shows PRs associated with the enkit repository.
In ~/gee/internal: shows PRs associated with the internal repository.
In ~/gee: shows PRs associated with the internal (default) repo.

Previously, all 3 operations would fail.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -576,6 +576,9 @@ function _startup_checks() {
     exit 1
   fi
 
+  # 1.5. Make sure we're not in one of bazel's weird symlink directories.
+  while [[ "$(pwd -P)" != "$(pwd -L)"  ]]; do cd ..; done
+
   # 2. Ensure that we are in a gee-controlled git repo.  If not,
   #    switch to the main branch of one.
   local GITDIR
--- a/scripts/gee
+++ b/scripts/gee
@@ -563,17 +563,35 @@ function _startup_checks() {
   # Ensure that (most) gee commands are run from within the gee repository,
   # otherwise strange things might happen (if the user's home directory is a
   # git repository, for instance).
-  local GITDIR
-  GITDIR="$(readlink -f "$("${GIT}" rev-parse --git-common-dir)")"
-  if [[ "${GITDIR}" != "${GEE_DIR}"/*/.git ]]; then
-    _warn "Error: refusing to run this gee command outside of the gee" \
-          "repository."
-    if [[ -d "${GEE_DIR}" ]]; then
-      _info "Use \"gcd\" to switch to a gee branch and try again."
-    else
-      _info "No gee repository found, run \"gee init\" to create one."
-    fi
+
+  # 1. Ensure that a gee repository exists.
+  if [[ ! -d "${GEE_DIR}/${REPO}" ]]; then
+    _info "Directory ${GEE_DIR}/${REPO} not found, run \"gee init\" to create."
     exit 1
+  fi
+  _set_main
+  if [[ ! -d "${GEE_DIR}/${REPO}/${MAIN}" ]]; then
+    _die "Directory ${GEE_DIR}/${REPO}/${MAIN} not found." \
+      "Somehow your ${MAIN} branch got removed?"
+    exit 1
+  fi
+
+  # 2. Ensure that we are in a gee-controlled git repo.  If not,
+  #    switch to the main branch of one.
+  local GITDIR
+  if ! GITDIR="$("${GIT}" rev-parse --git-common-dir 2>/dev/null )"; then
+    # not in a git directory
+    _set_main
+    _info "Switching to ${GEE_DIR}/${REPO}/${MAIN}"
+    cd "${GEE_DIR}/${REPO}/${MAIN}"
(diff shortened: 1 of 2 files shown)


jonathan-enf:

scripts/gee
```diff
@@ -563,17 +563,38 @@ function _startup_checks() {
...
+  fi
+  _set_main
+  if [[ ! -d "${GEE_DIR}/${REPO}/${MAIN}" ]]; then
+    _die "Directory ${GEE_DIR}/${REPO}/${MAIN} not found." \
+      "Somehow your ${MAIN} branch got removed?"
+    exit 1
+  fi
+
+  # 1.5. Make sure we're not in one of bazel's weird symlink directories.
+  while [[ "$(pwd -P)" != "$(pwd -L)"  ]]; do cd ..; done
```
eric-enf, line 580: I like the approach of walking back to find what repository I'm really in

scripts/gee
```diff
@@ -563,17 +563,38 @@ function _startup_checks() {
...
-    _warn "Error: refusing to run this gee command outside of the gee" \
-          "repository."
-    if [[ -d "${GEE_DIR}" ]]; then
-      _info "Use \"gcd\" to switch to a gee branch and try again."
-    else
-      _info "No gee repository found, run \"gee init\" to create one."
+  if ! GITDIR="$("${GIT}" rev-parse --git-common-dir 2>/dev/null )"; then
+    # not in a git directory
+    _set_main
+    _info "Switching to ${GEE_DIR}/${REPO}/${MAIN}"
```
eric-enf, line 588: I'm not sold on the idea of switching from some directory outside of git into an arbitrary branch/directory just so the git command can be run.  It's ok to to fail.  Many git commands don't make sense to run outside of a git repo.  Some commands are harder to undo than others.

eric-enf (approved):
Approved since at the end of the day, you can make the decision about how `gee` should work.

jonathan-enf:

scripts/gee
george-enf: I am not sure either what problem we are trying to solve here. If the user runs gee outside of the repository, what should be the expected behavior?

jonathan-enf: expected behavior should be: if gee can do something reasonable, it should.

gee is a wrapper around git and gh.  Some git and gh commands work outside of a git repository, others don't.  Rather than deal with all of the edge cases, gee previously would refuse to run if executed from outside a gee-controlled git repository.  But that meant that innocuous commands like `gee lspr` would fail unless the user `gcd master` first.

This PR saves the user a step -- if a command is run outside of a gee-controlled git repo, we normalize the environment by switching to the master branch of the gee-controlled repo, and then proceeding.  By eliminating this variable, our use of git and gh becomes more reliable.

eric-enf: You make a great point about "innocuous" commands.  What if you limit the behavior of switching into a git/gee directory to only innocuous commands?  Or only commands where gee "can do something reasonable?"

jonathan-enf: I guess I'm arguing that there aren't any gee commands for which this is harmful?  the only one I can think of is "gee commit", which will fail with a clear error message if executed from the master branch.

So: they're all innocuous.

jonathan-enf: Or perhaps put differently: _startup_checks is meant to run before (most) gee commands.  It's goal is to either fail with a clear error message if something in the environment isn't correct, or to "fail functional" (correct anything that needs correcting).

eric-enf: Oh, that's cool.  I didn't realize.  You have specific checks that prohibit commits to master?  That's a great feature!  

And presumably this works regardless of what a team decides to call their "MAIN" branch?  What about if we have multiple special branches such as if we moved to a branch based development model?

Commit is the easiest "harmful" command to come up with but it's hard to exhaustive.  What about `reset`?  In general I would think of "read-only" commands as the only innocuous ones.  For me it makes sense to require me to be IN the repo for anything that would modify the state of my repo.

eric-enf (approved):
Yeah, that seems great.

scripts/gee
jonathan-enf: gee does query the upstream repo metadata to get the correct name for the master branch, and falls back on "main" if all else fails.

so: even after all of my attempts to wriggle away from it, I ultimately decided that you're right, having a list of safe commands is the best option -- that way, if I add a future potentially unsafe command I'll have to think clearly about whether or not to add it to the "safe" list.  while testing, I also noticed a minor bug that was creating spurious warning messages from bash, and fixed that too.

george-enf (approved):

scripts/gee
eric-enf: You forgot:
```
40 profit!!!! 🤣
```

PR #702, opened 2022-07-09
astore: Move protos into package dirs

minor-fixes:
This is part of a larger effort to make the repository `go get`-clean.
This change moves protos around such that their import paths match their
actual package paths in the repository. Their BUILD files are cleaned up
such that `bazel run //:gazelle` creates no diffs, and gazelle doesn't
complain about ambiguous imports.

Tested:
* `bazel run //:gazelle` creates no diffs
* `bazel test //...` still works
* `go get -u ./...` no longer complains about astore/rpc/* packages


minor-fixes:
TODO before this leaves draft:
* [ ] Merge #700 
* [ ] Rebase

PR #703, opened 2022-07-09
astore: Move embed rules into package dir

minor-fixes:
This is part of a larger effort to make the repository `go-get`-clean.
This change moves rules that package embedded assets into a BUILD file in their
respective dir, and adds an empty source file to satisfy `go-get`.

Tested:
* `bazel test //...` still passes
* BUILD files are gazelle-clean
* `go get -u ./...` no longer complains about these packages


minor-fixes:
TODO before moving from draft:
* [ ] Merge #700 
* [ ] Merge #702 
* [ ] Rebase

PR #705, opened 2022-07-09
Make repo go get-clean

minor-fixes:
This change employs a couple tricks so that `go get -u ./...` can
successfully update `go.mod` and `go.sum`.

* For packages containing only proto-generated code, an empty source
  file is created to satisfy `go get`.
* For packages containing only go-embed generated code, the BUILD rules
  are moved to a BUILD file that reflects the import path, and an empty
  source file is created to satisfy `go get`.

The resulting changes are tested to be gazelle-clean; in some cases,
directives were added or left intact to prevent gazelle from
overzealously cleaning up rules. These directives can in some cases be
removed once rules_go is updated and we're using a non-deprecated way of
doing go embedding.

Tested:
* `bazel test //...` succeeds
* `bazel run //:gazelle` causes no diffs


minor-fixes:
TODO before this moves from draft:
* [ ] Merge #700 
* [ ] Merge #702

gpaussaenf (approved):
The bestie/ dir change is trivial. I glanced through the rest of the files and the changes look to be consistently applied.

PR #706, opened 2022-07-09
go: Update everything

minor-fixes:
This change updates:
* rules_go, to pick up a new version of go_googleapis
* the entire Go third-party dependency set, by running:
  * `go get -u ./...`
  * `bazel run //:gazelle_update_repos`

The new version of rules_go now warns about using `go_embed_data`, which
is deprecated; this can be fixed in a future change.

Tested: `bazel test //...` still works

PR #707, opened 2022-07-09
nomad: Add event listener 

minor-fixes:
This change adds a binary that listens to Nomad via its event stream
API, and exposes counts of events via metrics.

The impetus for this binary is to provide a more reliable counter of OOM
events, but it's possible we will want to count other events as well.

Tested: Manually, against actual Nomad installation with alloc that
reliably OOMs.


minor-fixes:
TODO before this moves from draft:
* [ ] Merge #700 
* [ ] Merge #702 
* [ ] Merge #705 
* [ ] Merge #706 
* [ ] Rebase

PR #708, opened 2022-07-11
astore_download fix: add default value for timeout

jonathan-enf:
Fixing presubmit failures blocking enkit integration with internal.  The
issue was that `astore_download_and_extract` was being called from an
unexpected context, and the "timeout" attribute wasn't being populated
in the ctx context.

Since `astore_download_and_extract` is a public function, it has to support
being invoked with arbitrary ctx contexts.  (Really, these kinds of
implementation functions should never be public.)

This PR adds "timeout" attributes at tthe call sites I could find, and adds
reasonable defaults to support the call sites I couldn't find
(fail-functional).

Tested: Ran the following test from internal/master:

```
blaze query --override_repository=enkit=${HOME}/gee/enkit/astore_timeout @testing-latest-kernel//:all
```

The above command fails without these changes, and passes with.

### DIFF
--- a/bazel/astore/defs.bzl
+++ b/bazel/astore/defs.bzl
@@ -127,7 +127,7 @@ With this rule, you can easily download
 files from an artifact store.""",
 )
 
-def astore_download_and_extract(ctx, digest, stripPrefix, path = None, uid = None):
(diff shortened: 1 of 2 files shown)


jonathan-enf:

minor-fixes (approved):

PR #712, opened 2022-07-14
bazel/remote: add support for copying input files.

ccontavalli:
Before this PR:
- the rule could only copy the files generated by a target
  and its run time dependencies.

After this PR:
- the rule can copy the files needed in order to build
  the target specified.

This is implemented by using aspects, and introducing a new inputs
parameter. Additionally, added ability to copy all deps, runtime
and non-runtime.


bazel/utils/remote.bzl
```diff
@@ -136,6 +160,56 @@ directory that a wrapper requires by default to be used.
...
+        direct.extend(locfiles)
+        allfiles.extend(locfiles)
+
+        attrvalue = getattr(ctx.rule.attr, attr)
+        if type(attrvalue) != "list":
+            attrvalue = [attrvalue]
+
+        for atarget in attrvalue:
+            if atarget and InputFiles in atarget:
+                print("FOUND INPUTS", atarget[InputFiles])
```
george-enf, line 199: debug print?

george-enf (approved):

PR #713, opened 2022-07-15
gee gcd: fix incorrect mapping of branch-to-dir

jonathan-enf:
@hanu-enf reported a bug in which "gcd master" was taking him to a feature
branch directory instead of master.  We root-caused the problem to gee's
parsing of the output of `git worktree list --porcelain` which produces
output like this if a branch is in a rebase-in-progress state:

```
worktree /home/hanu/gee/internal/master
HEAD b144e5c34ff7bd8f24717337bde4bfd03fc649d0
branch refs/heads/master

worktree /home/hanu/gee/internal/cmn_reg
HEAD b144e5c34ff7bd8f24717337bde4bfd03fc649d0
detached
```

`gee` was not parsing the `detached` line correctly.  This PR fixes the issue
by parsing the `detached` line and then falls back on a secondary technique to
find the branch name if the branch is in a rebase-in-progress state.

If all else fails, gee gives up and uses the basename of the worktree directory
as the branchname, which is almost always correct, but we prefer to use the
.git metadata as our ground truth whenever possible.

Tested: I put a branch into a rebase-in-progress state.  I observed the
failure.  Implemented workaround.  Ran `gee gcd` again for a variety of
directories and got the correct behavior.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -835,8 +835,10 @@ function _update_branch_to_worktree() {
   #   HEAD b144e5c34ff7bd8f24717337bde4bfd03fc649d0
   #   detached
   #
-  # For "detached" branches, we're stuck just using the directory
-  # name to infer the branch.
+  # For "detached" branches, we need to directly query .git meta
+  # data directories to determine the branch name.  This is a bit
+  # hacky, so I fall back on using the basename of the directory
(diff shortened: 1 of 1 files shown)


jonathan-enf:

ccontavalli (approved):

PR #714, opened 2022-07-16
enkit outputs: Take local tunnel port as flag

minor-fixes:
Currently, only one user can `enkit outputs mount` on a shared dev
machine, as when tunnels are required the local port used is hardcoded
to a single value.

This change moves the value to a flag, so that multiple people might be
able to use `enkit outputs mount` by modifying the flag value.

The flag is defaulted to the same value as the hardcoded value, to keep
this change non-breaking.

Jira: INFRA-1476

### DIFF
--- a/enkit/outputs/commands.go
+++ b/enkit/outputs/commands.go
@@ -25,10 +25,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var (
-	localTunnelPort = 8822
-)
-
 type Root struct {
 	*cobra.Command
 	*client.BaseFlags
@@ -38,6 +34,7 @@ type Root struct {
 	BuildBuddyUrl         string
 	BuildbarnHost         string
 	BuildbarnTunnelTarget string
+	TunnelListenPort      int
 	GatewayProxy          string
 }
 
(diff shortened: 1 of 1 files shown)


enkit/outputs/commands.go
```diff
@@ -109,10 +107,18 @@ func NewMount(root *Root) *Mount {
...
-func maybeSetupTunnel(hostPort string, tunnelTarget string, gatewayProxy string) (string, int, error) {
+//
+// More specifically, if a tunnel is not necessary, this returns the original
+// host:port (but split into host and port).
+// If a tunnel was necessary, this:
+// * creates the tunnel to tunnelTarget listening on tunnelPort on localhost
+// * returns the original host but changes the port to the tunnelPort
+// It is expected that the supplied host resolve to localhost, so that the
+// connection for the returned host and port combo jumps through the tunnel.
+func maybeSetupTunnel(hostPort string, tunnelTarget string, tunnelPort int, gatewayProxy string) (string, int, error) {
```
pete-enf, line 121: maybe is kinda weird function name... I cant really think of a better name though.

pete-enf (approved):
looks good.

PR #717, opened 2022-07-20
codegen: switch to libyaml parser

jonathan-enf:
A user was complaining that parsing a particular YAML file (`mrng_map.elab.yaml`, 18MB)
was taking minutes of build time, so I decided to look into it.  It turns out that
if I switch from using the native python yaml parser to the compiled libyaml library,
I get a 15x performance improvement with no apparent downside.

I added a benchmark for loading a medium-sized yaml file:

```
len(big_text)=19331
yaml version = 6.0
yaml.Loader: 4.944591138999385
yaml.SafeLoader: 4.906097150000278
yaml.CSafeLoader: 0.35437141400143446
data_loader: 0.39925295200009714
```

"data_loader" is the slightly customized version of yaml.CSafeLoader that codegen
now employs.

Tested: Re-ran all //tools/codegen/... tests.  I also added a small benchmark
to data_loader_test.py.

### DIFF
--- a/tools/codegen/BUILD.bazel
+++ b/tools/codegen/BUILD.bazel
@@ -30,6 +30,15 @@ py_test(
     ],
 )
 
+py_test(
+    name = "data_loader_benchmark",
+    srcs = ["data_loader_benchmark.py"],
+    deps = [
+        ":data_loader",
+    ],
+    tags = ["no-presubmit"],
+)
+
 py_binary(
     name = "codegen",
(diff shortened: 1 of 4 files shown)


jonathan-enf:

minor-fixes (approved):
`bazel test //tools/codegen:data_loader_test --runs_per_test=1000` fails 13 times out of 1000 on my machine, so consider reworking/removing the assertion

PR #719, opened 2022-07-20
gee repair: fix incorrect gh-resolved settings

jonathan-enf:
I'm not sure how, but for some users gh is configuring "origin" isntead
of "upstream" to be the base repo for managing PRs.  I still need to
root-cause why that happened to at least one user, but in the meanwhile
it's trivial to add a check to the "gee repair" command to detect the
condition and fix it.

I'm also changing the enkit agent environment variable check from looking
for SSH_AGENT_PID to SSH_AUTH_SOCK -- because in certain ssh-forwarded
environments, the latter will be set correctly even if the former isn't.

Tested: I manually broke my local git configuration, and then had gee
repair it.  I further verified that the gee repair command is harmless
in a healthy gee repo.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -416,8 +416,8 @@ function _set_ghuser() {
 
 function _check_ssh_agent() {
   # Check that the ssh-agent is loaded and reachable.
-  if [[ -z "${SSH_AGENT_PID}" ]]; then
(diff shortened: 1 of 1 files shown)


jonathan-enf:

ccontavalli (approved):

PR #720, opened 2022-07-21
bazel/linux: print qemu_test output to stdout

george-enf:
The output of commands executed by qemu_test is captured to a variable
and printed if the command fails. This behavior works fine with bazel
test which hides output by default. When the user executes a qemu_test
with bazel run, the expectation is to see the output in real time.

### DIFF
--- a/bazel/linux/test.bzl
+++ b/bazel/linux/test.bzl
@@ -34,7 +34,11 @@ run_test() {
     printf "%-73s" "${title}"
     start_time_ns=$(($(date +%s%N)))
 
-    output=$("${script}" "${args[@]}" 2>&1) && failed=0
+    # Print stdout and stderr to the terminal and capture it into a variable.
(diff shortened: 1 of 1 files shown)


bazel/linux/test.bzl
```diff
@@ -34,7 +34,8 @@ run_test() {
     printf "%-73s" "${title}"
     start_time_ns=$(($(date +%s%N)))
 
-    output=$("${script}" "${args[@]}" 2>&1) && failed=0
+    exec 5>&1
```
ccontavalli, line 37: add a comment here why it's necessary? I've had to read the commit message to understand what was happening here. Next guy looking at the code should not have to resort to tracking down the origin of the change through git blames :)

ccontavalli (approved):

bazel/linux/test.bzl
george-enf: good point. added a comment.

PR #723, opened 2022-07-29
bazel affected-targets: Fix target name bug

minor-fixes:
The external-target-attribute-rewrite logic has a bug where it captures
a loop variable by reference, causing targets to get renamed with the
incorrect name. This change makes a copy of the loop variable and
assigns a reference to the copy instead, so that it is not mutated as
the loop continues.

This fixes an issue where affected-targets will return extra targets that
aren't actually changed between the two revisions.

Tested: `bazel affected-targets list --loglevel-console=info --start=4b551678e --end=swc_copy`
with this change doesn't yield erroneous targets, whereas it used to
roughly ~20% of the time.

Jira: INFRA-1548


ccontavalli (approved):
wo! Good catch!

PR #725, opened 2022-07-30
enkit agent: use one path for all ssh-agent sockets

jonathan-enf:
Goal: When a user reconnects to a remote machine using ssh, we want
existing persistent terminals (ie. tmux, screen) to be able to access
the new socket.

To accomplish this, we always create (or update) a symlink to the current
ssh-agent socket, so that existing processes can always access the
latest and greatest instance of the agent.

* When starting a new ssh-agent, always use ~/.config/enkit/agent as the
  socket name.

* When discovering a running ssh-agent in the environment, create or
  update ~/.config/enkit/agent to be a symlink to the new socket.

Tested: Modified the regression test to use a mock version of
directory.GetConfigDir, tests pass.

manual testing:

```
$ SSH_AUTH_SOCK=/tmp/ssh-dalwl4pZhDch/agent.1029205; export SSH_AUTH_SOCK;

$ ssh-add -l
The agent has no identities.

$ ssh-add ~/.ssh/id_ed25519
id_ed25519      id_ed25519.pub

$ ssh-add ~/.ssh/id_ed25519
Identity added: /home/jonathan/.ssh/id_ed25519 (/home/jonathan/.ssh/id_ed25519)

$ ssh-add -l
256 SHA256:AbeM0ij/qiAgtXMVB+Eq6u4twnTqLWFawi9n+u1N2wQ /home/jonathan/.ssh/id_ed25519 (ED25519)

$ /home/jonathan/gee/enkit/enkit_agent2/bazel-bin/enkit/enkit_/enkit agent print
SSH_AUTH_SOCK=/home/jonathan/.config/enkit/agent; export SSH_AUTH_SOCK;

$ ls -l ~/.config/enkit/agent
lrwxrwxrwx 1 jonathan jonathan 35 Aug 31 19:11 /home/jonathan/.config/enkit/agent -> /tmp/ssh-dalwl4pZhDch/agent.1029205=

$ SSH_AUTH_SOCK=/home/jonathan/.config/enkit/agent ssh-add -l
256 SHA256:AbeM0ij/qiAgtXMVB+Eq6u4twnTqLWFawi9n+u1N2wQ /home/jonathan/.ssh/id_ed25519 (ED25519)

```

Further testing: installed new enkit in my container on build-06, and tried disconnecting
and reconnecting my ssh session, and observed the symlink get updated appropriately so that
my tmux session continued to be able to access my credentials.  Huzzah!


jonathan-enf:

lib/kcerts/ssh.go
```diff
@@ -100,6 +106,51 @@ func (a SSHAgent) Valid() bool {
...
+		if err != nil {
+			return err
+		}
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
+	if err := os.Chmod(a.Socket, 0700); err != nil {
+		return err
+	}
+	if err := os.Link(a.Socket, socket); err != nil {
```
ccontavalli, line 146: uhm .. this would fail on many many systems? hard links don't work across filesystems, /tmp is often a tmpfs, likely not on the same filesystem where ssh-agent creates the socket by default.

lib/kcerts/ssh.go
```diff
@@ -100,6 +106,51 @@ func (a SSHAgent) Valid() bool {
...
 }
 
+func (a *SSHAgent) UseStandardPaths() error {
+	user, err := UserCurrent()
+	if err != nil {
+		return err
+	}
+
+	// Make sure /tmp/ssh-${USER} exists with the right permissions
+	path := fmt.Sprintf("/tmp/ssh-%s", user.Username)
```
ccontavalli, line 116: This code makes me very uncomfortable, I don't think it's safe to create a file with a predictable name in /tmp, ever (although I see checks below that may prevent most problems), and code below looks racy. Eg, another user could create /tmp/ssh-jonathan, socket may be created in between the stat and remove, ...

ccontavalli: (why use /tmp, and not a dir created by the enkit cache, that is shared across containers?)

(also, enkit agent saves the agent information in the cache already, whatever the path is, it can re-use it, do a simple cd ~/.cache/enkit; find . -type f -name ssh.json)

ccontavalli: (also, see comment below on /tmp, I would not use hard links)

lib/kcerts/ssh.go
```diff
@@ -190,22 +249,48 @@ func FindSSHAgent(store cache.Store, logger logger.Logger) (*SSHAgent, error) {
...
 
 // CreateNewSSHAgent creates a new ssh agent. Its env variables have not been added to the shell. It does not maintain
 // its own connection.
 func CreateNewSSHAgent() (*SSHAgent, error) {
-	cmd := exec.Command("ssh-agent", "-s")
+	user, err := UserCurrent()
+	if err != nil {
+		return nil, err
+	}
+	path := fmt.Sprintf("/tmp/ssh-%s", user.Username)
```
ccontavalli, line 278: make this a method on the agent? the filepath/mkdir/etc as well, so it's re-used.

ccontavalli (commented):
In the PR description, could you add a why the change is necessary? I remember we talked about it, but lost the context. I can see the changes, I can't remember why we wanted those changes.

lib/kcerts/ssh.go
jonathan-enf: You should be uncomfortable!  This is security-critical code and many of our forebears have been crushed in these trenches.

1. Is it safe to create a directory with a predictable name / another user can create /tmp/ssh-jonathan?

Only if we can guarantee that the user has exclusive access to that directory, and that an attacker can't replace that directory with a fs node they can modify.  This process will fail if we cannot chmod /tmp/ssh-jonathan to 0700, but I should probably also add a check that the directory ends up with the right ownership and mode.

2. Race between stat and remove?

I don't think this is an issue as the directory is owned by the user and chmod 0700 at this point -- so nobody but the user can add or remove nodes in that directory at this point.  Failure to remove an existing socket node results errors out.

3. Why use /tmp?

Because sshd is already creating /tmp/ssh-XXXXX for the forwarded agent socket, and I wanted to keep the new name in the same directory / on the same filesystem.  If I switch to using a symlink, this is less of a requirement.

lib/kcerts/ssh.go
jonathan-enf: I thought I tried a symlink and had an issue, but maybe there was a user error in there.  I'll try again and see!  symlinks are definitely easier to inspect than hardlinks.

lib/kcerts/ssh.go
```diff
@@ -27,8 +29,9 @@ const (
 )
 
 var (
-	sockR = regexp.MustCompile("(?m)SSH_AUTH_SOCK=([^;\\n]*)")
-	pidR  = regexp.MustCompile("(?m)SSH_AGENT_PID=([0-9]*)")
+	sockR       = regexp.MustCompile("(?m)SSH_AUTH_SOCK=([^;\\n]*)")
+	pidR        = regexp.MustCompile("(?m)SSH_AGENT_PID=([0-9]*)")
+	UserCurrent = user.Current // to enable mocking
```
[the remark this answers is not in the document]
jonathan-enf: If I make UserCurrent private, is there an easy way for ssh_test.go to have access to it?  I feel like I would at least need a public setter function, unless there's a go testing technique that I'm missing?

lib/kcerts/ssh.go
jonathan-enf: Will do.

lib/kcerts/ssh.go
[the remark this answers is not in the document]
jonathan-enf: I tried making the package the same, but still got `lib/kcerts/ssh_test.go:41:9: cannot refer to unexported name kcerts.userCurrent` compilation error.  Shrug!  Left public.

lib/kcerts/ssh.go
jonathan-enf: Now using a symlink.

lib/kcerts/ssh.go
jonathan-enf: Done.

lib/kcerts/ssh.go
```diff
@@ -103,6 +106,57 @@ func (a SSHAgent) Valid() bool {
...
+	if err != nil {
+		return err
+	}
+
+	if a.Socket == socket {
+		// no standardization needed.
+		return nil
+	}
+
+	// Securely rename socket:
```
ccontavalli, line 139: there's a race condition, and as per comment above, I think we don't need to be as paranoid to write in the home directory of the user. If others can write there/predict a path, there are bigger problems.

I would either:
1) Just create the symlink (remove old one, if necessary)
2) Or rename in place (so it's atomic).

Something like:
```
tempname := fmt.Sprintf("%s.tmp%016x", rand.New(srand.Source).Uint64())
defer os.Remove(tempname)
if err := os.Symlink(a.Socket, tempname); err != nil {
...
}
if err := os.Rename(tempname, socket); err != nil {
...
}
```

lib/kcerts/ssh.go
```diff
@@ -173,26 +227,68 @@ func (a SSHAgent) GetEnv() []string {
...
+//   4) ensures the valid socket has the right path, otherwise does a symlink.
+//
+// The final ssh-agent socket returned by FindSSHAgent is always
+// ~/.config/enkit/agent.
 func FindSSHAgent(store cache.Store, logger logger.Logger) (*SSHAgent, error) {
+	var err error
 	agent := FindSSHAgentFromEnv()
-	if agent != nil && agent.Valid() {
-		return agent, nil
+	if agent != nil {
```
ccontavalli, line 245: nit: `if agent != nil && !agent.Valid() {`

lib/kcerts/ssh.go
```diff
@@ -173,26 +227,68 @@ func (a SSHAgent) GetEnv() []string {
...
+		if !agent.Valid() {
+			logger.Warnf("%s from env isn't a valid ssh-agent socket.", agent.Socket)
+			agent = nil
+		}
 	}
-	agent, err := FetchSSHAgentFromCache(store)
-	if err != nil {
-		logger.Warnf("%s", err)
+
+	if agent == nil {
```
ccontavalli, line 252: the inverted `if agent == nil` logic is killing me :/ How about we break down `FindSSHAgent` in 2 functions:
1) `FindOrCreateAgent()` -> that just finds the agent, either from env, cache, or creates a new one? And is pretty much the old unmodified `FindSSHAgent` except it does NOT write to cache.
2) `FindSSHAgent` now invokes `FindOrCreateAgent`, and then does `UseStandardPaths` and `WriteAgentToCache`?

Bonus points if we rename this to `PrepareSSHAgent` or something that's more meaningful than `FindSSHAgent` - which is a misnomer at this point.

lib/kcerts/ssh.go
```diff
@@ -219,7 +315,21 @@ func FindSSHAgentFromEnv() *SSHAgent {
...
+	}
+
+	os.Remove(socket) // ignore errors
+	_, err = os.Stat(socket)
+	if err == nil {
+		return nil, fmt.Errorf("unable to delete existing socket: %s", socket)
+	}
+
+	// TODO(jonathan): is it necessary or safer to delete the socket first?
+	cmd := exec.Command("ssh-agent", "-s", "-a", socket)
```
ccontavalli, line 332: now that we use a symlink, why not invoke `ssh-agent` as usual? I'm worried that otherwise we have two different "landing" ssh-agent configs. In some cases, our socket will be a symlink, in some cases, it will be directly the socket.

But... one configuration will work well in some cases, and not in others, and vice-versa, we'll pretty much end up having to debug and ensure that all programs work in all cases "oh, he's only seeing this problem when it's a symlink - that code is hardcoding the agent path to be in /tmp, or ... it doesn't allow for symlink or ...".

ccontavalli: (and there's no race conditions around socket deletion/removal and agent starting)

lib/kcerts/ssh_test.go
```diff
@@ -37,11 +37,20 @@ func TestAddSSHCAToClient(t *testing.T) {
 
 // TODO(adam): test cache failures and edge cases
 func TestStartSSHAgent(t *testing.T) {
+	tmpDir, err := ioutil.TempDir("", "en")
+	assert.Nil(t, err)
+	os.Mkdir(tmpDir+"/.config", 0700)
+	os.Mkdir(tmpDir+"/.config/enkit", 0700)
```
ccontavalli, line 43: both Mkdir should not be necessary, given our code above? which invokes mkdirall?

lib/kcerts/ssh_test.go
```diff
@@ -37,11 +37,20 @@ func TestAddSSHCAToClient(t *testing.T) {
 
 // TODO(adam): test cache failures and edge cases
 func TestStartSSHAgent(t *testing.T) {
+	tmpDir, err := ioutil.TempDir("", "en")
+	assert.Nil(t, err)
```
ccontavalli, line 41: assert.NoError is generally preferred.

lib/kcerts/ssh.go
jonathan-enf: Done

lib/kcerts/ssh.go
```diff
@@ -103,6 +106,57 @@ func (a SSHAgent) Valid() bool {
 	return err == nil
 }
 
+func (a *SSHAgent) GetStandardSocketPath() (string, error) {
+	path, err := GetConfigDir("enkit")
+	if err != nil {
+		return "", err
+	}
+
+	// Securely make sure path exists and is permission 0700
```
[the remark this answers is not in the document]
jonathan-enf: Done.

lib/kcerts/ssh.go
```diff
@@ -103,6 +106,57 @@ func (a SSHAgent) Valid() bool {
...
+	if err := os.MkdirAll(path, 0700); err != nil && !os.IsExist(err) {
+		return "", err
+	}
+	if err := os.Chmod(path, 0700); err != nil {
```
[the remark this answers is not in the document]
jonathan-enf: Done.

lib/kcerts/ssh.go
```diff
@@ -103,6 +106,57 @@ func (a SSHAgent) Valid() bool {
...
+		return "", err
+	}
+
+	socket := filepath.Join(path, "agent")
+
```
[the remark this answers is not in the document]
jonathan-enf: Done.

lib/kcerts/ssh.go
jonathan-enf: Makes sense.  Done.

lib/kcerts/ssh.go
jonathan-enf: Obsolete.

lib/kcerts/ssh.go
jonathan-enf: Done.

lib/kcerts/ssh_test.go
jonathan-enf: Done.

lib/kcerts/ssh_test.go
jonathan-enf: Removed.

jonathan-enf (commented):
Accepted all suggestions, PTAL.

proxy/ptunnel/commands/agent_test.go
```diff
@@ -25,6 +34,14 @@ func TestRunAgentCommand(t *testing.T) {
 }
 
 func TestRunAgentCommand_Error(t *testing.T) {
+	tmpDir, err := ioutil.TempDir("", "en")
```
minor-fixes, line 37: Enough instances of this to be a testutil func somewhere?

```
// go_library marked with `testonly = True`
package kcertmock

func ConfigDirForTestFunc(t *testing.T) func(string, ...string) (string, error) {
  t.Helper()
  tmpDir, err := ioutil.TempDir("", "en")
  assert.NoError(t, err)
  return func (string, ...string) (string, error) {
    return filepath.Join(tmpDir, ".config/enkit"), nil
  }
}
```

test code:

```
import (
  "github.com/enfabrica/enkit/lib/kcert/kcertmock" // or whatever
  "github.com/prashantv/gostub"
)

// ...

stubs := gostub.Stub(&kcerts.GetConfigDir, kcertmock.ConfigDirForTestFunc(t))
defer stubs.Reset()
```

minor-fixes: ...although, replacing a public function in a package you happen to know is a dep seems a bit unorthodox. I'm used to seeing functions that need to be mocked in this way in the tests for the same package only, and the var can be private as a result. In dependent packages - usually one has the opportunity to inject the dependency, via a func var or defining an interface and passing either a "real" impl or "test" impl

lib/kcerts/ssh.go
```diff
@@ -191,8 +241,33 @@ func FindSSHAgent(store cache.Store, logger logger.Logger) (*SSHAgent, error) {
...
+// either by discovering an existing ssh-agent or creating a new one.
+// It also ensures that we have an up-to-date symlink to that agent's
+// socket in the standard location.
+//
+// The final ssh-agent socket returned by PrepareSSHAgent is always
+// ~/.config/enkit/agent.
+func PrepareSSHAgent(store cache.Store, logger logger.Logger) (*SSHAgent, error) {
+	agent, err := FindOrCreateSSHAgent(store, logger)
+	if err != nil {
+		return nil, err
```
minor-fixes, line 260: Many unannotated errors returned through this code. Adding on some form of context can provide a sort of "poor man's backtrace" and make it easier to understand at a glance what happened/find the precise codepath that resulted in the error:

```
return nil, fmt.Errorf("while attaching to SSH agent: %w", err)
```

can result in:

```
while attaching to SSH agent: while creating new SSH agent: unable to delete existing socket: foo
```

instead of just the last part, passed all the way up

minor-fixes (approved):
Non-blocking opinions, since they are the beginning of the long, dark descent into perfectionist pedantry. Thanks for the fix! I'm excited to get this in so it can make an imminent release of enkit.

lib/kcerts/ssh.go
```diff
@@ -103,6 +107,52 @@ func (a SSHAgent) Valid() bool {
...
+	if err != nil {
+		return err
+	}
+	tempname := fmt.Sprintf("%s/enkit.tmp%016x", path, mathrand.New(srand.Source).Uint64())
+	defer os.Remove(tempname)
+	if err := os.Symlink(a.Socket, tempname); err != nil {
+		return err
+	}
+	// Rename symlink to the standard name
+	os.Remove(socket) // if it exists
```
ccontavalli, line 147: nit: `os.Remove` can go? `os.Rename` behaves like the syscall, "Rename renames (moves) oldpath to newpath. If newpath already exists and is not a directory, Rename replaces it." https://pkg.go.dev/os#Rename

(that's how it avoids race conditions - this is small and minor here, so not a big deal, but we should remove it)

lib/kcerts/ssh.go
```diff
@@ -219,6 +294,19 @@ func FindSSHAgentFromEnv() *SSHAgent {
...
 // its own connection.
 func CreateNewSSHAgent() (*SSHAgent, error) {
+	a := &SSHAgent{}
+
+	socket, err := a.GetStandardSocketPath()
+	if err != nil {
+		return nil, err
+	}
+
+	os.Remove(socket) // ignore errors
```
ccontavalli, line 304: this can also go? now that's a symlink? eg, ssh-agent will use a random path anyway, so no reason to check if the symlink/old path is in place here?

PR #726, opened 2022-08-01
bestie: Handle errors in test xml

jacob-adelmann:
* Updates bestie to handle `<error message="some error message">`
* Fixes issue where testcases that did not pass due to setup error were counted as passing
* Prior to this change, if a test case raised error in a setUp() class, the resultant test xml would have >0 error count, but since the error properties weren't handled they were being treated as passes.

### DIFF
--- a/bestie/server/xml_result.go
+++ b/bestie/server/xml_result.go
@@ -49,6 +49,7 @@ type TestCase struct {
 	Time       string     `xml:"time,attr"`
 	Properties Properties `xml:"properties"`
 	Failure    FailureMsg `xml:"failure"`
+	Error      ErrorMsg   `xml:"error"`
 	Skipped    SkippedMsg `xml:"skipped"`
 }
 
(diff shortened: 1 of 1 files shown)


bestie/server/xml_result.go
```diff
@@ -220,6 +226,8 @@ func parseStructuredXml(ts *TestSuite) ([]*xmlResult, error) {
 			xr.result = "fail"
 		} else if len(tc.Skipped.Message) > 0 {
 			xr.result = "skip"
+		} else if len(tc.Error.Message) > 0 {
+			xr.result = "error"
```
gpaussaenf, line 230: Please update the comment on line 224 to include 'error' in the list of sections mentioned.

gpaussaenf (approved):
I tried these changes on a localhost instance of the bestie server and they worked as intended.

bestie/server/xml_result.go
jacob-adelmann: Updated in latest

minor-fixes (approved):

PR #728, opened 2022-08-04
gee gcd: redirect savelog output to stderr

jonathan-enf:
Just a quick bugfix: gee periodically invokes `savelog` to rotate it's log
file.  `savelog` sometimes writes to stdout, which then interferes with
the `gee gcd` command, which requires that the only thing written to
stdout is supposed to be the generated path.

This PR quick-fixes the behavior so that savelog triggering a log rotation
won't interfere with the functioning of gcd.

Tested: inflated ~/gee.log to be larger than 64K, forcing a log rotation
on the next gee invocation, and then ran `gee gcd master > /tmp/foo.txt`
to capture the stdout of gee under those conditions.  Confirmed that
savelog output no longer pollutes our results.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -266,7 +266,7 @@ function _rotate_gee_log() {
     local size
     size="$(stat -c%s "${GEE_LOG_FILE}")"
     if (( size > 65536 )); then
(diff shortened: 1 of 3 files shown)


jonathan-enf:

george-enf (approved):

PR #729, opened 2022-08-06
enproxy: Allow forwarding of user info via headers

minor-fixes:
This change adds transform options to enproxy mappings to tell enproxy
to optionally add some headers on forwarded requests that signal the
authenticated user's identity to the backend.

Each value of the `Identity` object associated with the user is exposed
in its own header, and then an email-like string is exposed as well.

This is inspired by Grafana, which has a special ["Auth Proxy"](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/auth-proxy/)
authentication mode, whereby it will respect the values of particular
headers as a pre-authenticated user.

Since these header values are configurable within Grafana, there is no
need to stick to their default-selected `X-WEBAUTH-USER`; we can choose
our own names and configure Grafana appropriately. This PR selects names
that map more closely to the underlying field from which it is
populated, in an attempt to minimize confusion caused by additional
renames.

This will also be useful in setting up API authentication in Airflow,
which allows us to inject some Python code as a [custom auth backend](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/auth-proxy/).
Writing code that examines these headers and matches them to current
users should be trivial by following [existing examples](https://sourcegraph.com/github.com/apache/airflow@8070a781b6292ce0795ceed6f36217dd61023dad/-/blob/airflow/providers/google/common/auth_backend/google_openid.py?L110)

Tested: no

### DIFF
--- a/proxy/httpp/interface.go
+++ b/proxy/httpp/interface.go
@@ -1,11 +1,14 @@
 package httpp
 
 import (
-	"github.com/enfabrica/enkit/lib/khttp"
 	"net/http"
 	"net/url"
 	"regexp"
 	"strings"
+
+	"github.com/enfabrica/enkit/lib/logger"
+	"github.com/enfabrica/enkit/lib/khttp"
+	"github.com/enfabrica/enkit/lib/oauth"
 )
 
 type HostPath struct {
@@ -34,9 +37,22 @@ const (
 	// Default behavior: ignore X-Forwarded-For from the client, but provide one to the backend.
 	XForwardedForSet XForwardedForTreatment = "set"
 	// X-Forwarded-For header is forwarded to the backend, with the ip of the client added.
-	XForwardedForAdd = "add"
+	XForwardedForAdd XForwardedForTreatment = "add"
 	// No X-Forwarded-For supplied to the backend. If there is one, it is stripped.
-	XForwardedForNone = "none"
+	XForwardedForNone XForwardedForTreatment = "none"
+)
+
+type XWebauthTreatment string
+
+const (
+	// Default behavior: set no additional headers on the forwarded request.
+	XWebauthNone XWebauthTreatment = "none"
+	// Sets:
+	//   * X-Webauth-Userid to the user's unique ID
+	//   * X-Webauth-Username to the user's username
+	//   * X-Webauth-Organization to the user's organization
+	//   * X-Webauth-Globalname to the user's "global name" (see oauth.(*Identity).GlobalName)
+	XWebauthSet XWebauthTreatment = "set"
 )
 
 type Transform struct {
@@ -50,6 +66,9 @@ type Transform struct {
 
 	// Defines what to do with the X-Forwarded-For header, and the IP of the client.
 	XForwardedFor XForwardedForTreatment
+	// Defines whether to set additional headers signaling the user's identity on
+	// the request to the backend.
+	XWebauth XWebauthTreatment
 	// List of regular expressions defining which cookies to strip in requests to the backend.
 	StripCookie []string
 	// By default, requests to the backend are forwarded with the Host field set to the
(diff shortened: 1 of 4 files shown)


proxy/httpp/interface.go
```diff
@@ -39,6 +41,19 @@ const (
...
+
+const (
+	// Default behavior: set no additional headers on the forwarded request.
+	XWebauthNone XWebauthTreatment = "none"
+	// Sets:
+	//   * X-Webauth-Userid to the user's unique ID
+	//   * X-Webauth-Username to the user's username
+	//   * X-Webauth-Organization to the user's organization
+	//   * X-Webauth-Globalname to the user's "global name" (see oauth.(*Identity).GlobalName)
+	XWebauthSet = "set"
```
mhallmark, line 54: did u mean to have this set as an untyped string instead of `XWebauthTreatment`?

proxy/httpp/interface.go
```diff
@@ -120,6 +138,24 @@ func (t *Transform) Apply(req *http.Request) bool {
...
+
+	switch t.XWebauth {
+	case "":
+		fallthrough
+	case XWebauthNone:
+		// No extra headers set
+	case XWebauthSet:
+		creds := oauth.GetCredentials(req.Context())
+		if creds == nil {
+			// TODO: any opportunity to log here? Metrics?
```
mhallmark, line 150: can we get a jira ticket for this - this will hide issues and we really should get some sort of logging here for this - or refactor this method to return (bool,error)

mhallmark (approved):

proxy/httpp/interface.go
minor-fixes: It's still typed - only the first constant in a block of constants need to specify type: https://www.sohamkamani.com/golang/enums/#enum-types

proxy/httpp/interface.go
minor-fixes: yeah I'm hoping @ccontavalli can clue me into if a logger/metrics is accessible here? and I can fix it in this PR

proxy/httpp/interface.go
minor-fixes: Actually, I'm wrong, and the blog post is wrong, and the code above is wrong: https://go.dev/play/p/YaJgdoD7jgR

Fixed here and above

proxy/httpp/interface.go
ccontavalli: Need to extend the Transform object to carry metrics/logs, or (arguably better?) pass it down to Apply? extend contract of Apply so it can eg, fail, or log or increment counters.

That said, if the mapping has authentication required, you'll never get here without credentials being available? And if mapping does NOT have authentication (marked as public), then having WebAuth set to Set should fail config validation, and return an error there? So we can never end up silently in this condition? This PR is still a good step in the right direction.

ccontavalli (approved):

proxy/httpp/interface.go
minor-fixes: Passed the logger in via context instead - so it gets plumbed from Proxy -> AuthenticatedProxy, which puts it in the context of each request. Fetching the logger from the context returns a NilLogger if none is present - so a fallback to noop/safe behavior.

minor-fixes:
Presubmit is timing out (https://github.com/enfabrica/internal/pull/7517 has a fix) but in the meantime I've built/tested all targets listed on https://console.cloud.google.com/cloud-build/builds/22b6b7f2-38f4-431a-82df-173b7803609f;step=3?project=cloud-build-290921 successfully.

proxy/httpp/interface.go
ccontavalli: (good catch, btw, I suspect I had started with a iota here and then changed)

PR #731, opened 2022-08-09
gee: fix enkit auth invocation

jonathan-enf:
While it appears that `enkit auth` is capable of inferring the username
to use during authentication, it turns out that it just caches this
information after the user does it for the first time.  We want gee
to work correctly on a completely clean install, and so the safest
option is to always specify the full `enkit auth $USER@enfabrica.net`
commandline.

Tested: rm ~/.config/enkit/* and then run `gee init`

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -444,7 +444,7 @@ function _check_enkit_cert() {
   if (( COUNT == 0 )); then
     _warn "enkit certificate is expired."
     _info "Please authenticate:"
-    "${ENKIT}" login
+    "${ENKIT}" login "${USER}@enfabrica.net"
     COUNT="$("${ENKIT}" agent print | wc -l)"
     if (( COUNT == 0 )); then
       _fatal "No enkit certificate, aborting."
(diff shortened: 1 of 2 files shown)


jonathan-enf:

scripts/gee
```diff
@@ -444,7 +444,7 @@ function _check_enkit_cert() {
   if (( COUNT == 0 )); then
     _warn "enkit certificate is expired."
     _info "Please authenticate:"
-    "${ENKIT}" login
```
george-enf, line 447: The change from `login` to `auth` here is deliberate?

jonathan-enf: yep -- login and auth are aliases, and I thought it was better to pick one and be consistent (made it easier to grep my code).

george-enf: Makes sense.

george-enf (approved):

george-enf:
nit: but I assume you wanted your commit message to say `$USER` instead of your username.

jonathan-enf:
> nit: but I assume you wanted your commit message to say `$USER` instead of your username.

exactly correct.  git commit -m ".. $USER@..."

PR #734, opened 2022-08-10
data_loader_benchmark: also benchmark json.

jonathan-enf:
Interestingly, python's json parser is about 100x faster than even the
C-based libyaml parser.  Captured for future consideration.

Tested: Re-ran the benchmark.


jonathan-enf:

george-enf (approved):

PR #744, opened 2022-08-27
enkit tunnel: Support listening on Unix Domain Sockets

minor-fixes:
This change extends the functionality of the `-L`/`--listen` flag on
`enkit tunnel` to support Unix Domain Socket URLs as listen targets, in
the form `unix:///path/to/socket`. This involves:

* creating a function to contain the logic that normalizes the contents
  of the flag. The flag needs to continue to support `host:port`, as
  well as just a naked `port`, in addition to the new UDS path.
* fixing dependent code to not assume `TCPAddr`/`TCPConn`. In some
  places, type assertions are necessary, as TCP and Unix variants
  contain necessary methods not in the top-level interface; these should
  be relatively safe as long as we don't expand the network types to
  types that lack these methods.

Additionally, cleanup logic is improved by creating a Context that
becomes done on SIGINT, and plumbing that Context to relevant locations.
Care is taken to ensure that the listener is cleaned up at the
appropriate time depending on the listener's type and whether the tunnel
is running in background mode.

Tested: For each following case, ran `grpc_cli ls <endpoint>` to ensure that the tunnel was operational:
* TCP tunnel in background mode: `bazel run //enkit -- tunnel --proxy=https://gw.gcp01.corp.enfabrica.net --background -L 10000 buildbarn-stage-frontend.service.consul. 13011`
* TCP tunnel in foreground mode: `bazel run //enkit -- tunnel --proxy=https://gw.gcp01.corp.enfabrica.net -L 10000 buildbarn-stage-frontend.service.consul. 13011`
* UDS tunnel in background mode: `bazel run //enkit -- tunnel --proxy=https://gw.gcp01.corp.enfabrica.net --background -L unix:///tmp/enkit_test.sock buildbarn-stage-frontend.service.consul. 13011` (after `kill -2 <pid>`, confirmed socket file is deleted)
* UDS tunnel in foreground mode: `bazel run //enkit -- tunnel --proxy=https://gw.gcp01.corp.enfabrica.net -L unix:///tmp/enkit_test.sock buildbarn-stage-frontend.service.consul. 13011` (after `Ctrl-C`, confirmed socket file is deleted)

Jira: INFRA-1505
Jira: INFRA-1460

### DIFF
--- a/proxy/ptunnel/commands/tunnel.go
+++ b/proxy/ptunnel/commands/tunnel.go
@@ -1,30 +1,33 @@
 package commands
 
 import (
+	"context"
 	"errors"
 	"fmt"
-	"github.com/enfabrica/enkit/lib/client"
-	"github.com/enfabrica/enkit/lib/goroutine"
-	"github.com/enfabrica/enkit/lib/kflags"
-	"github.com/enfabrica/enkit/lib/kflags/kcobra"
-	"github.com/enfabrica/enkit/lib/khttp"
-	"github.com/enfabrica/enkit/lib/khttp/krequest"
-	"github.com/enfabrica/enkit/lib/khttp/protocol"
-	"github.com/enfabrica/enkit/lib/knetwork"
-	"github.com/enfabrica/enkit/lib/retry"
-	"github.com/enfabrica/enkit/proxy/nasshp"
-	"github.com/enfabrica/enkit/proxy/ptunnel"
-	"github.com/spf13/cobra"
 	"io"
 	"net"
 	"net/http"
 	"net/url"
 	"os"
 	"os/exec"
+	"os/signal"
 	"os/user"
 	"strconv"
 	"strings"
 	"syscall"
+
+	"github.com/enfabrica/enkit/lib/client"
+	"github.com/enfabrica/enkit/lib/goroutine"
+	"github.com/enfabrica/enkit/lib/kflags"
+	"github.com/enfabrica/enkit/lib/kflags/kcobra"
+	"github.com/enfabrica/enkit/lib/khttp/krequest"
+	"github.com/enfabrica/enkit/lib/khttp/protocol"
+	"github.com/enfabrica/enkit/lib/knetwork"
+	"github.com/enfabrica/enkit/lib/retry"
+	"github.com/enfabrica/enkit/proxy/nasshp"
+	"github.com/enfabrica/enkit/proxy/ptunnel"
+
+	"github.com/spf13/cobra"
 )
 
 type Tunnel struct {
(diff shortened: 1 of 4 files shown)


proxy/ptunnel/commands/tunnel.go
```diff
@@ -138,24 +182,69 @@ func (r *Tunnel) Run(cmd *cobra.Command, args []string) (err error) {
...
+// closing the returned channel and terminating.
+func listenerChan(l net.Listener, e *error) <-chan net.Conn {
+	c := make(chan net.Conn)
+	go func() {
+		defer close(c)
+		for {
+			conn, err := l.Accept()
+			if err != nil {
+				*e = err
+				return
```
ccontavalli, line 196: errors from accept can be transient (example: not enough file descriptors), I don't think you should always exit the goroutine? Generally you want to keep the listener running in that case.

That said, the error is reported correctly, so I don't feel strongly about adding more complexity here.

proxy/ptunnel/commands/tunnel.go
```diff
@@ -182,42 +271,65 @@ func IsBackgroundFork() bool {
...
 		if r.Background {
-			err := r.RunBackground(tlisten)
-			tlisten.Close()
+			err := r.RunBackground(listener)
+			// If using a UNIX domain socket, calling Close() here unlinks it
+			// from the FS. The forked process will still be able to listen
+			// since it has an open FD, but no other process can dial it, which
+			// kinda defeats the purpose.
+			//
+			// Instead, let the forked process clean up when it is signalled.
```
ccontavalli, line 321: should this just delete the file anyway? RunBackground failed, it's probably "correct" that we no longer use that failed background process? and allow another enkit to bind on the same tunnel?

(the forked child will eventually die, but at least other children can now start?)

ccontavalli: (the most annoying part about UDS is that if the program that created one crashes/dies, the file is left behind, and trying to bind on it now returns EADDRINUSE - so it's generally better to remove/close aggressively than risk leaving files behind)

proxy/ptunnel/commands/tunnel.go
```diff
@@ -138,24 +182,69 @@ func (r *Tunnel) Run(cmd *cobra.Command, args []string) (err error) {
...
-		}()
+			// The listener will report an error from Accept because we closed
+			// it, but this is expected and can be ignored.
+			return nil
+
+		case conn, ok := <-conns:
+			// Listener loop closed the channel unexpectedly, so no more
+			// connections, and there is probably an error.
+			if !ok {
+				return listenerErr
```
ccontavalli, line 225: on Done, listener.Close() is called. Should the same happen here? Eg, if Accept() returns error, channel is closed, we end up here, but nobody cleans/closes the listener?

proxy/ptunnel/commands/tunnel.go
ccontavalli: (except that when the background process dies, it may remove the file descriptor another process created ... ugh. Internet wisdom uses flock, but seems like overkill here - use your best judgement after testing)

ccontavalli (approved):
main concern is a) making sure the uds file is removed from disk if process dies/crashes, and 2) making sure a good error is printed on screen if the uds already exists (eg: can't start, uds file ... path of the file ... already exists) - so one can cleanup manually worst case.

proxy/ptunnel/commands/tunnel.go
minor-fixes: Clarified the error when using a socket and getting EADDRINUSE. I tried to make the cleanup automatic via some mechanism with flock, but this was nontrivial and makes this code even more branchy. Manual cleanup should be fine for now to reduce the risk of adding nastier bugs (deadlocks, etc.)

proxy/ptunnel/commands/tunnel.go
minor-fixes: ack

proxy/ptunnel/commands/tunnel.go
minor-fixes: Also added a Close() if RunBackground fails (but not if it succeeds, as per the reasoning in the comment)

PR #748, opened 2022-09-01
enkit version: Add build time info

minor-fixes:
This change adds a "Built at" line to `enkit version`, which displays
when the binary was built.

Tested:
* `bazel run //enkit -- version` - shows that it was built at the "zero"
  time
* `bazel run --stamp //enkit -- version` - shows that it was built now

### DIFF
--- a/enkit/version/version.go
+++ b/enkit/version/version.go
@@ -36,6 +36,7 @@ func (r *Root) Run(cmd *cobra.Command, args []string) error {
 	}
 	fmt.Printf("Built from branch: %s\n", stamp.GitBranch)
 	fmt.Printf("Builder: %s\n", stamp.BuildUser)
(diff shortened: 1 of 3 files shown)


enkit/version/version.go
```diff
@@ -36,6 +36,7 @@ func (r *Root) Run(cmd *cobra.Command, args []string) error {
 	}
 	fmt.Printf("Built from branch: %s\n", stamp.GitBranch)
 	fmt.Printf("Builder: %s\n", stamp.BuildUser)
+	fmt.Printf("Built at: %s\n", stamp.BuildTimestamp())
```
ccontavalli, line 39: For human consumption? yes. Not for script consumption, right?

ccontavalli (approved):

enkit/version/version.go
minor-fixes: Yes, human consumption

PR #755, opened 2022-09-12
enkit tunnel: Delete orphaned UDS sockets on start

minor-fixes:
Prior to this change, a tunnel could not be opened on a particular UDS path if the socket already existed, with the assumption being that there is another process potentially using that socket. Though the logic would check for this by dialing the socket and looking for failure, it would still error if the process controlling the previous socket instance had died.

With this change, the socket is aggressively removed so that the new instance can listen at that same path. The assumption is that it is rare that the old process still exists if the socket dial fails (it is more common that it was killed not-nicely, in which case removing the socket is the correct fix).

Tested: with a change to using UDS for buildbarn in internal:
* `BAZEL_PROFILE=buildbarn bazel test //guidelines/...` - creates tunnels
* `BAZEL_PROFILE=buildbarn bazel test //guidelines/...` - reuses tunnels
* `ps -aux | grep tunnel` and `kill -9` on all PIDs
* `BAZEL_PROFILE=buildbarn bazel test //guidelines/...` - creates tunnels, no error

Jira: INFRA-1505

### DIFF
--- a/proxy/ptunnel/commands/tunnel.go
+++ b/proxy/ptunnel/commands/tunnel.go
@@ -318,11 +318,18 @@ func (r *Tunnel) OpenAndListen(ctx context.Context, proxy *url.URL, id, host str
 					return err
 				}
 
-				// Otherwise, this must be an orphaned socket file - provide
-				// instructions to manually delete.
-				return fmt.Errorf("socket path %q already exists. Delete this socket file manually if it was left in place by a crashing tunnel.", localAddr)
+				// Otherwise, this must be an orphaned socket file - delete the file and
+				// try again
+				if err := os.Remove(localAddr); err != nil {
+					return fmt.Errorf("failed to clean up orphaned Unix socket %q: %w", localAddr, err)
+				}
+				l, err = net.Listen(network, localAddr)
+				if err != nil {
+					return fmt.Errorf("while re-listening after cleaning up socket %q: %w", localAddr, err)
+				}
+			} else {
+				return fmt.Errorf("failed to listen on %s %s: %w", network, localAddr, err)
 			}
-			return fmt.Errorf("failed to listen on %s %s: %w", network, localAddr, err)
 		}
 
 		var ok bool


proxy/ptunnel/commands/tunnel.go
```diff
@@ -318,11 +318,18 @@ func (r *Tunnel) OpenAndListen(ctx context.Context, proxy *url.URL, id, host str
 					return err
 				}
 
-				// Otherwise, this must be an orphaned socket file - provide
-				// instructions to manually delete.
-				return fmt.Errorf("socket path %q already exists. Delete this socket file manually if it was left in place by a crashing tunnel.", localAddr)
+				// Otherwise, this must be an orphaned socket file - delete the file and
+				// try again
+				if err := os.Remove(localAddr); err != nil {
+					return fmt.Errorf("failed to clean up orphaned Unix socket %q: %w", localAddr, err)
```
ccontavalli, line 324: "Try removing it manually?" nit

ccontavalli (approved):

minor-fixes:
> Did you also try a killall of enkit with SIGSTOP? eg, tunnel frozen? ok to fix it separately.

```
~/dev/internal] (playlib_instr)
scott-> killall -SIGSTOP enkit
[=== 2022-09-12 16:41:30 === SUCCESS ===]

[~/dev/internal] (playlib_instr)
scott-> BAZEL_PROFILE=buildbarn bazel test //guidelines/...
Using profile: buildbarn
Starting remote tunnel unix:////home/scott/.cache/buildbuddy.sock --proxy=https://gw.gcp01.corp.enfabrica.net. buildbuddy-grpc.service.consul. 13001... already running
Starting remote tunnel unix:////home/scott/.cache/buildbarn.sock --proxy=https://gw.gcp01.corp.enfabrica.net. buildbarn-frontend.service.consul. 12011... already running
| Profiles loaded from: /home/scott/dev/internal/bazel/profiles
| Fragments: /home/scott/dev/internal/bazel/profiles/fragments/common /home/scott/dev/internal/bazel/profiles/fragments/buildbarn /home/scott/dev/internal/bazel/profiles/fragments/buildbuddy /home/scott/dev/internal/bazel/profiles/fragments/gcp01
| Variables: BES_CLIENTADDR=unix:////home/scott/.cache/buildbuddy.sock RBE_CLIENTADDR=unix:////home/scott/.cache/buildbarn.sock BES_RESULTS=buddy.gcp01.corp.enfabrica.net
Running: /home/scott/.cache/bazelisk/downloads/bazelbuild/bazel-5.1.1-linux-x86_64/bin/bazel --bazelrc=/tmp/bazelrc.pjojNm test //guidelines/...
```

sigstop causes the tunnels to get reused (so Listen is failing with EADDRINUSE, but Dial on the existing path is succeeding) but then the build hangs

Adding a context to Dial/Listen therefore can't help - we'd have to actually send some data on the socket

ccontavalli:
still approved, merge at will, good improvement.

PR #756, opened 2022-09-13
remote.bzl: Make rules cquery-clean by default

minor-fixes:
`export()` and `remote()` rules are not cquery-clean, as they require targets to be passed by either attributes or custom bazel flags to execute successfully.

This change makes said rules clean by emitting a no-op script that fails on execution if no targets are specified by either mechanism.

Tested:

* `bazel run //:export --override_repository=enkit=<local path>` builds but fails at runtime with a script error
* `bazel cquery 'deps(//...)' --override_repository=enkit=<local path>` doesn't error due to `//:export` or `//:remote` issues


ccontavalli:
What happens after the WARNING is printed if the flag is not specified?

I would have expected some changes in the .sh script to print the errors.

Carlo

minor-fixes:
> What happens after the WARNING is printed if the flag is not specified? I would have expected some changes in the .sh script to print the errors

Yes, you get a usage error and a failure:

```
Use as:

  bazel run  -- [-r rsync options]... [-s ssh options]... [-t target options]... [-o|-h] [DEST]...

Accepted options:

  DEST         List of dests to copy the target to. The target will then be
               run on the first dest specified, unless -o is used.

               The list of dests is MANDATORY, unless it is already specified
               in the rule in the BUILD.bazel file. In that case, if you pass any
               dest on the command line the list will REPLACE the one supplied
               in the BUILD.bazel file.

               DEST can be:
                 1) A string without / or :, assumed to be a machine name.
		    In this case, ssh will be used to run the remote command,
                    and the defined destination directory will be appended.
                 2) A string containing / or :, assumed to be a full path
                    specified by the user. No path will be appended. ssh will
                    be used if there's a ':' in the name.

                 This roughly matches rsync and scp syntax: a string like
                 'machine00.corp' will be turned into 'machine00.corp:dest/path/',
                 and use ssh. A string like 'machine00.corp:whatever' is assumed
                 to already contain a path, and will not be mangled. A string like
                 './tests' will be used as a plain destination directory.

  -o           Don't run the binary, only copy all the files to the remote dests.

  -r [value]   Adds one or more rsync options.

     For example: "-r'-v' -r'--show-progress'" will append the "-v --show-progress"
     options to rsync.

  -t [value]   Adds one or more options to your target.

     For example: "-t'--baremetal'" will run the specified target with the "--baremetal"
     option. Those options are highly dependant on the target.

  -h           Prints this astonishingly helpful message.

****************************************************************************************
IMPORTANT/DANGER/ACHTUNG: rsync will overwrite and delete files in the output directory.
         Make sure to never point this script to your home, or to a populated directory.
****************************************************************************************

ERROR: You must specify one or more DEST to copy the output to
[=== 2022-09-13 15:50:53 === FAIL: 10 ===]
```

ccontavalli (approved):

PR #757, opened 2022-09-14
lib/bazel: add support for serial builds, more flags, more workspace

ccontavalli:
Recommendation: review one commit at a time, they are self contained.

This PR and set of commits allows to aggressively re-use the cache by running commands sequentially, while setting all the right flags. It includes some minor refactors.

- enkit/lib/bazel: move the code to compute affected targets in a function.
- /lib/bazel: add support for serial query or parallel query.
- lib/bazel: allow to override output_base and bazel query from command line.
- lib/bazel: remove unused method.
- lib/bazel: minor workspace.go refactor to lazily load variables.
- lib/bazel: set output_base from current client when doing things serially.
- lib/bazel: add ability to pass extra arbitrary flags to bazel.
- lib/bazel: improve debug output.

### DIFF
--- a/lib/bazel/affected_targets.go
+++ b/lib/bazel/affected_targets.go
@@ -214,6 +214,23 @@ func SerialQuery(opt GetModeOptions, log logger.Logger) (*GetResult, error) {
 }
 
 func ParallelQuery(opt GetModeOptions, log logger.Logger) (*GetResult, error) {
+	// BUG(INFRA-140) - By default, cargo will download packages to a well-known
+	// directory under $HOME; this will mean that parallel bazel invocations could
+	// race on this directory if they both fetch Cargo packages. Cargo respects
+	// the $CARGO_HOME environment variable, so set it to something unique for
+	// this invocation.
+	cargoHomeStart, err := ioutil.TempDir("", "bazel_cargo_home_*")
+	if err != nil {
+		return nil, fmt.Errorf("failed to create tmpdir for $CARGO_HOME: %w", err)
+	}
+	defer os.RemoveAll(cargoHomeStart)
+
+	cargoHomeEnd, err := ioutil.TempDir("", "bazel_cargo_home_*")
+	if err != nil {
+		return nil, fmt.Errorf("failed to create tmpdir for $CARGO_HOME: %w", err)
+	}
+	defer os.RemoveAll(cargoHomeEnd)
+
 	// Joining the new worktree roots to the relative path portion handles the
 	// case where bazel workspaces are not in the top directory of the git
 	// worktree.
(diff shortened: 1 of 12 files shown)


lib/bazel/options.go
```diff
@@ -22,6 +22,8 @@ type subcommand interface {
 type baseOptions struct {
 	// Bazel's cache directory for this workspace.
 	OutputBase string
+	// Additional flags appended to bazel after other common flags.
+	ExtraFlags []string
```
minor-fixes, line 26: nit: I think these are called "Startup" flags (https://bazel.build/reference/command-line-reference#startup-options) - maybe we should name this var as such (`ExtraStartupFlags`), to disambiguate from "Subcommand" flags (or whatever we call flags that come after the particular bazel command)

lib/bazel/commands/commands.go
```diff
@@ -103,6 +105,9 @@ func NewAffectedTargetsList(parent *AffectedTargets) *AffectedTargetsList {
 	command.Flags().BoolVar(&command.Parallel, "parallel", true, "If set, the bazel query is run in parallel")
 	command.Flags().StringVar(&command.AffectedTargetsFile, "affected_targets_file", "", "If set, the list of affected targets will be dumped to this file path")
 	command.Flags().StringVar(&command.AffectedTestsFile, "affected_tests_file", "", "If set, the list of affected tests will be dumped to this file path")
+	command.Flags().StringVar(&command.Start.OutputBase, "start_output_base", "", "If set, the directory to use as start output base")
+	command.Flags().StringVar(&command.End.OutputBase, "end_output_base", "", "If set, the directory to use as end output base")
+	command.Flags().StringVar(&command.Query, "query", "deps(//...)", "The query to use to find the targets")
```
minor-fixes, line 110: I doubt the rest of the logic can tolerate arbitrary queries - maybe a caveat on this flag?

lib/bazel/affected_targets.go
```diff
@@ -171,6 +116,68 @@ skipTarget:
 	return changedRules, changedTests, nil
 }
 
+func ParallelQuery(start, end, startOutputBase, endOutputBase string, log logger.Logger) (*QueryResult, error, *QueryResult, error, error) {
```
minor-fixes: 5 retvals seems excessive - maybe return a struct instance?

Bonus - struct instance can contain single error fetch method, which can encapsulate the startQueryErr/endQueryErr logic

minor-fixes (approved):

lib/bazel/commands/commands.go
ccontavalli: The reason I added it is 1) so I could test with a query that was faster, 2) I think the logic will work ok for /some/ queries. For example, a team could add a faster presubmit for only their subdirectory? or for eg, the website push, I could constrain to rdeps of web targets? And compute the delta there? Adding a comment.

lib/bazel/options.go
ccontavalli: done

lib/bazel/affected_targets.go
ccontavalli: returning a struct instance, not sure about your second suggestion.

ccontavalli:
another look?

PR #760, opened 2022-09-15
bestie: Raise max gRPC message size

minor-fixes:
Bestie is rejecting messages larger than the gRPC server default of 4MB. Since bazel targets message sizes of up to 50MB, bestie should support this as well.

This change adds a server option that sets the maximum allowed message size to a flag value, which has a default of 50MB.

Tested: no

Jira: INFRA-1828

### DIFF
--- a/bestie/server/main.go
+++ b/bestie/server/main.go
@@ -22,8 +22,8 @@ import (
 )
 
 var (
-	fileTooBigErr      = errors.New("File exceeds maximum size allowed")
-	maxFileSize   int  = (5 * 1024 * 1024)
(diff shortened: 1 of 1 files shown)


mhallmark (approved):

gpaussaenf (approved):
Looks good. I think the existing 5 MB output file size limit is sufficient for now.

msgoldflam (approved):
Approving, but when I looked it up I also saw MaxRecvMsgSize and MaxSendMsgSize.  Make sure these don't need to be changed also

ccontavalli (approved):

PR #763, opened 2022-09-16
bazel/utils: hack files.bzl to avoid error in CI pipeline.

ccontavalli:
Background:
files_to_dir uses tar to recursively copy and preserve files
attributes, while adjusting their paths.

Problem:
In our CI pipeline, one of the most frequent causes of error
is files_to_dir failing with an error indicating that the
input file has changed while tar was processing it.

So far, I have not been able to reproduce this locally, and
by hacking the script and rule in different ways I could not
figure this out. Seems to be happening frequenlty from CI,
but even with multiple manual runs on buildbarn I had no luck.
There's a log of things I tried below.

In this PR:
As this is making our tests extremely unreliable, attempt
the copy multiple times with a short delay in case of failure.
This should paper over the issue - until we figure it out.

We could also change command (eg, use cp or rsync) or instruct
tar to ignore the problem. Online, I've seen reports this could
be caused by clock/timing/file system metadata kind of differences.

But I'd rather make sure the files are correctly in place
by retrying than find out that a test is sporadically failing
because a file was not fully copied.

Tested:
- built our most common targets using this rule, both
  locally and remotely, all seem to be behaving correctly.
- artificially introduced a copy error (tar aliased to false),
  to see the repeated attempts with the sleep.

Things I tried:
- modified the script to output the {file, root, dest} that
  the rule computes as a comment in the script.
- processed this list to verify that:
  - none of the input paths (symlinks, directories, etc) end
    up including the output path (most common source of this error)
  - all the paths used to read the file from are the canonical
    paths provided by bazel (eg, not a mangled path that happens
    to work through some other symlink, matches bazel provided
    file.path).
  - number of files in output matches inputs of the rule/macro
    (again, to check the rule was not picking up undesired files by means of
    symlinks or paths)
- checked that all inputs are listed as inputs to the action.
  (so, eg, we are not reading an undeclared input)
- looked to see if we had input dirs overlapping with input
  files (could not find any, but this is tedious manual looking
  at the code, so I may have missed something).

(in shor: looks like our rule and our code is correct, there's some interaction with buildbarn/bazel that's causing the issue)

### DIFF
--- a/bazel/utils/files.bzl
+++ b/bazel/utils/files.bzl
@@ -164,7 +164,7 @@ def files_to_dir(ctx, dirname, paths, post = ""):
 
     pack = []
     for k, v in roots.items():
-        pack.append("tar -C {root} -hc {files} |tar -x -C {dest}".format(
+        pack.append("tarcopy {root} {dest} {files}".format(
             root = shell.quote(k or "."),
             files = escape_and_join(v),
             dest = dest,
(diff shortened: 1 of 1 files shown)


ccontavalli:
some folks upstream (thanks!) pointed out that tar is sensitive to metadata changes (eg, stuff the stat() call returns). Specifically, if the number of hard links to the file changes, it throws that error. Hard links are used a whole lot by build barn, and observed similar failures in the past. This theory seems very believable, I think the workaround in this PR is still appropriate - as the error could be caused by other less harmless issues. But will update the comment in the code.

pete-enf (approved):
The only thing that is unclear is why 5 retries. Does it fail often on the second pass? third? etc. What kind of coverage are we getting with 5? 90%? 99%?

ccontavalli:
> The only thing that is unclear is why 5 retries. Does it fail often on the second pass? third? etc. What kind of coverage are we getting with 5? 90%? 99%?

Part of the problem here is that neither me nor anyone in the team has been able to reproduce this in a controlled environment. Yet, we see it often enough in CI that it's almost crippling, eg, a PR triggering a large number of tests will most likely hit this problem once, and fail the presubmit. Two or three /gcbrun generally move us past that problem.

My overall feeling is that 2-3 retries would be sufficient, 5 sounded like a good number - the plan is to tune, or find a better fix. But if eg, let's say it fails once every 20 tests, 1/20. 5 consecutive failures should happen only once every ~3M tests.

PR #764, opened 2022-09-21
bazel/linux: kernel_module sets up include path properly

george-enf:
This change will allow linking a cc_library to a kernel module without having to specify the include path explicitly in the kernel module.

Jira: https://enfabrica.atlassian.net/browse/SF-164

### DIFF
--- a/bazel/linux/defs.bzl
+++ b/bazel/linux/defs.bzl
@@ -16,6 +16,8 @@ def _kernel_modules(ctx):
     bundled = []
     for arch in ctx.attr.archs:
         inputs = ctx.files.srcs + ctx.files.kernel
+        includes = []
+        quote_includes = []
         extra_symbols = []
 
         kdeps = []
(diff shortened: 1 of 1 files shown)


bazel/linux/defs.bzl
```diff
@@ -109,7 +113,13 @@ def _kernel_modules(ctx):
...
         cflags += " -fdiagnostics-color=always"
 
-        extra.append("EXTRA_CFLAGS='%s'" % cflags)
+        for include in depset(includes).to_list():
+            cflags += " -I $PWD/%s" % include
+
+        for include in depset(quote_includes).to_list():
+            cflags += " -iquote $PWD/%s" % include
+
+        extra.append("EXTRA_CFLAGS=\"%s\"" % cflags)
```
ccontavalli, line 122: minor nit: can you change the `=` to `+=`? I believe it's supported by `make`, I'm worried we'll need to pass `EXTRA_CFLAGS=...` via `extra` attribute manually, and won't notice it does not work because it is overridden here.

ccontavalli (approved):
this is great! Thank you!

bazel/linux/defs.bzl
george-enf: Good point. Fixed.

PR #768, opened 2022-09-24
gee: add timeout for ssh-add

jonathan-enf:
ssh-add has a failure mode such that, if SSH_AUTH_SOCK is a forwarded
credential socket created by an ssh process, and that ssh process goes
away, ssh-add -l will hang indefinitely.

This PR adds a timeout so that ssh-add won't hang indefinitely, and gee can
attempt to recover from the situation by creating a new ssh-agent.

Tested: Proved that the change is harmless, but haven't proved that it
resolves the issue in all circumstances yet.  At least, the situation
is better than it was.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -429,7 +429,7 @@ function _check_ssh_agent() {
       _fatal "Something is wrong with enkit's ssh agent."
     fi
   fi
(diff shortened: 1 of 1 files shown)


jonathan-enf:

george-enf (approved):

PR #770, opened 2022-09-30
bazel/linux/runner: add support for hard coding wrapper flags.

ccontavalli:
Without this PR, it's not possible to supply wrapper flags to
the rule - only through the command line.

The only alternative is to create a sh_binary wrapper to the
wrapper wrapping the emulator wrapping our code invoked by
the bazel wrappers. In short this PR can avoid yet one more
wrapping.

### DIFF
--- a/bazel/linux/runner.bzl
+++ b/bazel/linux/runner.bzl
@@ -25,6 +25,16 @@ If not specified, the current root of the filesystem will be used as rootfs.
             mandatory = True,
             doc = "List of executable targets to run in the emulator.",
         ),
+        "wrapper_flags": attr.string_list(
+            doc = """\
+Flags to append after '--' at the end of the qemu command line.
+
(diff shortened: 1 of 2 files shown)


bazel/linux/runner.bzl
```diff
@@ -181,6 +191,7 @@ def create_runner(ctx, archs, code, runfiles = None, extra = {}):
         "rootfs": rootfs,
         "init": init.short_path,
         "runtime": runtime_root.short_path,
+	"wrapper_flags": shell.array_literal(ctx.attr.wrapper_flags),
```
george-enf, line 194: `wrapper_flags` or `wrapper_opts`?

ccontavalli: good catch fixed!

george-enf (approved):

PR #773, opened 2022-10-04
enkit: improve enkit agent error message

jonathan-enf:
One time while logging in I saw the following message:

```
2022-10-04T16:12:08.202Z        WARN    error saving credentials, err: symlink /tmp/ssh-wKEavpGmnw/agent.41963 /home/jonathan/.config/enkit/agent: file exists
ERROR: symlink /tmp/ssh-wKEavpGmnw/agent.41963 /home/jonathan/.config/enkit/agent: file exists
```

Everything seemed to work, but this message wasn't sufficient for me to debug
the code, especially since we're never performing the operation in question.
Nothing seems broken, but let's improve the error messages anyway so we can
root cause this properly in the future.

Tested: All existing tests still pass.

### DIFF
--- a/lib/kcerts/ssh.go
+++ b/lib/kcerts/ssh.go
@@ -123,12 +123,12 @@ func (a *SSHAgent) GetStandardSocketPath() (string, error) {
 }
 
 func (a *SSHAgent) UseStandardPaths() error {
-	socket, err := a.GetStandardSocketPath()
+	standard_socket_path, err := a.GetStandardSocketPath()
 	if err != nil {
 		return err
 	}
 
-	if a.Socket == socket {
+	if a.Socket == standard_socket_path {
 		// no standardization needed.
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


lib/kcerts/ssh.go
```diff
@@ -141,15 +141,15 @@ func (a *SSHAgent) UseStandardPaths() error {
...
 	defer os.Remove(tempname)
 	if err := os.Symlink(a.Socket, tempname); err != nil {
-		return err
+		return fmt.Errorf("UseStandardPaths symlink failed: %w", err)
 	}
 	// Rename symlink to the standard name
-	os.Remove(socket) // if it exists
-	if err := os.Rename(tempname, socket); err != nil {
-		return err
+	os.Remove(standard_socket_path) // if it exists
```
ccontavalli, line 147: remove the Remove? os.Rename will overwrite/remove the target on its own if it exists.
Even better: it will do so atomically (on most posix filesystems), so there is no window during which the link may not exist.

See: https://pkg.go.dev/os#Rename, and documentation around the rename syscall.

ccontavalli (approved):

PR #778, opened 2022-10-06
gee: improve error reporting

jonathan-enf:
While working on an unrelated problem, I finally figured out why
gee's error reporting (trap ERR) wasn't working correctly.  I learned:

* `set -E` is required to enable "errtrace" mode, allowing the ERR trap
  to be inherited by functions and subprocesses.

* in bash 4.0, the EXIT trap behavior changed so that the LINENO environment
  variable is always reset to 1 inside EXIT handlers.  (This feels like a
  misfeature to me.)

Now that I can get ERR traps to work correctly, I can get gee to correctly
report the line number, command, and exit code of any failing command that
is causing the gee script to fail.  This should be make debugging user issues
a little easier in the future.

Tested: added a hidden "debug_error_reporting" subcommand to gee that
emulates an error condition.  Running this command results in the
expected error report.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -1888,7 +1888,8 @@ function _log_command_failure() {
   local lineno="$1"
   local cmd="$2"
   local rc="$3"
-  local msg="$(printf "gee:%s: command %q failed with rc=%d" "${lineno}" "${cmd}" "${rc}")"
+  local msg
+  msg="$(printf "gee:%s: command %q failed with rc=%d" "${lineno}" "${cmd}" "${rc}")"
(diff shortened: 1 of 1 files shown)


jonathan-enf:

jonathan-enf:
@george-enf can you take a look at this one, too?  Also #782 needs a gander.

scripts/gee
```diff
@@ -4446,6 +4451,11 @@ function gee__hello() {
   fi
 }
 
+function gee__debug_error_reporting() {
+  echo "About the run /bin/false, which should report an error:"
```
george-enf, line 4455: s/the/to/

george-enf (approved):

PR #779, opened 2022-10-06
Support downloading artifacts from astore by tags

ahungrynacho:
It is sometimes useful to specify download artifacts by human readable tags instead of by UID. Supporting `use_default_shell_env` also allows `--action_env` bazel command line arguments to set environment variables.

Testing:
`bazel run //bazel/astore:test_astore_download_file`

JIRA: INFRA-1942

### DIFF
--- a/bazel/astore/defs.bzl
+++ b/bazel/astore/defs.bzl
@@ -131,7 +131,7 @@ def _astore_download(ctx):
     if ctx.attr.uid:
         command += " --force-uid %s" % ctx.attr.uid
     else:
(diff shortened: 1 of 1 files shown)


mhallmark (approved):

jonathan-enf:
I saw this got submitted but wanted to say: I think we should discourage downloading astore artifacts by tags.  Tags can change at any time and aren't strictly version controlled, so if a tag change breaks something, there is no way to use "git bisect" or any other version control command to return to the last-known working state.

Always referencing astore items by UID does not have this problem.

PR #783, opened 2022-10-10
astore_upload: more sophisticated BUILD file updating

jonathan-enf:
This PR replaces the old shell scripts with a python script,
so that the code to update UID/SHA variables in BUILD files
can be a little more sophisticated.  This also adds the
ability to update a MAP variable that allows by-name lookup
of UID and SHA.

Also: upload-by-directory support is added.

This PR also adds a unit test for astore_upload.

Tested: manual_test.sh makes the expected BUILD file changes.

Also tested that fix_format could be deployed from internal:

`bazel run --override_repository=enkit=/home/jonathan/gee/enkit/uid_dict //tools/fix_format:deploy`


jonathan-enf:

* @aaahrens @ahungrynacho @ccontavalli @jonathan-enf @minor-fixes

PR #792, opened 2022-10-31
gee: fix error reporting from gee commands that report failures

jonathan-enf:
A previous PR (#778) introduces trap-based error reporting for all
unhandled error conditions.  In the course of my testing, I realized
that many gee commands report failure via a non-zero return code,
which was being trapped as an error instead of allowing gee to
terminate normally.

This PR captures the return code of all gee commands and returns
them correctly (using `exit`), removing the spurious error messages.

Tested: Manual testing of `gee pr_checks` in clients with and without
PR failures, which results in the gee command either returning success
or failure.


jonathan-enf:

george-enf (approved):

PR #794, opened 2022-11-03
gee: fix an ERR trap

jonathan-enf:
Now that gee has proper ERR trap notification, I'm discovering a few unhandled
ERR traps (ie. places in the code where a command returns a non-zero exit code
that isn't explicitly checked / handled).

This single-line PR fixes one such unhandled return code.

Tested: manual testing confirms that the err trap is squashed.


jonathan-enf:

george-enf (approved):

PR #795, opened 2022-11-03
machinist: Flag off most machine configuration

minor-fixes:
This change adds a new flag that disables much of the static machine configuration done, so that machinist does not fight with current/future puppet configuration on startup.

This flag defaults to true for backwards compatibility, and will have to be explicitly specified in Puppet.

Dynamic steps (writing out keys, certs, etc. received from the auth server at runtime) are not affected by this flag - only static configuration and installation steps are controlled.

Tested: `bazel build //machinist/cmd && bazel-bin/machinist/cmd/cmd_/cmd node enroll --require-root=false --loglevel-console=debug --modify-machine-config=false --auth-server=https://auth.corp.enfabrica.net --sshd-configuration-file=/tmp/machinist/sshd_config --ca-key-file=/tmp/machinist/ca.pub --host-key-file=/tmp/machinist/host_key`:

```
2022/11/03 20:00:01 [info] Skipping machine configuration steps, as ModifyMachineConfig was not set
2022/11/03 20:00:01 [info] Writing SSHD Configuration
2022/11/03 20:00:01 [info] Writing CA Public Key Configuration
2022/11/03 20:00:01 [info] Writing Host Cert
2022/11/03 20:00:01 [info] Writing Host Key
```

Jira: INFRA-2550

### DIFF
--- a/machinist/machine/commands.go
+++ b/machinist/machine/commands.go
@@ -43,6 +43,7 @@ func NewEnrollCommand(conf *config.Node) *cobra.Command {
 	}
 	// General Flags.
 	c.PersistentFlags().BoolVar(&conf.RequireRoot, "require-root", true, "should the enroll command require root for execution")
+	c.PersistentFlags().BoolVar(&conf.ModifyMachineConfig, "modify-machine-config", true, "perform additional setup (install libnss_autouser and config, PAM scripts, etc.)")
 
 	// NSS AutoUser flags.
 	c.PersistentFlags().StringVar(&conf.LibNssConfLocation, "nss-autouser-conf", "/etc/nss-autouser.conf", "the file location of libnss autouser configuration file")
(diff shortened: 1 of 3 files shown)


machinist/machine/commands.go
```diff
@@ -43,6 +43,7 @@ func NewEnrollCommand(conf *config.Node) *cobra.Command {
 	}
 	// General Flags.
 	c.PersistentFlags().BoolVar(&conf.RequireRoot, "require-root", true, "should the enroll command require root for execution")
+	c.PersistentFlags().BoolVar(&conf.ModifyMachineConfig, "modify-machine-config", false, "perform additional setup (install libnss_autouser and config, PAM scripts, etc.)")
```
ccontavalli, line 46: can you default the flag to true? on puppet machines, it's easy to set this flag to a different value. But at least it won't break the config/setup of non-puppet machines?

ccontavalli (approved):

machinist/machine/commands.go
minor-fixes: done

PR #796, opened 2022-11-07
build_nark: Add code from CloudBuildNotifiers fork

minor-fixes:
This change adds the code from matt@'s CloudBuildNotifiers fork directly to enkit, so it can be built from the enkit repo and further extended.

Most of the work is in resolving various Go/Bazel dependency issues, including:
* updating cel-go to bring in a working version of antlr
* overriding dependency maps where necessary for proto-generated vs. pre-generated Go code to be used

Tested: Added test from fork

Jira: INFRA-2258


build_nark/main.go
```diff
@@ -0,0 +1,235 @@
...
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"html/template"
+	"mime/quotedprintable"
+	"os"
+	"strings"
+
+	"github.com/enfabrica/enkit/lib/logger"
```
mhallmark, line 29: Common convention (and the way it is in internal) is

- stdlib
- third party 
- local

Doesn't look like enkit repo really has any consistency though

minor-fixes: Done

PR #814, opened 2023-01-07
proxy/ptunnel: correctly reload credentials after expiry.

ccontavalli:
Background:
Authentication credentials have a time limit. Generally around 1 day.
Once they expire, they can no longer be used.

Before this PR:

if I closed my laptop today and reopened it tomorrow, an existing
tunnel or ssh session would look frozen. Not because the protocol
cannot recover, but because the tunnel/ssh is stuck using the old
credentials - even after an enkit login.

There is code in enkit to re-load the credentials, but due to the
structure of the code this was only invoked once at session establishment
time, rather than at every attempt.

For the end user, this manifested as "I closed my laptop, re-opened
the day after, sessions are dead, no matter what".

With this PR:

Correctly re-evaluate options at each re-connection attempt, including
the ones to reload the authentication cookie.

For the end user, this now works as "I closed my laptop, re-opened
the day after, run enkit login, my sessions are back in business"

Tested: this is tricky to test in an isolated test, manually verified
by forcefully casuing a disconnect (disable wifi, move to hotspot)
and checking the logs for the message indicating that the credentials
are reloaded. No reloading before PR, successful reloading after.
Also, verified that the log in a running enproxy shows the credentials
correctly.

Additionally: some cleanups in lib/retry, as a separate commit.

- lib/retry: improve documentation, break down function
- proxy/ptunnel: bugifx, reload credentials correctly at attempt time.

### DIFF
--- a/lib/retry/README.md
+++ b/lib/retry/README.md
@@ -1,0 +1,79 @@
+[![Go Reference](https://pkg.go.dev/badge/github.com/enfabrica/enkit/lib/retry.svg)](https://pkg.go.dev/github.com/enfabrica/enkit/lib/retry)
+
+# Overview
+
+The `retry` library is a simple golang library to implement retry logic
+in a simple, configurable, and reliable way. 
+
+For example, let's say you have a `Scrape()` function, to scrape content
+from a remote website: `func Scrape() error`. Scraping fails at times, and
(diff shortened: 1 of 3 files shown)


cbrune (approved):
Looks OK to me.  Thanks for fixing this.

PR #817, opened 2023-01-11
gee bazelgc: handle very long lists of files to delete

jonathan-enf:
Unfortunately, in the real world:

```
$ gee bazelgc
No ~/.cache/bazel directories to delete.
About to delete 21053 old entries from ~/.cache/bazel-disk-cache
Proceed? (y/N)  y
/home/jonathan/bin/gee: line 4597: /usr/bin/rm: Argument list too long
Approximately 0MiB freed.
```

Looks like we ran into the ARG_MAX limit of the bash command line.  To work
around this issue, I've modified gee to delete files in batches of 100 files
per rm invocation.

Tested: Installed and re-ran in the same directory:

```
$ gee bazelgc
No ~/.cache/bazel directories to delete.
About to delete 21053 old entries from ~/.cache/bazel-disk-cache
Proceed? (y/N)  y
Approximately 6190MiB freed.
```

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -4585,6 +4585,7 @@ function gee__bazelgc() {
 
   if [[ -d ~/.cache/bazel-disk-cache ]]; then
     local -a OLD=()
+    local i
     mapfile -d '' OLD < \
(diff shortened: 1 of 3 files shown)


george-enf (approved):
There might a simpler way with xargs but that works as well.

PR #820, opened 2023-01-23
bazel: Nuke all web frontend code and rules from orbit

minor-fixes:
This change removes all frontend code, including:

* the `//ui` dir
* associated bazel rules/macros
* loading of npm packages, frontend-related rules from WORKSPACE

The catalyst for this is INFRA-4192 - use of these rules uses a `managed_directory` feature of WORKSPACE that is removed in bazel 6.0 and is thus blocking bazel updates. Fixing the UI code is:

* probably time-consuming, to understand the nuances of JS/TS compilation, by looking at https://docs.aspect.build/guides/rules_js_migration/
* of little value, since none of this code is actually used in anything

The only dependency that actually exists is on some font and CSS assets; these are inlined into the repo directly, with a note on how to update them if needed.

This also has ancillary benefits:

* reduced build times, since node modules don't need to be fetched
* no build-out-of-sync-type errors that necessitated a `bazel clean --expunge && rm -rf ui/node_modules to resolve

Tested: no

* TODO: should test that asset replacement doesn't cause issues

Jira: INFRA-4192


astore/server/assets/BUILD.bazel
```diff
@@ -1,18 +1,27 @@
...
-        "@npm//:node_modules/material-design-lite/material.min.js",
-        "@npm//:node_modules/material-design-lite/material.min.js.map",
+        "MaterialIcons-Regular.ttf",
+        "MaterialIcons-Regular.woff",
+        "MaterialIcons-Regular.woff2",
+        "material.min.css",
+        "material.min.css.map",
+        "material.min.js",
+        "material.min.js.map",
+        "material-icons.css",
```
minor-fixes, line 22: This might cause path changes - I can certainly start spelunking to see if that should cause any issues, but @ccontavalli maybe you know offhand if it definitely will/won't or alternatively the best way to test?

ccontavalli: I believe the path changes are flattened. I just verified on our current auth deployment, and see that the computed path https://auth.corp.enfabrica.net/material.min.css.

ccontavalli (approved):
on the public mailing list, there has been conversations around the fact that runfiles paths have changed, and not all languages picked up (eg, mentions that the go runfiles was broken). We hardcode runfiles paths, so I'm worried about the migration. Will point you to the thread.

PR #830, opened 2023-01-25
bes_publisher: Add new BEP handler server

minor-fixes:
This change adds a BEP handler server that will eventually publish important messages it sees onto PubSub.

In this iteration, the server simply increments metrics for messages seen, and optionally logs if the logging level is set high enough.

Tested:
* Unit tests
* Manually (see steps below)

To run manually:
* `bazel run //:gazelle && bazel run //bes_publisher -- --alsologtostderr --v=2`
* Run a bazel build: `bazel test //bes_publisher/... --bes_backend=grpc://localhost:6433`
* View metrics: `curl localhost:6433/metrics`

### DIFF
--- a/bes_publisher/buildevent/service.go
+++ b/bes_publisher/buildevent/service.go
@@ -1,0 +1,154 @@
+package buildevent
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/golang/protobuf/ptypes"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	bpb "google.golang.org/genproto/googleapis/devtools/build/v1"
+	"google.golang.org/protobuf/encoding/prototext"
+	"google.golang.org/protobuf/types/known/emptypb"
+
+	bes "github.com/enfabrica/enkit/third_party/bazel/buildeventstream" // Allows prototext to automatically decode embedded messages
+)
+
+var (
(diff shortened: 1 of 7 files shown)


bes_publisher/buildevent/service.go
```diff
@@ -0,0 +1,154 @@
...
+func (b *buildStream) handleMessages() error {
+	for {
+		req, err := b.stream.Recv()
+		if errors.Is(err, io.EOF) {
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+		if err := b.handleEvent(req); err != nil {
```
mhallmark, line 123: Dropping the error here. maybe log it or inc a metric?

bes_publisher/buildevent/service.go
```diff
@@ -0,0 +1,154 @@
+package buildevent
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+
+	bes "github.com/enfabrica/enkit/third_party/bazel/buildeventstream" // Allows prototext to automatically decode embedded messages
```
mhallmark, line 10: import order sorting

mhallmark (approved):

bes_publisher/buildevent/service.go
minor-fixes: `handleEvent` itself has metrics, and `bazelEventFrom` inside that also has a metric recording their outcomes

bes_publisher/buildevent/service.go
minor-fixes: Fixed

PR #831, opened 2023-01-26
codegen: add support for regex replacements.

ccontavalli:
This PR allows codegen to use regexes to replace patterns,
both from filters and from functions.

### DIFF
--- a/tools/codegen/README.md
+++ b/tools/codegen/README.md
@@ -196,6 +196,9 @@ The following generic jinja2 extensions are enabled within codegen:
 * [debug](https://jinja.palletsprojects.com/en/3.1.x/extensions/#debug-extension):
   Adds the "debug" statement to dump debugging information from within the
   template.
+* [case conversion](https://pypi.org/project/jinja2-strcase/):
+  Adds a few filters to convert `CamelCase` to `snake_case` and the other
+  way around.
 
 Additionally, codegen adds the following functions:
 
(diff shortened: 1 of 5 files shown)


tools/codegen/codegen.py
```diff
@@ -76,6 +76,10 @@ def re_split_function(regex, text):
     return re.split(regex, text)
 
 
+def re_replace_function(text, match, sub):
```
jonathan-enf, line 79: less cognitive dissonance if we call it `re_sub_function` and `re_sub` is the filter name below.

jonathan-enf (approved):
lgtm.  can you please also update codegen's documentation (the associated README file) with the new functions we've added?

jonathan-enf:
Though, one comment from long experience: if you think you need to add functionality like re_sub to a jinja template, maybe your template is too complicated and what you really need is an elaboration script that pre-processes the data context you hand to jinja?

Like bash, the best use of jinja templates is very limited use.  We've broken that rule a bunch of times in our code base, to our detriment (jinja templates to produce yaml data to feed to jinja templates?).

tools/codegen/codegen.py
ccontavalli: good suggestion, done.

tools/codegen/README.md
```diff
@@ -196,10 +196,15 @@ The following generic jinja2 extensions are enabled within codegen:
...
+  Adds a few filters to convert `CamelCase` to `snake_case` and the other
+  way around.
 
 Additionally, codegen adds the following functions:
 
 * `bitwise_and(x, y)`: performs a bitwise and operation for two integers.
 * `bitwise_or(x, y)`: performs a bitwise or operation for two integers.
 * `bitwise_xor(x, y)`: performs a bitwise xor operation for two integers.
 * `bitwise_not(x)`: performs a bitwise not operation for an integer.
+* `re_split_function(regx, text)`: splits text based on the regex.
```
jonathan-enf, line 209: itym `re_split` not `re_split_function`.

ccontavalli: good catch, autocomplete, fixed.

PR #832, opened 2023-01-27
tools/codegen: add support for 'as_filter'.

ccontavalli:
This commit adds support for a tiny extension, as_filter, that allows
to use macros defined in templates as filters. This is extremely convenient
to apply repetitive set of filters to - for example - mangle names in
systemic ways.

Examples provided in the code.

### DIFF
--- a/tools/codegen/codegen.py
+++ b/tools/codegen/codegen.py
@@ -80,6 +80,22 @@ def re_replace_function(text, match, sub):
     return re.sub(match, sub, text)
 
 
+@jinja2.pass_context
+def as_filter_function(context, text, macro, *args, **kwargs):
+    """Allows to use a macro as a filter.
+
+    For example:
+        {% macro UpperPrefix(text, prefix = "prefix_") -%}
+        {{prefix}}{{text|upper}}
+        {%- endmacro %}
+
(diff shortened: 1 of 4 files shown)


jonathan-enf (approved):
What's the value of

`{{ "test_string"|as_filter("ImAFilter", "foo_")}}`

vs

`{{ ImAFilter("test_string", "foo_" }}`

The only advantage I think this gives you is the ability to do

```
{{ filter as_filter("MyMacro", "foo") }}
blah blah blah
{{ endfilter }}
```

But, if you're convinced you need this, I'll approve.

ccontavalli:
> What's the value of
> 
> `{{ "test_string"|as_filter("ImAFilter", "foo_")}}`
> 
> vs
> 
> `{{ ImAFilter("test_string", "foo_" }}`
> 
> The only advantage I think this gives you is the ability to do
> 
> ```
> {{ filter as_filter("MyMacro", "foo") }}
> blah blah blah
> {{ endfilter }}
> ```
> 
> But, if you're convinced you need this, I'll approve.

It's syntactic sugar, true. The main advantage is the simplified piping with other filters.
Let me try without and see what it looks like. My initial PR added an `add_filter`, to turn a macro into a proper filter, but turns out the template is compiled, and all filters must be known at compile time, before any code is run.

PR #833, opened 2023-01-27
bes_publisher: Publish to pubsub topic

minor-fixes:
This change adds logic to propagate `BuildStarted` events into a pubsub topic, with the attributes specified in
https://docs.google.com/document/d/1kA17-zwUxFDHrcpogQv5fMa9MLJpTkhwEAMAqzYCBr0/edit#heading=h.kbi1exowvrim.

Unit tests are added to ensure:
* the expected messages are published
* the messages have the correct attributes
* PublishResults received from Publish calls are handled properly

Tested: unit tests

### DIFF
--- a/bes_publisher/buildevent/service.go
+++ b/bes_publisher/buildevent/service.go
@@ -5,18 +5,27 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand"
 	"strings"
+	"sync"
+	"time"
 
+	"cloud.google.com/go/pubsub"
 	"github.com/golang/glog"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	bpb "google.golang.org/genproto/googleapis/devtools/build/v1"
+	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/encoding/prototext"
 	"google.golang.org/protobuf/types/known/emptypb"
 
 	bes "github.com/enfabrica/enkit/third_party/bazel/buildeventstream" // Allows prototext to automatically decode embedded messages
 )
+
+func init() {
+	rand.Seed(time.Now().UnixNano())
+}
 
 var (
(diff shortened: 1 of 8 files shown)


bes_publisher/buildevent/service.go
```diff
@@ -146,9 +199,58 @@ func (b *buildStream) handleEvent(req *bpb.PublishBuildToolEventStreamRequest) e
...
+func (b *buildStream) updateAttrs(event *bes.BuildEvent) {
+	switch payload := event.Payload.(type) {
+	case *bes.BuildEvent_Started:
+		b.attrs["inv_id"] = payload.Started.GetUuid()
+	}
+}
+
+// maybePublish publishes the given event if it is one that we care about;
+// otherwise, the event is dropped.
+func (b *buildStream) maybePublish(event *bes.BuildEvent) error {
```
mhallmark, line 221: nit: the `maybe` prefix feels odd. down below you have a func called `recordErrFrom` yet it doesn't always record an error. Maybe a more descriptive name here is `handleBuildEvent` or something more descriptive to the action the method is doing.

bes_publisher/main.go
```diff
@@ -21,15 +25,40 @@ var (
...
 
+func exitIf(err error) {
+	if err != nil {
+		glog.Exit(err)
+	}
+}
+
 func main() {
 	flag.Parse()
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, os.Kill)
```
mhallmark, line 48: This causes the app to hang and not get killed. To handle this, wait on the ctx.Done() channel and kill the app when received.

mhallmark (approved):

bes_publisher/buildevent/service.go
minor-fixes: I figured `recordErrFrom` is named as such because it always records errors (though non-errors are ignored); `maybePublish` may or may not publish an event, even if it is present

bes_publisher/main.go
minor-fixes: Fixed in next PR

PR #836, opened 2023-01-28
 bazel/utils: add a dynamic_wrapper rule for LD_LIBRARY_PATH.

ccontavalli:
Problem:
When using cc_foreign_rules and dynamic binaries, the generated
binary is often not runnable as is: the linker cannot find the library
at run time due to things like rpath, bazel sandboxing, or the
generated directory names.

So far we've worked around this on a need by need basis, by hard
coding paths in shell or python scripts to export the correct
LD_LIBRARY_PATH.

This PR:
defines a cc_dynamic_wrapper rule that generates a shell wrapper
for dynamic binaries that sets the LD_LIBRARY_PATH correctly
before running the binary.

This rule also embeds some convenience in updating the runfiles
definition with all the specified dependencies.

Read the .bzl file for documentation and details on how to use
the rule itself.

### DIFF
--- a/bazel/utils/cc_dynamic_wrapper.bzl
+++ b/bazel/utils/cc_dynamic_wrapper.bzl
@@ -1,0 +1,192 @@
+"""Rules to generate wrappers for dynmaic binaries."""
+
+load("@bazel_skylib//lib:paths.bzl", "paths")
+load("@bazel_skylib//lib:shell.bzl", "shell")
+load("//bazel/utils:messaging.bzl", "fileowner")
+load("//bazel/utils:messaging.bzl", "package")
+
+def _cc_dynamic_wrapper_impl(ctx):
+    template = """#!/bin/bash
+export LD_LIBRARY_PATH={ldpaths}:$LD_LIBRARY_PATH
+DEBUG=${{DEBUG:-{debug}}}
+test -z "$DEBUG" || {{
+  echo "===== dynamic wrapper {target} - files available ======="
+  find -L .
+  echo "===== running {run} $@ ======="
+  echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH"
+}}
+exec {run} "$@"
+"""
+    run = ""
+    targets = ctx.attr.deps
+    if ctx.executable.bin:
+        targets = [ctx.attr.bin] + targets
+        run = shell.quote(ctx.executable.bin.short_path)
+    if ctx.attr.run:
+        run = ctx.expand_location(ctx.attr.run, targets = targets)
+
+    dirs = {}
+    runfiles = ctx.runfiles()
+    for dep in targets:
+        di = dep[DefaultInfo]
+
+        files = dep.files.to_list()
+        runfiles = runfiles.merge(ctx.runfiles(files = files))
+        if di.files_to_run and di.files_to_run.executable:
+            runfiles = runfiles.merge(ctx.runfiles(files = [di.files_to_run.executable]))
+
+        deprf = di.default_runfiles
+        if deprf and deprf.files:
+            runfiles = runfiles.merge(deprf)
+            files.extend(deprf.files.to_list())
+
+        for f in files:
+            # Why not use CcInfo or info provided through providers?
+            #
+            # Unfortunately, libraries and binaries can be brought in via
+            # rules_foreign_cc or custom rules that either don't provide
+            # the provider, or don't fill it enough to be useful.
+            #
+            # Both bazel native C/C++ rules and those rules, however,
+            # typically do bring in all dependencies as runfiles or deps,
+            # as the dependencies are needed at run time by the binary.
+            #
+            # This code exploits this by adding any .so file or any file
+            # containing .so. as part of the name in the LD_LIBRARY_PATH.
+            if not f.path.endswith(".so") and not ".so." in f.path:
+                if ctx.attr.debug:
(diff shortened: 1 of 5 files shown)


bazel/utils/cc_dynamic_wrapper.bzl
```diff
@@ -0,0 +1,172 @@
...
+load("@bazel_skylib//lib:shell.bzl", "shell")
+load("//bazel/utils:messaging.bzl", "fileowner")
+load("//bazel/utils:messaging.bzl", "package")
+
+def _cc_dynamic_wrapper_impl(ctx):
+    template = """#!/bin/bash
+export LD_LIBRARY_PATH={ldpaths}:$LD_LIBRARY_PATH
+DEBUG=${{DEBUG:-{debug}}}
+test -z "$DEBUG" || {{
+  echo "===== dynamic wrapper {target} - files availale ======="
```
george-enf, line 13: nit: available

bazel/utils/testdata/dynamic/hello-dynamic.cc
```diff
@@ -0,0 +1,10 @@
+#include <iostream>
+#include <gtest/gtest.h>
+
+int main(int argc, char** argv) {
+  // Ensure a function from gtest is included.
+  // Using argc so the compiler cannot optimize it out.
+  EXPECT_EQ(1, argc);
+
+  std::cout << "Hello, world!" << std::endl;
+}
```
george-enf, line 10: a `return 0;` here just to be safe?

george-enf (approved):

bazel/utils/cc_dynamic_wrapper.bzl
```diff
@@ -0,0 +1,172 @@
...
+""",
+    implementation = _cc_dynamic_wrapper_impl,
+    executable = True,
+    attrs = {
+        "bin": attr.label(
+            doc = "Label of a binary to run - if neither bin or run are specified, the wrapper will run $1 passed at run time.",
+            executable = True,
+            allow_files = True,
+            cfg = "target",
+            doc = "FIXME",
```
george-enf, line 149: needs to be removed

bazel/utils/cc_dynamic_wrapper.bzl
```diff
@@ -0,0 +1,172 @@
...
+        if di.files_to_run and di.files_to_run.executable:
+            runfiles = runfiles.merge(ctx.runfiles(files = [di.files_to_run.executable]))
+
+        deprf = di.default_runfiles
+        if deprf and deprf.files:
+            runfiles = runfiles.merge(deprf)
+            files.extend(deprf.files.to_list())
+
+        for f in files:
+            if not f.path.endswith(".so") and not ".so." in f.path:
```
rajachan, line 44: Would it not be better to just grab `out_shared_libs`?

rajachan: Also, FWIW, for some of our internal usage, we have targets that export `.so.X` versioned libs too (which is wrong and I never understood why we needed it, but we still have them)

bazel/utils/testdata/dynamic/hello-dynamic.cc
ccontavalli: good catch, done!

bazel/utils/cc_dynamic_wrapper.bzl
ccontavalli: good catch, done!

bazel/utils/cc_dynamic_wrapper.bzl
ccontavalli: good catch, done!

bazel/utils/cc_dynamic_wrapper.bzl
rajachan: I meant to say, can you not grab out_lib_dir and use it to set ldpaths in L#57

ccontavalli: re `out_shared_libs` is only a thing of `rules_foreign_cc`, `configure_make` and friends unfortunately, and the original attributes of a rule definition are not available to downstream rules. For example, `configure_make` would interpret its rule specific attributes and create a bunch of outputs, and annotate them with `providers`. Here, I can access the outputs and providers only - the original attributes are rule private. The proper way would be to look at the `CcInfo` provider, that `annotates all C/C++ compiled things with the linker properties`, but... that provider is only available / fully filled for .c/.c++ built through bazel, not through Make/Makefile/... (or at least, it does not look like rules_foreign_cc provides all that is needed).

In experiments, it seemed like going through the set of files brought in via dependencies or runfiles was the simplest / most reliable.

That said:
1. I've added a comment in the .bzl file explaining this.
2. `.so.X` files are captured correctly (see second half of the if)
3. Will do a couple more experiments before merging, if you have rules in mind I could test this with, that'd be awesome.

rajachan: I see, makes sense!

rajachan (approved):

bazel/utils/testdata/dynamic/hello-dynamic.cc
ccontavalli: well, this has been changed to check for argv and env, so now returns the correct status.

naveenaketi-enf (approved):

PR #837, opened 2023-01-30
bes_publisher: Add TargetComplete messages, more attributes

minor-fixes:
This change adds:
* propagation of `TargetComplete` messages, so that listeners can process the build completion of non-test targets
* addition of `rule_type` attributes to `TargetComplete` and `TestResult` messages, which will allow handlers to filter by rule type.

Tested: Stood up locally with debug handler, observed correct messages and attributes being propagated

### DIFF
--- a/bes_publisher/buildevent/service.go
+++ b/bes_publisher/buildevent/service.go
@@ -20,8 +20,8 @@ import (
 	"google.golang.org/protobuf/encoding/prototext"
 	"google.golang.org/protobuf/types/known/emptypb"
 
-	bes "github.com/enfabrica/enkit/third_party/bazel/buildeventstream" // Allows prototext to automatically decode embedded messages
 	"github.com/enfabrica/enkit/lib/gmap"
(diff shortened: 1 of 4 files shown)


bes_publisher/buildevent/service.go
```diff
@@ -248,18 +263,71 @@ func (b *buildStream) updateAttrs(event *bes.BuildEvent) {
...
+		if ok {
+			extraAttrs["rule_type"] = targetType
+		} else {
+			metricUnknownTargetType.WithLabelValues("unmatched_completed_target").Inc()
+		}
+
+		// Annoyingly, subsequent events can't be identified by label + aspect, so
+		// remember a label + configuration as well to match up target kinds.
+		k = targetKey(eventID.GetLabel(), eventID.GetConfiguration().GetId())
+		b.typeFromLabelAndConfiguration[k] = targetType
```
mhallmark, line 312: Is it intended to put the default value in here if it isnt found in the map on line 302?

mhallmark (approved):

bes_publisher/buildevent/service.go
minor-fixes: No - moved this into the "ok" block

PR #839, opened 2023-02-01
gee: add find and vimdiff commands

jonathan-enf:
This adds two useful convenience commands to gee.

   find: for quickly finding a named file in a branch.
   vimdiff: vimdiff a file against the parent branch version.

Tested: manually ran each command.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -2382,6 +2382,36 @@ function gee__diff() {
 }
 
 ##########################################################################
+# find command
+##########################################################################
+
+_register_help "find" "Finds a file by name in the current branch." <<'EOT'
+Usage: gee find [options] <expression>
+
+Searches the current branch for the specified expression.  Equivalent
+to running:
+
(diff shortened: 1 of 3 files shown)


jonathan-enf:

scripts/gee
```diff
@@ -5108,7 +5197,7 @@ function _gee_completion() {
     2)
       prev="\${COMP_WORDS[COMP_CWORD-1]}"
       case "\${prev}" in
-        diff|unpack|revert)
+        diff|log|unpack|vimdiff|revert)
```
george-enf, line 5200: is `log` here correct?

george-enf (approved):

scripts/gee
jonathan-enf: It should be!  But my testing exposed a bug in my implementation of `gee log` such that it wasn't handling filename arguments correctly.  `gee log` is now fixed.

PR #843, opened 2023-02-04
bazel: Add enkit_credential_helper

minor-fixes:
This change adds a binary that can act as a credential helper to bazel
as specified in
https://github.com/bazelbuild/proposals/blob/main/designs/2022-06-07-bazel-credential-helpers.md.

If successful, this will allow us to drop the tunnel requirement for
buildbarn.

Tested:
* Credential helper can successfully get through enproxy

Not tested (future PR):
* Helpful errors thrown when credentials don't exist

### DIFF
--- a/bazel/enkit_credential_helper/main.go
+++ b/bazel/enkit_credential_helper/main.go
@@ -1,0 +1,51 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/enfabrica/enkit/lib/config/defcon"
+	"github.com/enfabrica/enkit/lib/config/identity"
+	"github.com/enfabrica/enkit/lib/khttp/kcookie"
+)
+
+type Credentials struct {
+	Headers map[string][]string `json:"headers"`
+}
+
+func exitIf(err error) {
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%v\n", err)
+		os.Exit(1)
+	}
+}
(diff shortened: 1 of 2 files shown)


bazel/enkit_credential_helper/main.go
```diff
@@ -0,0 +1,48 @@
...
+
+	store, err := identity.NewStore("enkit", defcon.Open)
+	exitIf(err)
+
+	_, token, err := store.Load("")
+	exitIf(err)
+
+	creds := &Credentials{
+		Headers: map[string][]string{
+			"cookie": []string{fmt.Sprintf("Creds=%s", token)},
```
ccontavalli, line 41: can you use `http.Cookie` `.String()` to generate the header? The RFC mandates which characters are valid in a cookie, and how to escape the ones that are not valid. I don't think we can guarantee that the token - as is - is RFC/HTTP compliant. I'm worried the creds helper will work in most cases, but whenever the cookie contains some invalid character, it won't work as it's not escaped correctly.

ccontavalli (commented):
Looks great, only use cookie library for escaping, happy to approve at that point.

bazel/enkit_credential_helper/main.go
minor-fixes: Done! This actually turns out better because it eliminates the `fmt.Sprintf`

ccontavalli (approved):

PR #852, opened 2023-02-21
gee find: also search symlinks

jonathan-enf:
This means that `gee find` will also search the bazel-bin and similar
directories for generated files.

Tested: manual testing.


jonathan-enf:

george-enf (approved):

PR #859, opened 2023-03-01
qemu-template: Add tmpfs mount for root directory and /var/log

dsahern:
Allow VMs to have root access via ssh by creating a tmpfs based home directory for root (needed because VMs run as non-root and root id is squashed from host perspective) and /var/log (access to btmp file).

Remove the nosuid remount of base path.


dsahern:
@ccontavalli how are reviews assigned with enkit? not seeing a list available

PR #864, opened 2023-03-07
gee: fix gee pr_submit trashing changes bug

jonathan-enf:
This one is a doozy:  This fixes a bug in gee where failure of the simple `gh
pr merge` command would cause an entire PR to be reverted and the PR closed.

Root cause:

gee's pr_submit command contains roughly this sequence of commands:

```
··_gh pr merge \
····--squash \
····--body-file "${BODYFILE}" \
····--repo "${UPSTREAM}/${REPO}" \
····"${FROM_BRANCH}"
...
··_git checkout -B "${CURRENT_BRANCH}" "upstream/${MAIN}"
··_push_to_origin "+${CURRENT_BRANCH}"
```

The last two commands force the current branch to reset --hard to master,
and then force-pushes to update origin, so that the branch can continue
to be used.  This is safe to do, because if `gh pr merge` failed, the `set -e`
context of this bash script should cause the funciton to exit, and the last
two commands won't run.  Right?

Wrong.  Four months ago I submitted PR #792, which attempted to capture the
return code from any gee command to the log file for later analysis.  In
particular, it changed

```
"gee__${cmdname}" "$@"
```

into

```
"gee__${cmdname}" "$@" || RC=$?
```

Inside any given gee command, the `set -e` option is still set, so one would
naively expect any simple command within the gee function to still return-on-failure.
Unfortunately, because the function is now executing in the context of a left-hand-side
operand to a `||` operator, bash actually sets a extra hidden mode bit that prevents
`-e` from working, even though it remains set.  Additionally, attempting to `set -e`
within the function additionally has no effect.

I consider myself a bash expert and yet I keep learning new, painful lessons
about bash. The short-term fix is to never, ever put a bash function on the
left-hand side of a `||` operator.  The long term fix is to never use bash for
anything this complicated ever again, and to rewrite this script in python
ASAP.

Tested: I updated the `gee colortest` and `gee debug_error_reporting` commands to
demonstrate the problem, and then used them to validate that the problem is resolved
and that failing commands within individual functions will indeed cause gee to
immediately terminate.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -1123,6 +1123,9 @@ function _die() {
 
 function _cmd() {
   # Run a command and generate associated log messages.
+  #
+  # The complexity of this function's logging is all the proof anyone
+  # should need that this script should have been written in python.
   local COLS ESCAPED_CMD
   COLS="$(safe_tput cols)"
   ESCAPED_CMD="$( printf " %q" "$@")"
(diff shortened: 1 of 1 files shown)


jonathan-enf:

ccontavalli (approved):

george-enf:
Interesting related read: https://blog.janestreet.com/when-bash-scripts-bite/

PR #866, opened 2023-03-08
gee: make colors configurable

jonathan-enf:
Some users don't like the new colors, so I've decided to make the
colors configurable via environment variables.

I've also updated the secret "gee colortest" command to print a
quick color table for our users, so they can see what they're getting
for their particular terminal's color palette.

Tested: `gee colortest` seems to work right.


jonathan-enf:

george-enf (approved):

PR #867, opened 2023-03-14
hw dev v2: Support using local path to enkit for astore_download

ahungrynacho:
This change adds a new field to the `astore_download()` rule to allow using the locally installed `/opt/enfabrica/bin/enkit astore` binary instead of building astore from source. This will decouple PD and DFT targets in the Sonora container from using gcc-10 to build protobuf dependencies required by `enkit astore`. If `HOME` is not set, then `enkit` will complain in the sandbox that it cannot find it. In the future, PD and DFT targets should either download dependencies via the workspace rule `astore_package()` or via an absolute path under `/edatools` automatically deployed via pubhook.

Tested in the Sonora container:
- `bazel build --action_env=HOME=$HOME --override_repository=enkit=/home/bbhuynh/enkit //hw/projects/mlm/imp/pd/fp_defs:swc_bm_4bank_128b_pd-fp_def` passes when `astore_client = "/opt/enfabrica/bin/enkit"` in the target as expected.
- `bazel build --action_env=HOME=$HOME --override_repository=enkit=/home/bbhuynh/enkit //hw/projects/mlm/imp/pd/fp_defs:swc_bm_4bank_128b_pd-fp_def` fails as expected when the `astore_client` attribute is omitted since bazel attempts to build `enkit astore` from source.

Expected passing case:
```
INFO: From Action hw/projects/mlm/imp/pd/fp_defs/swc_bm_4bank_128b_pd.fp.pg.def.gz:
Directly downloadable
 | Created                 Creator                        Arch           MD5                              UID                              Size    TAGs          
 | 2022-11-07 19:14:25.621 antarsingh@enfabrica.net       all            91d23c3f787250611a54964c4f2fbf83 86ux83xvetzu36rmehikzi33gi8bfio3 22 MB   [latest]
Target //hw/projects/mlm/imp/pd/fp_defs:swc_bm_4bank_128b_pd-fp_def up-to-date:
  bazel-bin/hw/projects/mlm/imp/pd/fp_defs/swc_bm_4bank_128b_pd.fp.pg.def.gz
INFO: Elapsed time: 2.571s, Critical Path: 1.40s
INFO: 2 processes: 1 internal, 1 linux-sandbox.
INFO: Build completed successfully, 2 total actions
```
Expected failing case:
```
Running: /home/bbhuynh/.cache/bazelisk/downloads/bazelbuild/bazel-6.0.0-linux-x86_64/bin/bazel --bazelrc=/tmp/bazelrc.vS8UtA build --action_env=HOME=/home/bbhuynh --override_repository=enkit=/home/bbhuynh/enkit :swc_bm_4bank_128b_pd-fp_def
INFO: Analyzed target //hw/projects/mlm/imp/pd/fp_defs:swc_bm_4bank_128b_pd-fp_def (388 packages loaded, 13020 targets configured).
INFO: Found 1 target...
ERROR: /home/bbhuynh/.cache/bazel/_bazel_bbhuynh/f5f4b4917f8f987ac1333504c41186dc/external/com_google_protobuf/BUILD:155:11: Compiling src/google/protobuf/arenastring.cc [for tool] failed: (Exit 1): gcc-10 failed: error executing command (from target @com_google_protobuf//:protobuf_lite) /usr/bin/gcc-10 -U_FORTIFY_SOURCE -fstack-protector -Wall -Wthread-safety -Wself-assign -Wunused-but-set-parameter -Wno-free-nonheap-object -fcolor-diagnostics -fno-omit-frame-pointer -g0 -O2 ... (remaining 33 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
ERROR: This is a dummy gcc-10 used to bypass bazel toolchain checks
```

JIRA: INFRA-4783

### DIFF
--- a/bazel/astore/defs.bzl
+++ b/bazel/astore/defs.bzl
@@ -113,13 +113,17 @@ TODO(jonathan): add support for the "dir" attribute.
 """,
 )
 
-def _astore_download(ctx):
+def _astore_download_impl(ctx):
     if ctx.attr.output:
       output = ctx.outputs.output
     else:
       output = ctx.actions.declare_file(ctx.attr.download_src.split("/")[-1])
-    command = ("%s download --no-progress --overwrite -o %s" %
-               (ctx.executable._astore_client.path, output.path))
+
+    if ctx.attr.local:
+        command = "/opt/enfabrica/bin/enkit astore"
+    else:
+        command = ctx.executable.astore_client.path
+    command += " download --no-progress --overwrite -o %s" % output.path
     execution_requirements = {
             # We can't run these remotely since remote workers won't have
             # credentials to fetch from astore.
@@ -145,7 +149,7 @@ def _astore_download(ctx):
       command += " && (echo \"%s\" %s | sha256sum --check -)" % (ctx.attr.digest, output.path)
     ctx.actions.run_shell(
         command = command,
-        tools = [ctx.executable._astore_client],
+        tools = [] if ctx.attr.local else [ctx.executable.astore_client],
         outputs = [output],
         execution_requirements = execution_requirements,
         use_default_shell_env = True,
@@ -157,8 +161,9 @@ def _astore_download(ctx):
 
 # TODO: add an optional "uid" attribute to this rule
 # TODO: add an optional "digest" attribute to this rule
-astore_download = rule(
-    implementation = _astore_download,
+# TODO(INFRA-2288): Always download by UID by default
+_astore_download = rule(
+    implementation = _astore_download_impl,
     attrs = {
         "download_src": attr.string(
             doc = "Provided the full path, download a file from astore.",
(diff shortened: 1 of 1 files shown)


bazel/astore/defs.bzl
```diff
@@ -188,19 +193,32 @@ astore_download = rule(
...
 files from an artifact store.""",
 )
 
+def astore_download(local = False, *args, **kwargs):
+    if local:
+        kwargs["local"] = local
+    else:
+        kwargs["astore_client"] = Label("//astore/client:astore")
+    return _astore_download(*args, **kwargs)
+
```
minor-fixes, line 221: Is the macro specifically necessary, if _astore_client is not added to the dependencies? If not, it can be dropped

ahungrynacho: I modified the macro in a subsequent commit, but the macro is still required since the default behavior is to build astore from source which is the primary use-case. However for PD and DFT targets that need to opt-out of the default usage pattern of astore_download, the bazel rule needs to be explicitly told to omit the astore dependency. This is to avoid having to update every instance of astore_download in the repo to explicitly declare `astore_client = Label("//astore/client:astore")`.

bazel/astore/defs.bzl
```diff
@@ -201,6 +208,11 @@ With this rule, you can easily download
 files from an artifact store.""",
 )
 
+def astore_download(*args, **kwargs):
```
ahungrynacho, line 211: Example Usage:

```
    astore_download(
        name = "{}-fp_def".format(block_name),
        download_src = "pd/fp_defs/{}.fp.pg.def.gz".format(block_name),
        astore_client = None, # use /opt/enfabrica/bin/enkit astore instead
        tags = [
            "manual",
            "no-presubmit",
        ],
    )
    for block_name in get_ispatial_blocks(fp_blocks)
```

ahungrynacho:
> How about something like #869 instead?

Yes, setting the default enkit binary to be fetched by a WORKSPACE target would be more reliable and less error prone than providing a user-configurable toggle that switches between a build-from-source version of enkit vs a hard-coded path on the file system. This is less mental overhead.

PR #868, opened 2023-03-16
gee check: wait for checks to start

jonathan-enf:
Previously, we had this bug:

`gee rerun; gee pr_submit --wait`

Would incorrectly submit a PR even though the checks hadn't started yet.  The
issue is that github is slow, and so after requesting presubmits, there is
a period of time where `gh pr check` tells us that there are no presubmits
and returns a "success" return code.

We don't want that -- so now gee will verify that at least one presubmit check
is being reprorted by `gh pr check` before confirming that checks have passed.
There's still a risk of fast presubmit checks completing before slow presubmit
checks have even started, but I don't know quite what the right solution is
(yet), and this fix will correct the problem >99% of the time.

Tested: `gee rerun && gee check --wait`


jonathan-enf:

george-enf (approved):

PR #871, opened 2023-03-20
python: Revamp how requirements are generated

minor-fixes:
This change introduces an "intent" file `requirements.in` which lists the modules we would like to depend on with non-specific version specifications, which gets programmatically turned into `requirements.txt` by enumerating the transitive dependencies of the modules that we want.

A `compile_pip_requirements` rule is added to BUILD.bazel to help in generating `requirements.txt`; now the process for adding Python dependencies is:

* Add a dependency line to `requirements.in`
* Regenerate `requirements.txt` by running `bazel run //:python_deps.update`


minor-fixes:
@dwalton76 Adding you since you're familiar with requirements updates in internal - is something like this worth doing in internal as well?

mhallmark (approved):

PR #872, opened 2023-03-20
pygee: Create location, skeleton for Python port of gee

minor-fixes:
Currently, `gee` is 5700 lines of bash that infra-team generally doesn't
maintain, and currently isn't an org priority. However, there are
challenges approaching, as some teams depend on gee for their daily
workflow, and gee may not run in older environments due its requirement
of a modern version of bash.

infra-team's current position/understanding is that the org is expected
to know at least `git` for day-to-day workflows, and that gee being
usable is not currently a priority. In the event that infra-team's
position is too harsh and gee is deemed to be a critical tool, this PR
paves the path for medium-term support by beginning a port to Python.
Porting to Python (as opposed to Golang, etc.) allows for a happy middle-ground between the
conciseness of scripting and the coherence/readability of not-bash, and
maximizes the number of people in the org that can contribute.

This port is named `pygee` (`py` + `gee`, pronounced "piggy"); using the
`click` argument parsing library will allow for the incremental addition
of subcommands, so that the most important subcommands can be
implemented first. The first subcommand added is `version`, which is
left unimpmlemented as the stamp library needs to be backported from
internal.


jonathan-enf (approved):
I had already made progress on a separate effort but I don't want to slow you down if you want to contribute down this path.

PR #877, opened 2023-04-06
enkit: Add `machine-cert` subcommand

minor-fixes:
This change adds boilerplate for (unimplemented) `machine-cert` subcommands `print` and `install` which show a machine's SSH certificate or generates and installs a new one, respectively.

Both commands are left unimplemented; this PR proposes both the names, commands, and file locations only.

`machine-cert` was chosen rather than `login` (which is usually used for people doing one-time logins) and `machine` (which I believe is implemented via plugin bash script.

Tested: Builds, commands print unimplemented

Jira: INFRA-5652

### DIFF
--- a/enkit/machinecert/commands.go
+++ b/enkit/machinecert/commands.go
@@ -1,0 +1,153 @@
+// Package machinecert provides the machine-cert subcommands to enkit.
+package machinecert
+
+import (
+	"fmt"
+
+	"github.com/enfabrica/enkit/lib/client"
+
+	"github.com/spf13/cobra"
+)
+
+type Root struct {
+	*cobra.Command
+	*client.BaseFlags
+
+	PublicKeyPath string
+	PrivateKeyPath string
+	SignedCertPath string
+	SshdConfigPath string
+}
+
+func New(base *client.BaseFlags) (*Root, error) {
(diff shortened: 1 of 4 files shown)


enkit/machinecert/commands.go
```diff
@@ -0,0 +1,89 @@
...
+
+type Print struct {
+	*cobra.Command
+	root *Root
+}
+
+func NewPrint(root *Root) *Print {
+	command := &Print{
+		Command: &cobra.Command{
+			Use:     "print",
```
ccontavalli, line 52: print ... would... read a new certificate from ...? I was expecting a "write" or "generate" that generates a certificate and writes it into a file. Maybe a sign to take an existing certificate and have it signed by the server. Maybe an install like the one you have to setup ssh (or authorized_keys file) as necessary.

ccontavalli (approved):

enkit/machinecert/commands.go
minor-fixes: As we discussed in chat, in the interest of time - going to move ahead with a single command that does the whole flow, with flags to disable disruptive portions/short-circuit portions, as this is more likely to be useful in the short term.

PR #878, opened 2023-04-07
bazel/linux: various feature additions in preparation for next PR

ccontavalli:
The next PR will add support for always being able to start an sshd
server within the test, and providing a cozier debugging environment.

A pre-requisite is to define the steps/commands necessary to run a
test, versus the steps/commands necessary to setup the environment.

This set of commits expands the capabilities of vm_bundle and friends
to support this.

- bazel/linux: remove nv_.* related rules, that are unused and obsolete.
- bazel/linux: add support for initialization steps in vm setup.
- bazel/linux: expand bundle capabilities to support list of binaries.
- bazel/linux: expand bundles to support recursive expansion.
- bazel/linux: improve qemu_test target to propagate init correctly.

### DIFF
--- a/bazel/linux/bundles.bzl
+++ b/bazel/linux/bundles.bzl
@@ -1,5 +1,6 @@
 load("//bazel/linux:providers.bzl", "KernelBundleInfo", "KernelImageInfo", "RuntimeBundleInfo", "RuntimeInfo")
 load("//bazel/linux:utils.bzl", "expand_deps", "get_compatible")
+load("//bazel/linux:runner.bzl", "expand_targets_and_bundles")
 load("//bazel/utils:messaging.bzl", "location", "package")
 load("//bazel/utils:files.bzl", "files_to_dir")
 load("@bazel_skylib//lib:shell.bzl", "shell")
@@ -19,7 +20,7 @@ def _add_attr_bundle(ctx, bundle, name):
         info["binary"] = di.files_to_run.executable
         info["runfiles"] = di.default_runfiles
 
-    bundle[name] = RuntimeInfo(**info)
+    bundle[name] = [RuntimeInfo(**info)]
 
 def _vm_bundle(ctx):
     bundle = {}
@@ -220,8 +221,8 @@ def _kunit_bundle(ctx):
     return [
         DefaultInfo(files = depset([init, check]), runfiles = inside_runfiles.merge(outside_runfiles)),
         RuntimeBundleInfo(
-            run = RuntimeInfo(binary = init, runfiles = inside_runfiles),
-            check = RuntimeInfo(binary = check, runfiles = outside_runfiles),
+            run = [RuntimeInfo(binary = init, runfiles = inside_runfiles)],
+            check = [RuntimeInfo(binary = check, runfiles = outside_runfiles)],
         ),
     ]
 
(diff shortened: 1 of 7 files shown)


ccontavalli:
This PR is mostly a noop from user perspective, but it's a pretty significant change, and would like to merge it on its own, so I can see the PRESUBMITs not regress.

From a user standpoint, there are some less visible improvements:

1) runfiles dependencies are now more precise - I expect this may speed up the build slightly, but may also expose some latent bugs. "more precise" meaning that before this change, all runfiles for all steps (prepare, setup, ...) were brought in all times, now only the runfiles used.

2) the targets built for the VM were incorrectly marked as `cfg = "exec"` instead of `cfg = "target"`. This likely means that `-c dbg` and similar were not propagated correctly! @rajachan as someone who likely suffered for this. Additionally, this caused more rebuilds.

3) `vm_bundle` rules can now be chained together, so a `vm_bundle` can include other vm_bundles.

bazel/linux/bundles.bzl
```diff
@@ -100,11 +126,22 @@ You can set up the bundle like this:
...
         ),
+        "init_cmds": attr.string_list(
+            doc = "Shell commands to run INSIDE THE VM to prepare the VM (before init_bin - optional)",
+        ),
+        "init_bin": attr.label(
+            doc = "Binary to run INSIDE THE VM to prepare the VM (optional)",
+            executable = True,
+            cfg = "exec",
+        ),
+        "init_args": attr.string(
```
cbrune, line 164: should this be a `string_list` instead of just `string`?

bazel/linux/bundles.bzl
```diff
@@ -38,22 +39,45 @@ This is only needed if you need to provide a specific argv, or to
...
+When in debugging mode, a shell is provided after the "init" step is run,
+but before the "run" step is run.
+The "check" step is only run if the "run" step succeeds.
+
+As an example, let's say you need to create an "internal-test" that:
+  1) runs some setup commands outside the VM ("parepare" step)
+  2) has to do something in the VM to prepare it
+     (for example, tune the kernel - "init" step)
+  3) runs a test binary in the VM ("run" step)
+  4) runs some cleanup commands outside the VM ("cleanup" step)
```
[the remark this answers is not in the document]
ccontavalli: I am planning to add a step to run in the VM at the end of the test/run. I was planning on calling it "collect", though, as cleanup is rarely necessary (as the VM is killed) and the most common use case will be collecting dmesg and simialr.

bazel/linux/bundles.bzl
ccontavalli: I've used string instead of string_list throughout here in the hope of "clarifying" the escaping behavior. Maybe it's my brain, but if I see an argv passed as an array, I assume each element can contain spaces or weird characters, and it will be handled correctly (each element passed as is to the next process). I also assume that things like ~ or environment variables are NOT handled, as the array is passed as is to the next process.

This is NOT the case here: the arguments are passed via shell, if it was an array the element separation in the array would not match what the final command would see, as the shell will expand environment variables, and potentially interpret whitespace.

So I used a string. Somewhat similar to the old native python APIs where if you pass an array to subprocess it is propagated as is for argv, while a string is passed via `sh -c` and exposed to shell expansion.

minor-fixes:
I think this breaks ib tests in internal, by misgenerating one of the runner scripts - see https://github.com/enfabrica/internal/pull/17184#issuecomment-1515417000

PR #890, opened 2023-04-11
enproxy: Define config changes for conditionally adding headers

minor-fixes:
This change defines the new types and additions to the enproxy config struct that enable mapping headers to request based on the user's groups.

The proposal enables a config spelled like:

```yaml
- from:
    host: 'monitoring.corp.enfabrica.net'
  to: http://127.0.0.1:3000/
  transform:
    xwebauth: "set"
    mapRequestHeadersByGroup:
    - header: X-Webauth-Role
      groupMapping:
      - group: foo
        value: Admin
      - group: bar
        value: Editor
      - group: ""
        value: Viewer
```

To translate to: "Add `X-Webauth-Role: Admin` to users in the `foo`
group; add `X-Webauth-Role: Editor` to users in the `bar` group; add
`X-Webauth-Role: Viewer` to everyone else".

Tested: N/A

Jira: INFRA-4919

### DIFF
--- a/proxy/httpp/interface.go
+++ b/proxy/httpp/interface.go
@@ -20,6 +20,23 @@ type Regex struct {
 	match      *regexp.Regexp
 }
 
+type HeaderGroupMapping struct {
+	// Header which should get a value
+	Header string
+	// List of groups that should be tested. The header gets the value of the
+	// first group that matches the current request. If no match occurs, the
+	// request is rejected.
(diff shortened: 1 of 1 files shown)


ccontavalli:
Thumbs up, this looks very reasonable. Grafana also supports x-webauth-groups, btw, where the groups are provided verbatim? Not sure that's better.

ccontavalli (approved):

PR #891, opened 2023-04-12
enproxy: Implement mapRequestHeadersByGroup

minor-fixes:
This change implements the "map request headers by group" functionality, which allows header values to be assigned to a given header conditionally based on group matching. #890 has context on how the feature works.

In order to simplify the logic a bit, a helper library is created to convert slices into maps, reducing the need for nested `for` loops within the feature implementation itself.

Tested: unit tests only

Jira: INFRA-4919

### DIFF
--- a/proxy/httpp/interface.go
+++ b/proxy/httpp/interface.go
@@ -6,9 +6,10 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/enfabrica/enkit/lib/logger"
 	"github.com/enfabrica/enkit/lib/khttp"
+	"github.com/enfabrica/enkit/lib/logger"
 	"github.com/enfabrica/enkit/lib/oauth"
+	"github.com/enfabrica/enkit/lib/slice"
 )
 
 type HostPath struct {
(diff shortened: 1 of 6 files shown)


minor-fixes:
Example of usage in https://github.com/enfabrica/internal/pull/17019/files

proxy/httpp/interface.go
```diff
@@ -182,6 +183,28 @@ func (t *Transform) Apply(req *http.Request) bool {
...
+		userGroups = slice.ToSet(creds.Identity.Groups)
+	}
+
+nextHeader:
+	for _, header := range t.MapRequestHeadersByGroup {
+		for _, groupMap := range header.GroupMapping {
+			// Empty group means this header value should always apply
+			if groupMap.Group == "" {
+				req.Header.Set(header.Header, groupMap.Value)
+				continue nextHeader
```
ccontavalli, line 198: nit: simple `break` here? rather than using a label on the outer loop? Don't feel strongly.

ccontavalli (approved):

proxy/httpp/interface.go
minor-fixes: I think the continue reads a bit more explicitly/naturally, and is more likely to work (or get re-adjusted) if code is copy/pasted

PR #895, opened 2023-04-19
Revert "bazel/linux: various feature additions in preparation for next PR"

minor-fixes:
Reason: IB test scripts in internal mis-generated: https://github.com/enfabrica/internal/pull/17184#issuecomment-1515417000


ccontavalli:
https://github.com/enfabrica/enkit/pull/896 should fix the problem here.

PR #896, opened 2023-04-19
bazel/linux: fix breakages introduced by PR#878, variuos improvements

ccontavalli:
This PR fixes issues introduced in PR#878, and uses a better and simpler
approach to handle expansion of parameters.

More details in the descriptions of individual commits.

Tested: run all tests that failed in PRESUBMIT for:
https://github.com/enfabrica/internal/pull/17184

- bazel/linux: propagate arguments correctly.
- bazel/linux: use a better approach to handle recursion, fix bugs.


cbrune (approved):
Looks OK to me.

PR #897, opened 2023-04-19
bazel/workspace_status.sh: fixed stamping of binaries.

ccontavalli:
Background:
lib/stamp/... seems to use the variable naming convention defined by
the buildbuddy documentation, used in the internal/ repository.

When building from the enkit repository (as is the default now with
the new release process? based on astore? not sure) stamping has
broken variables.

In this PR:
- use the same naming convention as the internal workspace_status.sh command.


ccontavalli:
(sorry, had accidentally branched over another branch rather than master)

minor-fixes (approved):

PR #902, opened 2023-04-26
enkit machine-cert: Implement `install` subcommand

minor-fixes:
This change implements the install command.

Not a ton of thought is put into adding tests; this command is useful in
the short-term to quickly issue some certs for the Cadence chamber
machines so that people can SSH between them cleanly.

Tested:
* Run locally; is able to create a keypair and obtain a signed
  cert

Jira: INFRA-5652

### DIFF
--- a/enkit/machinecert/install.go
+++ b/enkit/machinecert/install.go
@@ -1,0 +1,177 @@
+package machinecert
+
+import (
+	"context"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/enfabrica/enkit/astore/rpc/auth"
+	"github.com/enfabrica/enkit/lib/kcerts"
+	"golang.org/x/crypto/ssh"
+)
+
+func (i *Install) Run(cmd *cobra.Command, args []string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	// Get connection to auth server
+	conn, err := i.root.Connect()
+	if err != nil {
+		return fmt.Errorf("can't connect to auth server: %w", err)
+	}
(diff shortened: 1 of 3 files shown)


enkit/machinecert/install.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		publicKey = pubKey.Marshal()
+		publicKeyAuthLine := ssh.MarshalAuthorizedKey(pubKey)
+
+		// Write private key to disk
+		privateKey, err = privKey.SSHPemEncode()
+		if err != nil {
+			return fmt.Errorf("while encoding private key: %w", err)
+		}
+		if err := ioutil.WriteFile(i.root.PrivateKeyPath, privateKey, 0400); err != nil {
```
minor-fixes, line 67: All perms are `0400` which isn't super convenient; need to double-check what they should be

enkit/machinecert/install.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return fmt.Errorf("can't parse public key content: %w", err)
+	}
+
+	req := &auth.HostCertificateRequest{
+		Hostcert: pem.EncodeToMemory(&pem.Block{
+			Type:  "PUBLIC KEY",
+			Bytes: ssh.MarshalAuthorizedKey(pubKey),
+		}),
+		Hosts: []string{"localhost", hostname},
```
minor-fixes, line 92: @ccontavalli are these the expected principles for the machine?

enkit/machinecert/install.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// Check file overwriting
+	_, overwritten, err := i.getAffectedFiles()
+	if err != nil {
+		return fmt.Errorf("while determining changed files: %w", err)
+	}
+	if len(overwritten) > 0 && !i.Overwrite {
+		return fmt.Errorf("--overwrite not specified, but command would change files: %v", overwritten)
+	}
+
```
minor-fixes, line 40: Need to ensure parent dirs exist, maybe here

mhallmark (approved):

PR #906, opened 2023-05-02
bestie: Improve error reporting around fetch failures

minor-fixes:
404's in bestie logs don't show enough info to be debuggable. This change causes these errors to be logged with the URL that was attempted, in order to debug whether the URL is malformed.

Tested: deployed; saw URLs logged in errors

Jira: INFRA-5854

### DIFF
--- a/bestie/server/test_result.go
+++ b/bestie/server/test_result.go
@@ -119,18 +119,18 @@ func openBytestreamFile(fileName, bytestreamUri string) (io.ReadCloser, error) {
 
 	hash, size, err := kbuildbarn.ParseByteStreamUrl(bytestreamUri)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("parsing bytestream url %q: %w", bytestreamUri, err)
 	}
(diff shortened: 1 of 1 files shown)


bestie/server/test_result.go
```diff
@@ -119,18 +119,18 @@ func openBytestreamFile(fileName, bytestreamUri string) (io.ReadCloser, error) {
...
 	client := http.DefaultClient
 	resp, err := client.Get(fileUrl)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("fetching URL %q: %w", fileUrl, err)
 	}
 	respStatus := resp.StatusCode
 	if respStatus != http.StatusOK {
-		return nil, fmt.Errorf("HTTP error status %d", respStatus)
+		return nil, fmt.Errorf("HTTP error status %d while fetching %q: %w", respStatus, fileUrl, err)
```
gpaussaenf, line 133: Won't `err` always be nil here due to preceding check on line 128?

minor-fixes: Good eye, fixed!

PR #907, opened 2023-05-02
kbuildbarn: Fix buildbarn-browser URL generation

minor-fixes:
buildbarn-browser URL patterns changed with a recent upgrade - they now
include the hash function used to calculate a particular blob.

We currently use SHA256 everywhere and don't plan to change, so this
change modifies the client libraries to accept a hash function name, and
buildbarn/unit test code to hardcode "sha256".

Tested: unit tests only

Jira: INFRA-5855

### DIFF
--- a/bestie/server/test_result.go
+++ b/bestie/server/test_result.go
@@ -121,7 +121,10 @@ func openBytestreamFile(fileName, bytestreamUri string) (io.ReadCloser, error) {
 	if err != nil {
 		return nil, fmt.Errorf("parsing bytestream url %q: %w", bytestreamUri, err)
 	}
-	fileUrl := kbuildbarn.Url(deploymentBaseUrl, hash, size, kbuildbarn.WithFileName(fileName))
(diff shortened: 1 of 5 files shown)


bestie/server/test_result.go
```diff
@@ -119,18 +119,21 @@ func openBytestreamFile(fileName, bytestreamUri string) (io.ReadCloser, error) {
...
 	client := http.DefaultClient
 	resp, err := client.Get(fileUrl)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("fetching URL %q: %w", fileUrl, err)
 	}
 	respStatus := resp.StatusCode
 	if respStatus != http.StatusOK {
-		return nil, fmt.Errorf("HTTP error status %d", respStatus)
+		return nil, fmt.Errorf("HTTP error status %d while fetching %q: %w", respStatus, fileUrl, err)
```
gpaussaenf, line 136: Same comment as other PR: `err` is nil here.

PR #910, opened 2023-05-12
k8s_dummy: Add periodic logging

minor-fixes:
This change adds a log statetment that executes every minute, so that it can be used to verify that logs collection works.

Tested: Ran; saw log heartbeat messages

### DIFF
--- a/infra/k8s_dummy/main.go
+++ b/infra/k8s_dummy/main.go
@@ -2,9 +2,14 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
+	"log"
 	"net/http"
+	"os"
+	"os/signal"
+	"time"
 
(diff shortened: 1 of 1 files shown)


infra/k8s_dummy/main.go
```diff
@@ -30,8 +35,30 @@ func printPath(w http.ResponseWriter, r *http.Request) {
...
 
+func logHeartbeat(ctx context.Context, interval time.Duration) {
+	t := time.NewTicker(interval)
+	defer t.Stop()
+	i := 0
+
+	for {
+		select {
+		case <-t.C:
+			log.Printf("Log heartbeat %d", i)
```
pete-enf, line 46: might want a : in the string for string parsing.

pete-enf (approved):
LGTM

PR #911, opened 2023-05-12
auth server: Ensure that exactly one instance is always running

minor-fixes:
This change modifies the AppEngine manifest to enforce that there is always one instance running of the auth server, which should eliminate the possibility of auth errors due to the request timing out before an instance is available.

Running an instance consistently shouldn't incur too much cost, as it runs in a cheap application class.

Tested: N/A

Jira: INFRA-6047


pete-enf (approved):
Ship It!

PR #921, opened 2023-06-07
bazel/dependencies/abseil: patch abseil to support multiple flag libraries.

ccontavalli:
In the latest abseil release there is no sane way to use multiple flag
libraries: there is no easy way to tell abseil to 1) ignore the flags
it does not know about and 2) LEAVE THEM in the returned argv 3) strip
the flags it does parse and know about. 2 out of 3 are possible using
the internal API, 3 is not.

This patch adds the ability to ignore flags abseil does not know about
AND leave them in the returned argv, which enables us to use multiple
flag libraries for code that is not based on abseil.

abseil in head provides a different function to achieve a similar
result, but it's very clumsy to use as it CHANGES the order of argv
arguments, which affects other flag parsing libraries.

Tested:
See corresponding PR in internal using and testing the feature.

### DIFF
--- a/bazel/dependencies/abseil/0001-absl-flags-parse.cc-provide-a-mechanism-to-let-other.patch
+++ b/bazel/dependencies/abseil/0001-absl-flags-parse.cc-provide-a-mechanism-to-let-other.patch
@@ -1,0 +1,61 @@
+From de04c828de9541c853bf4698b3913fac5f1f90a4 Mon Sep 17 00:00:00 2001
+From: Carlo Contavalli <carlo@enfabrica.net>
+Date: Tue, 6 Jun 2023 23:46:13 +0000
+Subject: [PATCH] absl/flags/parse.cc: provide a mechanism to let other parse
+ unknown flags.
+
+Problem:
+If multiple flag parsing libraries are used (sigh) it's important
+to have a mechanism so that 1) unknown flags are left where they are
+in argv (not moved at the beginning, not stripped, ...), and 2) a
+new argv with all known args stripped is returned.
+
+This commit implements just that.
+---
+ absl/flags/internal/parse.h | 7 +++++--
+ absl/flags/parse.cc         | 8 +++++++-
+ 2 files changed, 12 insertions(+), 3 deletions(-)
+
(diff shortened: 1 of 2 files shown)


bazel/dependencies/abseil/0001-absl-flags-parse.cc-provide-a-mechanism-to-let-other.patch
```diff
@@ -0,0 +1,62 @@
...
+ enum class UsageFlagsAction { kHandleUsage, kIgnoreUsage };
+ enum class OnUndefinedFlag {
+-  kIgnoreUndefined,
++  // Undefined flags are ignored, but considered parsed (for ArgvListAction).
++  kIgnoreUndefined,   
++  // Undefined flags are ignored, and considered not parsed (for ArgvListAction).
++  kKeepUndefined,
+   kReportUndefined,
+-  kAbortIfUndefined
++  kAbortIfUndefined,
```
jonathan-enf, line 34: I can't remember which versions of C support the trailing comma and which don't, but maybe better off without it?

ccontavalli: fixed

bazel/dependencies/abseil/0001-absl-flags-parse.cc-provide-a-mechanism-to-let-other.patch
```diff
@@ -0,0 +1,62 @@
...
+index fa953f55..8dff9eb6 100644
+--- a/absl/flags/parse.cc
++++ b/absl/flags/parse.cc
+@@ -776,10 +776,16 @@ std::vector<char*> ParseCommandLineImpl(int argc, char* argv[],
+         continue;
+       }
+ 
+-      if (on_undef_flag != OnUndefinedFlag::kIgnoreUndefined) {
++      if (on_undef_flag != OnUndefinedFlag::kIgnoreUndefined &&
++	  on_undef_flag != OnUndefinedFlag::kKeepUndefined) {
```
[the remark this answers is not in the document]
ccontavalli: fixed

PR #927, opened 2023-06-08
bestie: Fix error loop; add fetch metrics

minor-fixes:
This change adds a metric that records HTTP success/failure + status code for each bytestream fetch.

While looking at this metric, it became clear that all fetches are returning an HTTP 200 even when the item is not found, and the stream is closing early. This might hypothetically cause any wrapper on the `io.Reader` to error. In the case of zip files, we append this error to a list in a hot loop and retry the next file, which could cause an OoM if this codepath always returns a non-EOF error. To guard against this, the code exits the loop on any zip iteration failure.

Tested: Built, deployed, no OoMs in 8h: 
image

Jira: INFRA-6331

### DIFF
--- a/bestie/server/test_result.go
+++ b/bestie/server/test_result.go
@@ -18,6 +18,8 @@ import (
 	"github.com/enfabrica/enkit/lib/kbuildbarn"
 	"github.com/enfabrica/enkit/lib/multierror"
 	bes "github.com/enfabrica/enkit/third_party/bazel/buildeventstream" // Allows prototext to automatically decode embedded messages
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 
 	"github.com/golang/glog"
 	"github.com/xenking/zipstream"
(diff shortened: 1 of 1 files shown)


gpaussaenf (approved):

gpaussaenf:
@minor-fixes What do you think about reducing the severity level of this message to `fmt.Infof` (or `fmt.Warnf`)?
https://github.com/enfabrica/enkit/blob/master/bestie/server/xml_result.go#L148

Historically, bestie gets lots of these `E0608 12:53:47.541332       1 test_result.go:220] Error extracting XML metrics: Unstructured XML test results not supported (use junitxml)` messages, which are "expected" for run-of-the-mill `bazel test` invocations on buildbarn. Even though bestie won't process the test.xml file, should it stop calling it an error?

If you agree, that change could go into this PR.

minor-fixes:
On second thought, unfortunately it's not a straight log - it's an error that bubbles up two functions (among other possible errors) and then is logged, so removing it is not trivial.

My thought here is to do an "if err continue" instead of "if err return" in this case, which should keep the logic, but slightly more testing would need to be done (ensure that insertion can tolerate a non-nil structure with no actual results, etc.) so this is probably better left for another PR.

PR #928, opened 2023-06-15
bazel/dependencies: pull dropbear into the repository.

ccontavalli:
Once this PR is merged, it'll be possible to depend (or run) dropbear (and dropbearkey) directly with:
- bazel run @dropbear//:dropbear -- --help
- bazel run @dropbear//:dropbearkey -- --help

The PR also introduces two patches in dropbear, meant for use in VMs only open on loopback:
1. Adds the `-Z` option, forcing dropbear to always accept a blank password for login.
2. Adds the `-O /path/to/authorized_keys_file` option, forcing dropbear to always use this authorized_keys file.

Option 2. is what I'm using in script (to come in separate PR) to automatically bring up dropbear for each test instance (basically: if you ask to start the test with sshd, an ephemeral keypair is generated, public key stored in an authorized_keys file).

For testing, I typically use commands like:
```
./dropbear -r /tmp/host_key -EF -p 0:2223
./dropbearkey -t ed25519 -f /tmp/host_key
```


bazel/dependencies/dropbear/0000-allow-blank-password.patch
```diff
@@ -0,0 +1,57 @@
+diff --git runopts.h runopts.h
+index 1675836..a42e99b 100644
+--- runopts.h
++++ runopts.h
+@@ -107,6 +107,7 @@ typedef struct svr_runopts {
+ 	int noauthpass;
+ 	int norootpass;
+ 	int allowblankpass;
++        int forceblankpass;
```
nzmsv, line 9: Tabs vs spaces

bazel/dependencies/dropbear/0001-override-authorized-keys.patch
```diff
@@ -0,0 +1,113 @@
...
+ #endif
++        char* authorized_keys_file = NULL;
+ 
+ 
+ 	/* see printhelp() for options */
+@@ -180,6 +183,7 @@ void svr_getopts(int argc, char ** argv) {
+         svr_opts.pubkey_plugin = NULL;
+         svr_opts.pubkey_plugin_options = NULL;
+ #endif
++        svr_opts.authorized_keys_file = NULL;
```
nzmsv, line 90: More tabs vs spaces

bazel/init/stage_1.bzl
```diff
@@ -276,3 +276,16 @@ def stage_1():
...
+
+    maybe(
+        name = "dropbear",
+        repo_rule = http_archive,
+        build_file = "//bazel/dependencies:dropbear.BUILD.bazel",
+        sha256 = "e02c5c36eb53bfcd3f417c6e40703a50ec790a1a772269ea156a2ccef14998d2",
+        urls = ["https://github.com/mkj/dropbear/archive/refs/tags/DROPBEAR_2022.83.tar.gz"],
+        strip_prefix = "dropbear-DROPBEAR_2022.83",
+        patches = [
+            "//bazel/dependencies/dropbear:0000-allow-blank-password.patch",
```
nzmsv, line 288: Could we set the user password in the test VM to blank instead?  This would let us run unpatched dropbear (and simplify future updates)

ccontavalli: We're using 9p in many cases, we don't have a dedicated image for the test. (which has a lot of benefits).
Also, the option being here does not mean we're using it. Eg, in cases we have a full custom image -> yes, let's setup authentication properly in the image itself.

bazel/init/stage_1.bzl
```diff
@@ -276,3 +276,16 @@ def stage_1():
...
+            "//bazel/dependencies/dropbear:0001-override-authorized-keys.patch",
```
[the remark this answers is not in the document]
ccontavalli: I'm not automating the use of blank passwords - it is too risky with the full filesystem shared with the VM and the VM process running as yourself, pretty much. This option I've added mostly for debugging / testing / quick iteration purposes.

The next commits/PR integrating dropbear in the qemu_test targets... what they do is "if debugging via ssh is enabled, generate an ephemeral key", so even on a shared machine - any VM started via testing is only accessible by you, with no password / nothing else in between.

bazel/dependencies/dropbear/0001-override-authorized-keys.patch
ccontavalli: good catch, tried to fix as much as possible (dropbear itself seems inconsistent in a few places?)

bazel/init/stage_1.bzl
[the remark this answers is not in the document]
nzmsv: Got it.  Though the /etc/passwd we'd be sharing through 9p is the one from the container, no?  I don't think we have a mount that would expose the one from the host.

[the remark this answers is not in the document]
ccontavalli: Yes, it's the one from the container. But we're not configuring it much today, and I still think ensuring the test and ssh plumbing just works regardless of the container setup is convenient (also, not sure I want to configure the default container for passwordless auth).

bazel/dependencies/dropbear/0000-allow-blank-password.patch
```diff
@@ -0,0 +1,57 @@
+diff --git runopts.h runopts.h
+index 1675836..a42e99b 100644
+--- runopts.h
++++ runopts.h
+@@ -107,6 +107,7 @@ typedef struct svr_runopts {
+ 	int noauthpass;
+ 	int norootpass;
+ 	int allowblankpass;
++ int forceblankpass;
```
nzmsv, line 9: Nit: indentation

nzmsv (approved):

bazel/dependencies/dropbear/0000-allow-blank-password.patch
ccontavalli: I changed my .vimrc settings recently, and looks like I broke the tab configuration. Thanks for noticing, fixed.

PR #939, opened 2023-06-25
enkit_credential_helper: Use cobra; add check subcommand

minor-fixes:
This change adds a `check` subcommand that will allow for credentials to be tested. This is necessary because bazel does not propagate a credential helper error in an obvious manner, so we will instead have the wrapper script use the `check` subcommand to test credentials before passing them to bazel.

To avoid checking in org-specific URLs, the URL that `check` uses is taken as a command arg, which is OK since it does not have the same interface constraints as the `get` subcommand.

Code is refactored to use `cobra` with this change, since it introduces a new subcommand.

Tested:
* `enkit_credential_helper get` with creds succeeds, prints creds
* `enkit_credential_helper check` succeeds, in <1s
* `enkit_credential_helper check` without creds fails with a (semi-)helpful error

Jira: INFRA-4386


bazel/enkit_credential_helper/main.go
```diff
@@ -21,31 +36,110 @@ func exitIf(err error) {
...
+	return nil
+}
+
+func checkCommand(cmd *cobra.Command, args []string) error {
+	creds, err := fetchCreds()
+	if err != nil {
+		return fmt.Errorf(friendlyCredsError, err)
+	}
+
+	ctx := context.TODO()
```
mhallmark: nit: Use context from command param

mhallmark (approved):

PR #940, opened 2023-06-26
dropbear: fix a few issues in internal integration

ccontavalli:
- dropbear: use @enkit to refer to the BUILD.bazel files and patches.
- dropbear: make sure it builds correctly in RBE environment.


salil-enf (approved):
lgtm.

PR #942, opened 2023-06-27
diff_test: compare binary files

jonathan-enf:
A diff_test user reported a problem where "update_goldens" was failing with a UnicodeDecoderError.
The problem turned out to be the `diff` tool reporting a binary diff, which contained
an invalid utf-8 sequence, which was raising this exception inside RunCheckOutput.

This PR updated the public version of update_goldens to be identical to our internal
version, enabling us to reproduce the issue.  The issue is then addressed by adding a
"binary" flag to diff_test that causes the "cmp" tool to be used instead of "diff"
for diff'ing binary files.  There were other ways to solve this issue, but this
is broadly the correct way to handle diffing binary files.

Tested: Added a foobar.bin and foobar.bin-diff_test.  Verified that update_goldens
fails with binary = False, but succeeds with binary = True.

### DIFF
--- a/bazel/update_goldens
+++ b/bazel/update_goldens
@@ -1,21 +1,39 @@
 #!/usr/bin/python3
-#
-# Usage: update_goldens <args>
-#
-# Each element of args can either be:
-#    a specific bazel diff_test (:file-diff_test),
-#    a wildcard bazel rule (bar:all or //foo/bar/...),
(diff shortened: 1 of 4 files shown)


jonathan-enf:

ccontavalli (approved):

PR #946, opened 2023-07-05
Remove duplicate kernel modules from dependencies

amichalol:
If a kernel_module rule has multiple dependencies that introduce the same module, we need to filter them out before we try to load them. (Functionality is described in the utils.bzl file but not fully implemented).

Tested:
 kunit_test target that was failing due to `insmod: ERROR: could not insert module driver/core/enf-ubuntu-impish-generic/enf/host/enf.ko: File exists` is now passing

### DIFF
--- a/bazel/linux/utils.bzl
+++ b/bazel/linux/utils.bzl
@@ -117,6 +117,6 @@ The modules still to expand are:
         if key in dups:
             continue
         result.append(mod)
-        dups|={key : ""}
+        dups[key] = True
 
     return result
--- a/bazel/linux/utils.bzl
+++ b/bazel/linux/utils.bzl
@@ -117,5 +117,6 @@ The modules still to expand are:
         if key in dups:
             continue
         result.append(mod)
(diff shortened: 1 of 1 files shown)


bazel/linux/utils.bzl
```diff
@@ -117,5 +117,6 @@ The modules still to expand are:
         if key in dups:
             continue
         result.append(mod)
+        dups|={key : ""}
```
ccontavalli, line 120: good catch! so, starlark is not quite like python (would have used a set() instead),  I've never used this syntax. How about a simple `dups[key] = True` for example?

ccontavalli (approved):

bazel/linux/utils.bzl
amichalol: Changed. 
Picked `dict` mostly because `dups` was introduced as a dict a few lines above (and has an O(1) lookup time, if we ever expect this list to grow long).

PR #951, opened 2023-07-17
enproxy: Don't include port number in hostname matching

minor-fixes:
It would seem that if clients include the port number in requests to enproxy, the port number is including in the host matching logic, which will result in 404s. This is inconvenient, and will result in duplicate mappings/more complex wildcard matching to ensure all the possible spellings of a particular hostname are matched correctly.

This change proposes that the port number be stripped from requests before they are matched. The rationale is:
* enproxy only listens on a small, fixed number of ports (typically 80 and 443); by the time a hostname is considered for matching, the port has already taken effect, by nature of the request being processed in the first place
* Matching with ports would be useful if enproxy had multiple sets of matchers that it used for requests coming in on different ports; this is not a use case worth supporting at this time
* Ports should be stripped out of the host when propagating the request anyway, as downstream reverse proxies may have the same behavior/issue.

This fix comes in the form of a wrapping matcher type that:
* strips the port from the request, if present
* delegates to another matcher

...including metrics so that we can monitor this behavior over time.

Jira: INFRA-6906

### DIFF
--- a/proxy/amux/mux_test.go
+++ b/proxy/amux/mux_test.go
@@ -119,4 +119,9 @@ func TestMuxConformance(t *testing.T) {
 	assert.Equal(t, http.StatusOK, Request(m, "host2.net", "/"))
 	assert.Equal(t, http.StatusOK, Request(m, "host2.net.", "/whatever"))
 	assert.Equal(t, 2, countHost2)
+	
+	// Ports shouldn't affect host mappings
+	assert.Equal(t, http.StatusOK, Request(m, "host2.net:443", "/"))
+	assert.Equal(t, http.StatusOK, Request(m, "host2.net.:443", "/whatever"))
+	assert.Equal(t, 4, countHost2)
 }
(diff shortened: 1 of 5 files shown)


proxy/amux/mux_test.go
```diff
@@ -119,4 +119,9 @@ func TestMuxConformance(t *testing.T) {
 	assert.Equal(t, http.StatusOK, Request(m, "host2.net", "/"))
 	assert.Equal(t, http.StatusOK, Request(m, "host2.net.", "/whatever"))
 	assert.Equal(t, 2, countHost2)
+	
+	// Ports shouldn't affect host mappings
+	assert.Equal(t, http.StatusOK, Request(m, "host2.net:443", "/"))
+	assert.Equal(t, http.StatusOK, Request(m, "host2.net.:443", "/whatever"))
+	assert.Equal(t, 4, countHost2)
```
minor-fixes, line 126: @ccontavalli I'm guessing that these should be aliased to match `host.net` as well?

PR #954, opened 2023-07-28
bazel/linux: Explicitly list all kernel tree outputs

nzmsv:
Listing only directories as outputs of the kernel tree rule causes the "dependency checking of directories is unsound" warning and can fail under some scenarios.  One example is when the tree is used as an input of a genrule.

See https://github.com/bazelbuild/bazel/issues/1415#issuecomment-482490890 as well as https://bazel.build/reference/be/general#filegroup

We can explicitly specify all the files in the kernel tree because it's a repository rule.

For testing, I had a genrule that depended on the kernel tree.  Prior to the change it would receive an empty directory as an input (along with the "unsound" warning).  After the change the genrule could use kernel headers as expected.

### DIFF
--- a/bazel/linux/templates/kernel_tree.BUILD.bzl
+++ b/bazel/linux/templates/kernel_tree.BUILD.bzl
@@ -2,7 +2,7 @@ filegroup(
     name = "{name}-tree",
     # Why allow_empty = True? To support compatibility with different package formats,
     # they may expand the content in different directories, and not use others.
-    srcs = glob(["lib", "usr", "install"], allow_empty = True, exclude_directories = 0),
+    srcs = glob(["lib/**/*", "usr/**/*", "install/**/*"], allow_empty = True),
     visibility = [
         "//visibility:public",
     ],


cbrune:
This generally seems OK to me.

Do the regular enf / enf-rxe kernel modules continue to build OK?

nzmsv:
> This generally seems OK to me.
> 
> Do the regular enf / enf-rxe kernel modules continue to build OK?

I tried building the enf and enf-rxe modules and that worked.  But how do we officially test enkit changes?

ccontavalli:
So... originally I believe the rule looked something like what you have in this PR. I changed it because I **thought** it made the rule faster/more efficient.

Eg, the kernel tree has a bazillion files, by using "exclude_directories = 0" and listing only the top level directories, my theory was that the dependency  graph would only have ~3 nodes instead of like ~80k (much less memory at run time, for the target that has the most files), and I was hoping it'd save the hashing / checking of all files (speed! ... and I think that's why there is the "unsound" warning - when bazel stops at the directory, not much is done/checked for the files within the directory).

That said, this was a good while back (probably bazel 4? probably buildfarm? instead of buildbarn as our clustering code, using the tinder bazel-diff I was testing rather than the internal one?), I had not looked at the code so this was all theory in my head, and I was taking stabs in the dark to speed things up. I'm not actually sure this helped, not sure by how much.

My take would be:
1) Change is probably ok - I'm approving it, happy to merge it.
2) However, I'd love if you could check that building the kernel module targets from internal with `--override_respository=enkit=/path/to/your/modified/enkit` changes the build time of kernel module targets in any sensible way with and without buildbarn before merging.

If the build time is pretty much the same - go ahead and merge. If they build time changes significantly, then let's discuss again.

Even if build time is pretty much the same, there's a change the target differ that computes the tests/targets for presubmit will get slower? Once merged, the presubmit will run, and we can roll back if that's the case - I think the probability is low.

Giving @minor-fixes a heads up anyway as this is likely a change that is riskier than average.

ccontavalli (approved):

PR #955, opened 2023-08-04
rules_foreign_cc: Update to latest master commit

minor-fixes:
rules_foreign_cc was recently updated to the latest release, though this release is now a year old and doesn't contain the newest `meson` rule, which we can use to drop our current implementation. This PR updates `rules_foreign_cc` to the latest master commit.

Tested: `@qemu-uds` targets still build in internal with this change (and tweaks to make `meson()` rules compatible with rules_foreign_cc)

### DIFF
--- a/bazel/init/stage_1.bzl
+++ b/bazel/init/stage_1.bzl
@@ -227,9 +227,9 @@ filegroup(
     maybe(
         name = "rules_foreign_cc",
         repo_rule = http_archive,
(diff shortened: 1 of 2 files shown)


minor-fixes:
Adding @george-enf, who wrote the rules in enkit - not sure of the history here, but it looks like maybe meson support in rules_foreign_cc used to be sparse, but is better now? I think I can replace uses of `meson()` in internal with the upstream rule with a few attribute tweaks; is there a reason I should reconsider/tread more carefully here?

jonathan-enf (approved):

PR #957, opened 2023-08-10
update kunit_parser.py to latest version

george-enf:


bazel/linux/kunit/kunit_parser.py
```diff
@@ -17,7 +17,10 @@
 from enum import Enum, auto
 from typing import Iterable, Iterator, List, Optional, Tuple
 
-from kunit_printer import stdout
+try:
+	from kunit_printer import stdout
+except:
+	from .kunit_printer import stdout
```
ccontavalli, line 23: in past PRs we just changed the `from` to `from bazel.linux.kunit import stdout` without try and except.

george-enf: good point. fixed.

ccontavalli (approved):

PR #964, opened 2023-08-30
storage: Add storage service implementing cas

mhallmark:
Adds an unimplemented storage service that implements the content addressable storage grpc api. This service does nothing except for returning the unimplemented error. The purpose of this is to decide on location of the service, names, etc before adding implementation on top of it all.

Also ran go mod tidy to import the other new dependencies, and ran gazelle_update_repos. go.mod and go.sum changed, but the go_repositories didnt.

Tested:
- `bazel build //...`
- `bazel run storage/server` and used grpc_cli to list the services (verify the server starts up correctly)


minor-fixes (approved):
can this be named something less generic than "storage"?

PR #966, opened 2023-09-07
astore: Split out auth main.go

mhallmark:
Previous PR moved the bulk of the files from astore to auth directory with the intention of splitting out the auth server from the astore binary.

Some files were moved to the auth directory that astore still needed like assets and templates, so those directories were copied back to astore. This makes duplicate auth/server/assets and astore/server/assets directories. Some of these assets may be removed from either directory if they arent needed.

The main part of this PR creates a new main.go under auth/server so auth and astore can be deployed separately.

Testing - neither of these binaries have been fully tested. I was able to get the astore server running locally, but wasn't able to start auth (was missing some flag values).


jungleraptor:
This seems to have gone stale - feel free to reassign if the cobwebs get shaken off

PR #967, opened 2023-09-10
codegen: add a 'deps' attribute to the rule

cbrune:
This patch adds a `deps` label_list attribute to the `codegen()` rule.
The files listed in the `deps` are passed as `inputs` to the
implementation's `ctx.action.run()`.

This allows codegen top-level jinja2 templates to import or include
other templates, which is nice for creating re-usable template macro
libraries.

### DIFF
--- a/tools/codegen/codegen.bzl
+++ b/tools/codegen/codegen.bzl
@@ -16,7 +16,7 @@ def _codegen_impl(ctx):
         args.add("--override", kvpair)
 
     ctx.actions.run(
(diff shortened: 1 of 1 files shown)


cbrune:
This PR has an example of using the `deps` attribute:
https://github.com/enfabrica/internal/pull/24193

silas-enf (approved):
LGTM

jonathan-enf (approved):

PR #968, opened 2023-09-11
go: Convert to `import_alias` target name style 

minor-fixes:
This change migrates all BUILD files from the legacy `go_default_xxx`
naming convention to the newer `import` naming convention, while still
maintaining aliases to the old target types.

This allows `enfabrica/internal` to migrate to the new style of imports
in a non-breaking fashion if needed, and then the aliases can be dropped
by migrating again to the `import` target name style

`auth/server/templates` gets deleted, as it looks like a duplicate of `auth/server/assets/templates` inadvertently created in #961 

Tested: `bazel test //infra/... --override_repository=enkit=<this PR>`
passes in internal


lib/khttp/kclient/BUILD.bazel
```diff
@@ -1,9 +1,15 @@
...
     visibility = ["//visibility:public"],
-    deps = ["//lib/kflags:go_default_library"],
+    deps = ["//lib/kflags"],
+)
+
+alias(
+    name = "go_default_library",
+    actual = ":kclient",
+    visibility = ["//visibility:public"],
 )
```
james-enf, line 15: I assume you're adding these aliases because to not update them would require many more file changes, and this keeps the already-huge PR line diff smaller.

i.e. Are these added only in the places they are needed to pass `bazel build` / `bazel test`?

mhallmark (approved):

lib/khttp/kclient/BUILD.bazel
minor-fixes: Since internal imports this repository, I'm nervous about a change here that breaks internal mostly.

They're currently everywhere - gazelle gets put in a mode where it creates both (which is the simplest to manage).

After we're sure internal can handle it, I can send a similar change that puts gazelle in the new-only mode, deleting all the aliases

bb_reporter/BUILD.bazel
```diff
@@ -3,23 +3,23 @@ load("@io_bazel_rules_docker//container:container.bzl", "container_image", "cont
 load("@rules_pkg//:pkg.bzl", "pkg_tar")
 
 go_library(
-    name = "go_default_library",
+    name = "bb_reporter_lib",
```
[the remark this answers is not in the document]
minor-fixes: I imagine most are updated by gazelle, there are a few that are hand-edited (2 qtpl libraries?). If it's beyond that - no idea; could be gazelle getting confused when generating a lib name that collides with a pre-existing rule; could be gazelle generates `_lib` rules when there's a main in the same BUILD file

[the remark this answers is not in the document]
james-enf: Noted, in summary "They are generated"

lib/khttp/kclient/BUILD.bazel
james-enf: sgtm

james-enf (approved):

PR #969, opened 2023-09-11
go: Update bb-remote-execution

minor-fixes:
This change updates the `bb-remote-execution` dependency to the latest version in preparation for fixing the dependencies in enkit such that libraries from this dependency can build successfully.

Tested: `bazel test //...`:


james-enf (approved):
Approving assuming `bazel test //...` actually lets you run all the tests.
When I ran it I saw a bunch of tests didn't run (maybe because I was on my Chromebook and not a hardware kit).

PR #975, opened 2023-09-20
gee: remove "--insecure-storage" from gh auth command

jonathan-enf:
Even though this is the correct workaround for gh's dbus-launcher bug,
it turns out that the `--insecure-storage` flag is not supported on
every version of gh.  Rather than force every user to upgrade, I'm just
removing the flag.  I haven't seen many reports of the dbug-launcher
issue lately anyway, maybe github got around to fixing it after all.

Tested:


jonathan-enf:

george-enf (approved):

PR #980, opened 2023-10-13
gee: disable window title when not running in screen or tmux

jonathan-enf:
priority: low

I didn't realize that the "set window title" escape sequence was specific to
screen/tmux and isn't supported by, say, xterm, causing users of the default
gee prompt to see some extra garbage on their screens.

Tested:

```
$ echo $TERM
screen-256color
$ eval "$(./gee bash_setup)"
$ gcd master
  ( confirmed that window title changed )
$ gcd gee_prompt
  ( confirmed that window title changed )

$ TERM=xterm
$ gcd master
  ( confirmed that window title did not change )
```

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -5589,10 +5589,12 @@ function __gee_prompt_set_colors() {
 function __gee_prompt_set_window_title() {
   local title="$1"
 
-  local ESC=$'\033'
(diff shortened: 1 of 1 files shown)


jonathan-enf:

george-enf (approved):

PR #985, opened 2023-10-19
gee: add bisect command

jonathan-enf:
gee's bisect command is a wrapper around `git bisect` but automates the
process somewhat:
  1. a temporary branch is created for use during bisection, so the
     user's main branch is never changed.
  2. a heuristic is used to try to find a previous good commit at some
     point in the past.
  3. after finding the suspect commit, the tool cleans up after itself
     and removes the temporary branch.

Tested: Used to find a suspect commit in a branch I'm debugging.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -4922,9 +4922,10 @@ a commit that causes the provided command to transition from a success
 to a failure.  During this process, gee will create a special branch
 named `bisect_<branchname>` to perform the bisect operation in.
 
-gee will scan backwards in time until it finds a commit for which the
-provided command passes, and then will use `git bisect` to attempt to
-determine which commit causes the provided command to start to fail.
+gee will attempt to find a previous last good commit by testing a day, a
+week, a month, 3 months, and finally six months into the past.  Once
+a past good commit is found, the `git bisect` command will be used to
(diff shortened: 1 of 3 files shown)


jonathan-enf:

scripts/gee
```diff
@@ -4910,6 +4910,65 @@ function gee__bazelgc() {
...
+
+function gee__bisect() {
+  local CMD
+  CMD=( "$@" )
+  local CUR_BRANCH BISECT_BRANCH
+  CUR_BRANCH="$(_get_current_branch)"
+  BISECT_BRANCH="BISECT_${CUR_BRANCH}"
+  gee__mkbr "${BISECT_BRANCH}"
+  _checkout_or_die "${BISECT_BRANCH}"
+  _git reset --hard "${CUR_BRANCH}"
```
george-enf, line 4943: This will erase any local changes. Is that OK?

george-enf (approved):

scripts/gee
jonathan-enf: It will erase any local changes in the newly created bisect branch.  I should probably make sure a branch with that name doesn't already exist, but it won't change the user's original branch.

PR #994, opened 2023-11-20
device plugin: Add internal interfaces, some plugin impl

minor-fixes:
This PR implements some of the top-level plugin logic:

* A "main" is added that looks like the main in the Nomad-provided
  skeleton example
* `PluginInfo`, `ConfigSchema`, `SetConfig`, and `Fingerprint` are
  implemented with high-level logic
* Unit tests for `Fingerprint` are added

This PR also defines interfaces for `Reserver` (how the plugin
communicates reservations) and `Notifier` (how the plugin receives
notifications that license state has changed, including from jobs on the
same node as well as jobs from other nodes). An implementation is
stubbed out for a postgres-backed table, and mocks are used for the
plugin unit tests.

Tested: unit tests added

### DIFF
--- a/experimental/nomad_resource_plugin/licensedevice/plugin.go
+++ b/experimental/nomad_resource_plugin/licensedevice/plugin.go
@@ -3,14 +3,29 @@ package licensedevice
 import (
 	"context"
 	"fmt"
+	"sort"
 	"time"
 
+	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/nomad/plugins/base"
 	"github.com/hashicorp/nomad/plugins/device"
 	"github.com/hashicorp/nomad/plugins/shared/hclspec"
+
+	"github.com/enfabrica/enkit/experimental/nomad_resource_plugin/licensedevice/sqldb"
+	"github.com/enfabrica/enkit/experimental/nomad_resource_plugin/licensedevice/types"
+	"github.com/enfabrica/enkit/lib/str"
 )
 
 type Plugin struct {
+	Log hclog.Logger
+
+	reserver types.Reserver
+	notifier types.Notifier
+}
+
+type Config struct {
+	DatabaseConnStr string `codec:"database_connection_string"`
+	TableName       string `codec:"database_table_name"`
 }
 
 func NewPlugin() *Plugin {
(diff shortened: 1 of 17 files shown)


experimental/nomad_resource_plugin/licensedevice/plugin.go
```diff
@@ -40,3 +90,83 @@ func (p *Plugin) Reserve(deviceIDs []string) (*device.ContainerReservation, erro
 func (p *Plugin) Stats(ctx context.Context, interval time.Duration) (<-chan *device.StatsResponse, error) {
 	return nil, fmt.Errorf("Stats not implemented")
 }
+
+func (p *Plugin) fingerprintLoop(ctx context.Context, notifyChan <-chan struct{}, resChan chan<- *device.FingerprintResponse) {
+	for {
```
james-enf, line 95: As far as I can tell, this is constantly polling.
Feels like with any notifier we supply we'll want some purposeful delay between polls as having potentially 10s or 100s of machines continually polling the same notifier could have scale issues.

If you agree with the above, then my assumption would be that we could put some rate-limiting in the notifier code. However we set a 5-second timeout in this loop when we call the notifier. So if the notifier rate-limited, this timeout wouldn't place nicely with it.

All of the above to ask:

What's your plan for rate-limiting and timeout handling?
If you agree with adding it, are you thinking about adding it here or in the notifier?

experimental/nomad_resource_plugin/licensedevice/plugin.go
```diff
@@ -40,3 +90,83 @@ func (p *Plugin) Reserve(deviceIDs []string) (*device.ContainerReservation, erro
...
+
+		resChan <- &device.FingerprintResponse{Devices: groups}
+	}
+}
+
+func deviceGroupsFromLicenses(ls []*types.License) ([]*device.DeviceGroup, error) {
+	deviceGroupMap := map[string]*device.DeviceGroup{}
+
+	for _, l := range ls {
+		groupName := fmt.Sprintf("%s::%s", l.Vendor, l.Feature)
```
james-enf, line 128: `continue` (log to follow) if Vendor or Feature are nil?

james-enf (approved):

experimental/nomad_resource_plugin/licensedevice/plugin.go
minor-fixes: Each nomad client needs to run a copy of this application, so each will have a connection to the DB for listening for updates, and each will be watching its own `/proc/mounts` file. In that way, multiple machines aren't using a single notifier object (because each machine is running its own instance of this plugin).

The notifier interface is supposed to discourage polling; instead, it offers a channel that signals that a change has happened, and the top-level code should go re-read. Under the hood, the notifier impl could be polling, or it could have some push-based mechanism.

For the DB, postgres has `LISTEN`/`NOTIFY` that we can use to implement this without doing e.g. a SELECT loop: https://www.postgresql.org/docs/16/sql-notify.html

For files, specifically `/proc/mounts`, it's less clear what to use, though it sounds like `inotify` doesn't work for proc files; the alternative is a syscall called (confusingly) `poll`: https://linux.die.net/man/2/poll (basically, register to receive events on a set of file descriptors)

---

The interface is currently a function that returns a channel of empty struct (notification only, no data) + a function to "get current" - I'm not sure why it can't be a channel of current data (so each channel receive implies an event), and I might change it to that in a future PR.

experimental/nomad_resource_plugin/licensedevice/plugin.go
minor-fixes: ...they really shouldn't be nil (not null constraints in the DB, etc.) We should ensure that's the case, but I'm not sure this is the best place for such a check

james-enf: True, how do we handle it here? I know Matt adds "required" to the proto files on fields we *need* to exist. But I didn't see that here, nor a check.

experimental/nomad_resource_plugin/licensedevice/plugin.go
james-enf: Okay so in this case, the fingerprint loop is waiting for a notifier to trigger, that there was a change (notifyChan)
Then this code goes to that notifier, gets the current state, does some processing, and signals the current state to the resChan (Nomad)

PR #998, opened 2023-11-27
licensedevice plugin: Add SQL table definition 

minor-fixes:
This change adds SQL to define two tables:
* `license_state` tracks state of each license
* `license_state_log` tracks changes to the `license_state` table, to
  ease debugging

The SQL includes minimal comments for some fields that may not be
self-explanatory.

Tested: TODO

### DIFF
--- a/experimental/nomad_resource_plugin/licensedevice/schema/create_tables.sql
+++ b/experimental/nomad_resource_plugin/licensedevice/schema/create_tables.sql
@@ -1,0 +1,44 @@
+/* Uncomment this line to drop + recreate the table. This will generate new UUIDs
+ * for each license instance, which could be disruptive. */
+-- DROP TABLE IF EXISTS license_state;
+CREATE TABLE IF NOT EXISTS license_state (
+  /* This should be an opaque ID, but is stably generated from:
+   * - the vendor
+   * - the feature
+   * - an "index"/license number, if there are multiple
+   * ...and then hashed into a 64-bit number via fnv-1a and base64-encoded.
+   * Creating such an ID will reduce the need to log tuples of (vendor, feature,
+   * license#) everywhere - we can use a single ID instead. Avoiding purely
+   * random IDs like UUIDs could bring about a situation where all the IDs need
+   * to change when clients still remember old IDs.
+   */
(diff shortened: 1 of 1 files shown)


experimental/nomad_resource_plugin/licensedevice/schema/create_tables.sql
```diff
@@ -0,0 +1,44 @@
+/* Uncomment this line to drop + recreate the table. This will generate new UUIDs
+ * for each license instance, which could be disruptive. */
+-- DROP TABLE IF EXISTS license_state;
+CREATE TABLE IF NOT EXISTS license_state (
+  /* This should be an opaque ID, but is stably generated from:
+   * - the vendor
+   * - the feature
+   * - an "index"/license number, if there are multiple
+   * ...and then hashed into a 64-bit number via fnv-1a and base64-encoded.
```
james-enf, line 9: Did a little research because I thought I knew the answer but couldn't find what I'm looking for.

Taking this string and running fnv-1a and base64.
I remember doing something like this to get an explicitly-sized string to save... but I can't remember which one of these cuts the size down. My two questions are basically:
1. Do you need both?
2. Are you sure a VARCHAR 16 is enough to store the result for *all* values.

Again, IIRC when I did something like this I hashed to get a string of a limited-size, and then I would guess the b64 encoding ensures it sits in the DB well. Just wanted to confirm that with you.

james-enf (approved):

experimental/nomad_resource_plugin/licensedevice/schema/create_tables.sql
minor-fixes: I figured the ID should be a string to be maximally flexible (if ID pivots to another construction medium in the future)

Hash creates a number; I figured that encoding using a large alphabet keeps the representation as short as possible, to make the IDs a bit more human-friendly, rather than e.g. a 16 or 32 char hex value

We can change the type to `TEXT` to be maximally accommodating in the future though

minor-fixes: fixed

PR #1000, opened 2023-11-27
licensedevice plugin: Implement docker-based Notifier 

minor-fixes:
This change ends up being a sort of "rest of the owl" change to produce
something nominally demo-able. It includes:

* SQL scripts that initialize a DB with dummy data
* Docker Notifier implementation that listens for containers to be
  started/killed, and looks at their environment vars to deduce which
  licenses it uses
* Various other bug fixes and improvements that allow the thing to
  actually work, be demoed.

There are still lots of glaring holes to be fixed in future changes
and/or a rewrite to a non-experimental dir.

Tested:
* ran locally 2x with:

  ```
  NOMAD_PLUGIN_MAGIC_COOKIE=e4327c2e01eabfd75a8a67adb114fb34a757d57eee7728d857a8cec6e91a7255 \
  bazel run \
    //experimental/nomad_resource_plugin/licensedevice/cmd/nomad_license_plugin \
    -- \
    'postgresql://<USER>:<URL_ENCODED_PW>@<HOSTNAME>:5432/cj_stage' \
    'license_state' \
    'scott-workstation-2'
   ```

 (the last arg is the local hostname, which is changed between the two
  instances
* on one instance, ran `grpc_cli call unix://<SOCKET_FILE>
  hashicorp.nomad.plugins.device.DevicePlugin.Fingerprint ''` to print
  out streaming Fingerprint responses (plugin prints out the socket file
  to use on start)
* on the other instance, made a `Reserve` call, saw Fingerprint respond
  on the first instance as unhealthy
* ran `docker run -it -e LICENSEPLUGIN_RESERVED_IDS=example_ax_1 ubuntu
  /bin/bash` to simulate a workload starting; saw DB get updated and
  more Fingerprint responses
* exited aforementioned container; saw DB get updated and more
  Fingerprint responses


.vscode/settings.json
```diff
@@ -32,5 +32,17 @@
   },
   "go.useLanguageServer": true,
   "go.lintOnSave": "off",
-  "go.vetOnSave": "off"
+  "go.vetOnSave": "off",
+  "sqltools.connections": [
```
mhallmark, line 36: i dont think these were meant to get in this PR? seems like its not a common connection for all the people in the org

experimental/nomad_resource_plugin/licensedevice/cmd/docker_events/main.go
```diff
@@ -0,0 +1,48 @@
...
+
+	eventsChan := c.Chan(ctx)
+
+	fmt.Println("Listening for events...")
+
+loop:
+	for {
+		select {
+		case <-ctx.Done():
+			break loop
```
mhallmark, line 35: unless im missing something...

the jump back to `loop` will cause this case to get called again no? and at this point it would either cause an infinite loop (assuming the Done() channel is closed), or never reach this case again because the Done() channel is drained.

mhallmark (approved):

PR #1008, opened 2023-12-30
gee bisect: handle absence of commits

jonathan-enf:
I noticed that `gee bisect` fails when searching for commits if any of the `git log --since` operations
returns an empty set (ie. during the holidays).  This PR detects and handles this condition.

Tested: Manualy testing in my local client.


jonathan-enf:

* @amichalol @ccontavalli @jonathan-enf @minor-fixes

ccontavalli (approved):

PR #1010, opened 2024-01-04
tools/codegen - Prevent --output when multiple template files are passed

marcelo-enf:


tools/codegen/codegen.py
```diff
@@ -33,10 +33,14 @@
 flags.DEFINE_multi_string("load", [], "List of data files to load.")
 flags.DEFINE_multi_string("override", [], "List of key=value pair data.")
 flags.DEFINE_string("schema", None, "JSON schema to check against.")
-flags.DEFINE_multi_string("output", None, "Output files to generate.")
```
jonathan-enf, line 36: We need FLAGS.output to be a multi_string for use in multigen mode.

marcelo-enf: Not really, in the original code, when `--multigen_mode` was activated it was only accepted one output file: https://github.com/enfabrica/enkit/blob/master/tools/codegen/codegen.py#L237

marcelo-enf: Thanks for the clarifications on the call. I changed the implementation to only block calls where multiple templates are informed and output is specified.

marcelo-enf:
Put in draft to wait for https://github.com/enfabrica/internal/pull/29845 to be merged

ccontavalli (approved):
this looks reasonable to me, but I'd like @jonathan-enf to comment.

jonathan-enf (approved):

PR #1016, opened 2024-01-22
enkit: Add `bazel invocations list-targets` subcommand

minor-fixes:
This change adds a `bazel invocations` subcommand to interact with bazel invocation data from Buildbuddy/Buildbarn. The first subcommand implemented is `list-targets`, to provide an easy way to list targets that were aborted (which is not immediately visible on the Buildbuddy UI).

This command is spelled:

```
enkit bazel invocations list-targets --invocation-id <id>
```

which will list all targets, grouped by status and then sorted alphabetically.

The command can be shortened:

```
enkit bazel inv lt -i <id>
```

Only tests of a particular status can be shown, with the `--status-filter` flag:

```
enkit bazel inv lt -i <id> --status-filter=incomplete
```

The command prints a pretty list when connected to a TTY; otherwise, it simply lists the targets matching the filter.

---

Note: This command will have inconvenient flags set until we roll out a private config with e.g. the API key pre-populated.

Tested: manually, with various invocation IDs

Jira: INFRA-9369

### DIFF
--- a/lib/bazel/commands/target_list.go
+++ b/lib/bazel/commands/target_list.go
@@ -1,0 +1,215 @@
+package commands
+
+import (
+	"fmt"
+	"io"
+	"regexp"
+	"sort"
+	"strings"
+
+	"golang.org/x/exp/maps"
+
+	bespb "github.com/enfabrica/enkit/third_party/bazel/buildeventstream"
+)
+
+// Assume that tests end in `_test`
+// BUG: DV regression tests also add a number suffix, like `test_12`.
+var labelIsTestRegexp = regexp.MustCompile(`_test(_\d+)?$`)
+
+func validStatuses() []string {
+	valid := maps.Keys(bespb.TestStatus_value)
+	for i := range valid {
+		valid[i] = strings.ToLower(valid[i])
+	}
+	return valid
+}
+
+func parseRuleType(r string) string {
+	return strings.TrimSuffix(r, " rule")
+}
+
+type target struct {
+	name   string
+	status bespb.TestStatus
+	rule   string
+	isTest bool
+}
+
+type invocation struct {
+	targets  map[string]*target
+	finished bool
+}
+
+func emptyInvocationStatus() *invocation {
+	return &invocation{
+		targets:  map[string]*target{},
+		finished: false,
+	}
+}
+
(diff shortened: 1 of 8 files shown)


lib/bazel/commands/target_list.go
```diff
@@ -0,0 +1,191 @@
...
+
+	for _, event := range events {
+		switch id := event.Id.Id.(type) {
+		case *bespb.BuildEventId_TargetConfigured:
+			new.targets[id.TargetConfigured.GetLabel()] = bespb.TestStatus_NO_STATUS
+
+		case *bespb.BuildEventId_TargetCompleted:
+			lbl := id.TargetCompleted.GetLabel()
+			if !labelIsTestRegexp.MatchString(lbl) {
+				new.targets[lbl] = bespb.TestStatus_PASSED
```
mhallmark, line 49: Is this meant to be hard coded to PASSED? Double checking to make sure Completed means passed and not completed/success || completed/failed type of thing

lib/bazel/commands/target_list.go
```diff
@@ -0,0 +1,191 @@
...
+	io.WriteString(w, header+"\n")
+	for _, t := range ts.Sorted() {
+		io.WriteString(w, icon+" "+t+"\n")
+	}
+	io.WriteString(w, "\n")
+}
+
+func (t *targetStatus) Print(w io.Writer) error {
+	for _, name := range t.Sorted() {
+		io.WriteString(w, name+"\n")
```
mhallmark, line 188: nit: `fmt.Fprintln` would eliminate the need to concat the newline

mhallmark (approved):

lib/bazel/commands/target_list.go
minor-fixes: Changed all the `io.WriteString`s to `fmt.Fprintln` or `fmt.Fprintf`

lib/bazel/commands/target_list.go
minor-fixes: Yeah - some targets are build only, while some are tests. Here I'm trying to check if a target is build-only, and if so, then the TargetCompleted message holds the final status - if it's a test, we're waiting for a TestResult message for the final status (it should be considered as "still running")

I'll add a comment here to this effect - also the way tests were detected is error-prone, so I beefed this portion up a bit.

PR #1018, opened 2024-02-01
Remove rules_docker

ahungrynacho:
This change removes `rules_docker` from `enkit` which will not be imported by the internal repo. This requires a new version of enkit to be released and referenced by internal.

Tested:

JIRA: INFRA-9463


minor-fixes:
Before merging, recommend spot-checking with internal by doing something like `bazel query 'deps(//...)' --override_repository=enkit=<path to enkit checkout>` (passing the path to your local enkit root with this change checked out)

This will spot-check for any dependency issues that removing all this might cause. It's not an exhaustive check, but if it uncovers issues it will save you from doing a revert, or trying to figure out how to quickly fix-forward.

minor-fixes (approved):

ahungrynacho:
> Before merging, recommend spot-checking with internal by doing something like `bazel query 'deps(//...)' --override_repository=enkit=<path to enkit checkout>` (passing the path to your local enkit root with this change checked out)
> 
> This will spot-check for any dependency issues that removing all this might cause. It's not an exhaustive check, but if it uncovers issues it will save you from doing a revert, or trying to figure out how to quickly fix-forward.

Good call. I will do that.

PR #1020, opened 2024-02-09
licensedevice plugin: Implement docker-based Notifier

jrp-enf:
(Original changes from Scott M) This change ends up being a sort of "rest of the owl" change to produce something nominally demo-able. It includes:

* SQL scripts that initialize a DB with dummy data
* Docker Notifier implementation that listens for containers to be started/killed, and looks at their environment vars to deduce which licenses it uses
* Various other bug fixes and improvements that allow the thing to actually work, be demoed.
* Additional changes to permit to work fully as a nomad plugin.

Tested:
* ran locally 2x with:

  ``` NOMAD_PLUGIN_MAGIC_COOKIE=e4327c2e01eabfd75a8a67adb114fb34a757d57eee7728d857a8cec6e91a7255 \ bazel run \ //experimental/nomad_resource_plugin/licensedevice/cmd/nomad_license_plugin \ -- \ 'postgresql://<USER>:<URL_ENCODED_PW>@<HOSTNAME>:5432/cj_stage' \ 'license_state' \ 'scott-workstation-2' ```

 (the last arg is the local hostname, which is changed between the two
  instances
* on one instance, ran `grpc_cli call unix://<SOCKET_FILE> hashicorp.nomad.plugins.device.DevicePlugin.Fingerprint ''` to print out streaming Fingerprint responses (plugin prints out the socket file to use on start)
* on the other instance, made a `Reserve` call, saw Fingerprint respond on the first instance as unhealthy
* ran `docker run -it -e LICENSEPLUGIN_RESERVED_IDS=example_ax_1 ubuntu /bin/bash` to simulate a workload starting; saw DB get updated and more Fingerprint responses
* exited aforementioned container; saw DB get updated and more Fingerprint responses
* Tested on cj-stage
* Tested multiple license requests
* Tested competing requests (there is a up to a 15 minute delay for the second request to get the resources it needs). 
* Verified nomad_license_manage functionality, tested dynamic addition of a license.

Jira: INFRA-9623


.vscode/settings.json
```diff
@@ -32,5 +32,17 @@
...
+  "sqltools.connections": [
+    {
+      "previewLimit": 50,
+      "server": "localhost",
+      "port": 5432,
+      "driver": "PostgreSQL",
+      "name": "cj_stage",
+      "database": "cj_stage",
+      "username": "postgres",
+      "password": ",FV-X{C<hs$T]v3S"
```
minor-fixes, line 45: Definitely don't check this secret in!

experimental/nomad_resource_plugin/licensedevice/sqldb/sqldb.go
```diff
@@ -24,66 +27,128 @@ const (
...
+		Namespace: "licensedevice",
+		Subsystem: "sqldb",
+		Name:      "get_current_duration_seconds",
+		Help:      "GetCurrent execution time",
+	})
+	metricMyLicenses = promauto.NewGauge(prometheus.GaugeOpts{
+		Namespace: "licensedevice",
+		Subsystem: "sqldb",
+		Name:      "my_licenses",
+		Help:      "How many licenses do I currently have",
```
minor-fixes, line 41: Any chance you'll want to break this down by license type?

experimental/nomad_resource_plugin/licensedevice/sqldb/sqldb.go
```diff
@@ -24,66 +27,128 @@ const (
...
+	})
+	metricSqlFailCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "licensedevice",
+		Subsystem: "sqldb",
+		Name:      "sql_failiures",
+		Help:      "The number of times sql has errored in various sections of the code",
```
minor-fixes, line 47: Typically I'll put a counter at a specific point in the code (e.g, counting `Foo()` function executions, by incrementing a counter at `Foo`'s exit) and then count successful executions into an `outcome=ok` label, and error executions into an `outcome=some_err_desc` label.

This allows one to count:
* successful result rate
* error rate (broken down by error)

...and this can be rolled into a histogram as well, if execution time of `Foo` is important.

---

This has the advantage of one counter for all errors, but it doesn't convey what the non-error rate is; instrumenting at a particular place in the code ("all exit points of Foo on a single counter, with different exit points having different label values") gives a more complete picture of whether the code is behaving as expected (allows us to see issues like repeated successful spamming on the DB, for instance)

"various sections of the code" could reduce the power of this counter as well - if events are received, is it clear where in the code the error is coming from? (by inspection of the labels, this appears to be the case - but if a label is reused, it may get confusing)

experimental/nomad_resource_plugin/licensedevice/sqldb/sqldb.go
```diff
@@ -24,66 +27,128 @@ const (
...
+		nodeID:    nodeID,
 	}, nil
 }
 
 func (t *Table) GetCurrent(ctx context.Context) ([]*types.License, error) {
+	startTime := time.Now()
+	defer metricGetCurrentDuration.Observe(float64(time.Now().Sub(startTime).Seconds()))
 	rows, err := t.db.Query(ctx, queryAllLicenses)
 	if err != nil {
+		metricSqlFailCounter.WithLabelValues("GetCurrentQueryAllLicenses").Inc()
```
minor-fixes, line 78: I've been trying to do `snake_case` for label names and values - `UpperCamelCase` is fine, but we'll want to standardize on something everywhere

jrp-enf: We discussed and agreed on new labelvalues.

experimental/nomad_resource_plugin/licensedevice/plugin.go
```diff
@@ -50,84 +72,83 @@ func (p *Plugin) PluginInfo() (*base.PluginInfoResponse, error) {
...
 	licenses, err := p.reserver.Reserve(ctx, deviceIDs, p.nodeID)
 	if err != nil {
+		metricPluginCounter.WithLabelValues("Reserve", "error_reserve").Inc()
 		return nil, fmt.Errorf("failed to reserve %v: %w", deviceIDs, err)
 	}
 
 	cr := &device.ContainerReservation{}
+	//	for _, l := range licenses {
+	//		cr.Mounts = append(cr.Mounts, l.MountInfo(p.licenseHandleRoot))
+	//	}
```
minor-fixes, line 126: dead code can be removed?

experimental/nomad_resource_plugin/licensedevice/plugin_test.go
```diff
@@ -174,17 +181,20 @@ func TestReserve(t *testing.T) {
...
+					HostPath: filepath.Join(bazel.TestTmpDir(), "vendor_a/feature_1/aaaa"),
+					TaskPath: "/tmp/license_handles/vendor_a/feature_1/aaaa",
+					ReadOnly: true,
+				},
+				{
+					HostPath: filepath.Join(bazel.TestTmpDir(), "vendor_b/feature_2/bbbb"),
+					TaskPath: "/tmp/license_handles/vendor_b/feature_2/bbbb",
+					ReadOnly: true,
+				},
+			}, */
```
minor-fixes, line 195: Dead code

minor-fixes (approved):

PR #1027, opened 2024-02-22
licensedevice: Use hcl logger instead of slog

jrp-enf:
To avoid any conflicts over the stdout interface which the plugin seems to utilize, change all logging to go to the passed-in hcl logger.

Tested: locally

Jira: INFRA-9678

### DIFF
--- a/experimental/nomad_resource_plugin/licensedevice/sqldb/sqldb.go
+++ b/experimental/nomad_resource_plugin/licensedevice/sqldb/sqldb.go
@@ -3,9 +3,9 @@ package sqldb
 import (
 	"context"
 	"fmt"
-	"log/slog"
 	"time"
 
+	"github.com/hashicorp/go-hclog"
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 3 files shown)


experimental/nomad_resource_plugin/licensedevice/sqldb/sqldb.go
```diff
@@ -57,9 +57,10 @@ type Table struct {
 	db        *pgxpool.Pool
 	tableName string
 	nodeID    string
+	Log       hclog.Logger
 }
 
-func OpenTable(ctx context.Context, connStr string, table string, nodeID string) (*Table, error) {
+func OpenTable(ctx context.Context, connStr string, table string, nodeID string, log hclog.Logger) (*Table, error) {
```
minor-fixes, line 63: How about instead:
* importing https://github.com/evanphx/go-hclog-slog at the toplevel
* in `NewPlugin` - calling `slog.SetDefault`, passing a handler that adapts the incoming logger

Then you should be able to log using slog-package-level functions still, and have the messages go to the correct sink. This has the benefit of being less invasive - the logger doesn't need to get passed to every function, which is verbose and makes testing more difficult later.

minor-fixes: ...you're going to have to add more rationale than that - the alternative suggested has articulated benefits

jrp-enf: I am not comfortable with using someone's arbitrary package when we have an existing, simple solution. Your articulated objection of whether it will clobber the default logger is also a concern. This hashicorp plugin stuff looks a little twitchy and the risk of something getting out through that interface seems not worth the benefit.

jrp-enf:
Different PR accepted. Closing this.

PR #1031, opened 2024-02-26
Add deferred close to attempt to clean up connections

jrp-enf:
with the addition of defer db.Reset(), it doesn't seem to leave connections open.

Jira: INFRA-9746

Tested: tried various combinations of success and failure, doesn't leave connections open any more.

### DIFF
--- a/experimental/nomad_resource_plugin/licensedevice/cmd/nomad_license_manage/main.go
+++ b/experimental/nomad_resource_plugin/licensedevice/cmd/nomad_license_manage/main.go
@@ -166,7 +166,7 @@ func main() {
 			}
 		}
 		if !bFoundId {
-			slog.Error("Can not find a license to remove named:", "command", *runFree)
+			slog.Error("Can not find a license to remove named:", "license", *runFree)
 			return
 		}
 		_, err = db.Exec(ctx, "update license_state set usage_state = $1, last_state_change = CURRENT_TIMESTAMP where id = $2",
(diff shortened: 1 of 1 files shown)


experimental/nomad_resource_plugin/licensedevice/cmd/nomad_license_manage/main.go
```diff
@@ -30,6 +30,9 @@ func main() {
 	if err != nil {
 		log.Fatal("Error opening database: ", err)
 	}
+	defer db.Close()
+	defer db.Reset() // If we don't do this, we seem to leak active connections.
```
minor-fixes, line 34: `defer Close()` makes sense; `defer Reset()` makes less sense - maybe some other tactics will help. You can try:

- moving the bulk of `main()` to its own function that takes no args (or maybe better - a single context arg, and main passes it `context.Background()`, or some other context - see next bullet), returns `error`; all the `log.Fatal` calls turn into `return fmt.Errorf(...)` calls. Then `main` can log any error once, at the top level. This will allow all the defers (happening in the new function) to happen before any calls to `log.Fatal`
- instead of using context.Background, use something like https://github.com/enfabrica/enkit/blob/40d4c60844c2ba4f7a6c81d761d18f639041c64d/proxy/ptunnel/commands/tunnel.go#L120 - will allow for Ctrl-C to cancel (relatively) gracefully, and also has a cancel func that will cancel the context, which I believe pgxpool also will close connections on

minor-fixes: `db.Reset()` is still present - is `db.Close()` still insufficient?

mhallmark (approved):

PR #1055, opened 2024-04-23
bazel/utils/merge_kwargs.bzl: fix latent bug in merging dicts or dicts

ccontavalli:
Problem:
When you have a source dict like {"key": { ... another dict ...}} where
"key" does NOT exist in the destination dict, the inner dict is copied as
is, with a simple `dict(source["key"])` pretty much (outer dict is NEW,
independent, inner DICT is shared between SOURCE and DEST).

This is BAD: if the INNER dict is later modified, the SOURCE dict is also
modified. This was caught by @salil-enf in https://github.com/enfabrica/internal/pull/37246/commits/af987146705c809679404778d36b1dccfa22d4cf,
as bazel complained about a READ ONLY / FROZEN dict being modified.

In this PR:
- make dicts always recursively copied
- update unit tests

Fun fact:
The unit test was actually catching the bug, I suspect when I originally
edited it I paid attention to the first dict and last dict, but not
the middle one - so did not notice that the second dict was also getting
changes from the first edit! Wrong!

### DIFF
--- a/bazel/utils/merge_kwargs.bzl
+++ b/bazel/utils/merge_kwargs.bzl
@@ -24,13 +24,13 @@ def merge_kwargs(d1, d2, limit = 5):
                     m[k] = list(d[k])
                     continue
 
-                if type(d[k]) == "dict":
-                    m[k] = dict(d[k])
+                if type(d[k]) != "dict":
+                    # neither list nor dict, must be scalar:
+                    m[k] = d[k]
                     continue
 
-                # type must be scalar:
-                m[k] = d[k]
-                continue
+                # Let the next if fill the dict by recursively expanding the values.
+                m[k] = {}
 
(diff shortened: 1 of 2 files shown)


jonathan-enf (approved):
Ugh.  This looks like a starlark mis-feature to me.  In python, `dict(d)` creates a copy of the dict, not a reference.

```
$ python3
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a = {1: 2, 3: 4}
>>> b = dict(a)
>>> b
{1: 2, 3: 4}
>>> b[1] = 4
>>> a
{1: 2, 3: 4}
>>> b
{1: 4, 3: 4}
>>> 
```

PR #1059, opened 2024-05-23
linux: kernel_module support for ubuntu kernels and cross compiling

cbrune:
This patch changes the definition of "kernel", which previously meant a KernelTreeInfo (a pre-built kernel tree specific to a particular CPU architecture and kernel configuration), to mean a "kernel specification".

A "kernel specification" includes:

- kernel_tree: A label for external kernel release, independent of CPU architecture and flavour.
- arch: The CPU architecture for the kernel build
- flavour: The specific kernel configuration for an architecture
- optional cross compiler info required by the CPU architecture

A helper macro is provided to help create kernel specs for end users.

'kernel_tree' + 'arch' + 'flavour' together yields a KernelTreeInfo.

The core kernel_modules_rule() rule is updated to add the cross compiler and related deps to the build.

Note: There is more to do here has the core kernel_modules_rule() rule has the notion of building multiple archs at once, which does not make sense since the kernel tree build areas are arch specific. kernel_modules_rule() can only compile a module for a single build area at a time.

The parallelism comes from the 'kernels' list passed into _kernel_module_targets(), which instantiates kernel_modules_rule() once for each kernel spec.  See the example of how this is used in the internal repo PR.

### DIFF
--- a/bazel/linux/defs.bzl
+++ b/bazel/linux/defs.bzl
@@ -12,6 +12,16 @@ def _kernel_modules(ctx):
     if not ctx.attr.archs:
         fail(location(ctx) + "rule must specify one or more architectures in 'arch'")
 
+    data_dependencies = ctx.attr.cross_compiler_data
+    tool_runfiles = []
+    for data in data_dependencies:
+        tool_runfiles += data[DefaultInfo].default_runfiles.files.to_list()
+
+    cross_compiler = ""
+    if ctx.attr.cross_compiler != None:
+        # cross compiler string expected to end with "-gcc".  strip that off
+        cross_compiler = ctx.attr.cross_compiler[DefaultInfo].files.to_list()[0].path[:-4]
+
     ki = ctx.attr.kernel[KernelTreeInfo]
     bundled = []
     for arch in ctx.attr.archs:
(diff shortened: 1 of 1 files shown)


cbrune:
See the companion internal repo PR:
https://github.com/enfabrica/internal/pull/39805

Note: neither this PR nor the enkit PR addresses the kernel release artifact creation part of the flow.  All the work was done pointing at a local kernel source tree.

bazel/linux/defs.bzl
```diff
@@ -228,16 +243,24 @@ Available format values are:
             allow_files = True,
             doc = "The list of files that constitute this module. Generally a glob for all .c and .h files. If you use **/* with glob, we recommend excluding the patterns defined by BUILD_LEFTOVERS.",
         ),
+        "cross_compiler": attr.label(
```
ccontavalli, line 246: So, bazel has a native mechanism to provide toolchains for cross compiling - which has several advantages (and of course, some drawbacks). There's a good explanation here: https://bazel.build/extending/toolchains#writing-rules-toolchains

I'm ok with a rule specific way to override a toolchain / compiler, but IMHO the default should come from the standard cpp compiler definition? `cc_common`, https://bazel.build/docs/cc-toolchain-config-reference could be pointers around that.

PR #1060, opened 2024-05-30
Create kernel_qemu_aarch64_run rule

sudhakar-mamillapalli:
Create kernel_qemu_aarch64_run rule.

* We want to be able to run a aarch64 qemu machine emulating E1 system complex which connect to the SFA FE model via netpci for dev and testing.

* So based on kernel_qemu_run which works for x86 arch,  kernel_qemu_aarch64_run target is introduced which pretty much follows the same flow as kernel_qemu_run with changes needed for aarch64.


bazel/linux/providers.bzl
```diff
@@ -77,3 +77,12 @@ RuntimeBundleInfo = provider(
         "check": "list of RuntimeInfo, executable (and its runfiles) to run OUTSIDE the VM AFTER the RUN to check if the run was successful",
     },
 )
+
+BootRamImageInfo = provider(
+    doc = """Maintains the information necessary to represent a BootRam image, needed for aarch64 qemu acf-machine boot.""",
+    fields = {
+        "name": "Name of the rule that defined this bootram image.",
```
cbrune, line 84: my gut tells me we don't need the `name` field in the provider.

if you scroll up and look at other providers in this file, none of them have a `name` field.

bazel/linux/defs.bzl
```diff
@@ -553,3 +553,44 @@ def kernel_test(name, kernel_image, module, **kwargs):
 
     # Use the UML runner for the legacy tests
     kunit_test(name, kernel_image, module, runner = kernel_uml_run, **kwargs)
+
+def _bootram_image(ctx):
+    return [DefaultInfo(files = depset([ctx.file.image])), BootRamImageInfo(
+        name = ctx.attr.name,
```
cbrune, line 559: I'm pretty sure we don't need `name` in the provider.

bazel/linux/runner.bzl
```diff
@@ -218,6 +231,8 @@ def create_runner(ctx, archs, code, runfiles = None, extra = {}):
         post = "cd {dest}; cp -L %s ./init.sh" % (shell.quote(init.short_path)),
     )
     outside_runfiles = torun.outside_runfiles.merge(ctx.runfiles([runtime_root, ki.image]))
+    if ctx.attr.bootram_image:
+       outside_runfiles = outside_runfiles.merge(ctx.runfiles([bootram_info.image]))
```
cbrune, line 235: it seems a little dirty to check for one variable (`ctx.attr.bootram_image`) and then use a value from a different variable (`bootram_info.image`).  maybe you can move the `ki.arch == aarch64` test down here?

bazel/linux/qemu.bzl
```diff
@@ -127,3 +132,102 @@ See the RuntimeBundleInfo provider for details.
         }
     ),
 )
+
+def _kernel_qemu_aarch64_run(ctx):
```
cbrune, line 136: making a copy of this rule does not feel right.  perhaps this can be parameterized and a common rule impl can be used for x86 and aarch64 ?

bazel/linux/templates/runner.aarch64_template.sh
```diff
@@ -0,0 +1,170 @@
+#!/bin/bash
```
cbrune, line 1: same thing here -- can we parameterize this script and use it for both x86 and aarch64?

bazel/linux/runner.bzl
```diff
@@ -195,6 +201,13 @@ def create_runner(ctx, archs, code, runfiles = None, extra = {}):
             ),
         )
 
+    if ki.arch == "aarch64":
```
cbrune, line 204: just a dumb note, nothing to do --- the compiler and bazel both use `aarch64`, but the Linux kernel uses `arm64`.  it so dumb these are different in different places...  i think what you have here is correct.  when writing anything bazel use `aarch64`.

bazel/linux/providers.bzl
```diff
@@ -77,3 +77,12 @@ RuntimeBundleInfo = provider(
...
+        "arch": "Architecture this image was built for.",
```
cbrune, line 85: looking at the other patch -- I don't see this `arch` field being used either.  I see a check for `if ki.arch == 'aarch64'` and then the `.image` is just used.   I suppose you could leave this `.arch` field here and in the other place verify that `ki.arch == BootRamImageInfo.arch` as a safety check.

sudhakar-mamillapalli: Added the check.

bazel/linux/templates/runner.aarch64_template.sh
sudhakar-mamillapalli: Ok did that now.

bazel/linux/qemu.bzl
sudhakar-mamillapalli: Ok did that now.

bazel/linux/runner.bzl
sudhakar-mamillapalli: Yeah.  Rearanged the code a little.

bazel/linux/providers.bzl
sudhakar-mamillapalli: Ok.  Removed it now.

sudhakar-mamillapalli:
Rebased.

bazel/linux/qemu.bzl
```diff
@@ -28,6 +33,10 @@ def _kernel_qemu_run(ctx):
 QEMU_FLAGS={qemu_flags}
 QEMU_SEARCH={qemu_search}
 KERNEL_FLAGS={kernel_flags}
+BOOTRAM_IMAGE={bootram_image}
+ARCH={arch}
+
+echo $ARCH
```
cbrune, line 39: Is this `echo` for debug?  If so, remove it.

bazel/linux/qemu.bzl
```diff
@@ -38,33 +47,50 @@ test -n "$QEMU_BINARY" || {{
...
+    ITB_IMAGE_LOADER="-device loader,file=$KERNEL,addr=0x10000000000"
 
-test -z "$KERNEL" || QEMU_FLAGS+=("-kernel" "$KERNEL")
-test -z "$SINGLE" || KERNEL_FLAGS+=("init=/bin/sh")
+    QEMU_FLAGS+="$RAW_BOOTRAM_LOADER $ITB_IMAGE_LOADER"
+
+    echo "KERNEL": $KERNEL
+    echo "BOOTRAM_IMAGE": $BOOTRAM_IMAGE
+    echo "QEMU_BINARY: $QEMU_BINARY"
+    echo "QEMU_FLAGS: $QEMU_FLAGS"
```
cbrune, line 76: Are these for debug?  I would remove them.

bazel/linux/runner.bzl
```diff
@@ -13,6 +13,12 @@ def create_runner_attrs(template_init_default):
             providers = [DefaultInfo, KernelImageInfo],
             doc = "The kernel image that will be used to execute this test. A string like @stefano-s-favourite-kernel-image, referencing a kernel_image(name = 'stefano-s-favourite-kernel-image', ...",
         ),
+       "bootram_image": attr.label(
+            doc = "A target defining the bootram image (used with aarch64 qemu).",
+	    allow_single_file = [""],
```
cbrune, line 18: minor nit.  I think the standard way to express this is:
```bazel
allow_single_file = True,
```

cbrune (approved):
Looks pretty good to me. Just a few minor nits that you could ignore.

bazel/linux/qemu.bzl
sudhakar-mamillapalli: Yeah.  Removed them.

bazel/linux/qemu.bzl
sudhakar-mamillapalli: Yeah removed it.

cbrune (approved):

bazel/linux/qemu.bzl
```diff
@@ -38,33 +45,45 @@ test -n "$QEMU_BINARY" || {{
...
 if [ ${{#WRAPPER_OPTS[@]}} -ne 0 ]; then
     WRAPPER_OPTS=('--' "${{WRAPPER_OPTS[@]}}")
 fi
 
+# Pass qemu/kernel/wrapper opts as three separate args to Qemu binary.
+
 echo 1>&2 '$' "$QEMU_BINARY" "${{QEMU_FLAGS[@]}}" "${{WRAPPER_OPTS[@]}}"
 if [ -z "$INTERACTIVE" -a -z "$SINGLE" ]; then
-    "$QEMU_BINARY" "${{QEMU_FLAGS[@]}}" "${{WRAPPER_OPTS[@]}}" </dev/null | tee "$OUTPUT_FILE"
+    "$QEMU_BINARY" "${{QEMU_FLAGS[*]}}"  "${{KERNEL_FLAGS[*]}}" "${{WRAPPER_OPTS[*]}}" </dev/null | tee "$OUTPUT_FILE"
```
ccontavalli, line 84: The original code had:
```
QEMU_FLAGS+=("-append" "${{KERNEL_FLAGS[*]}} ${{KERNEL_OPTS[*]}}")
```

... what happened to KERNEL_OPTS in the new code? Can't find anywhere the -append, and handling of KERNEL_OPTS?

sudhakar-mamillapalli: @ccontavalli  Very sorry.  Seems like I dropped KERNEL_OPTS when I mucked with args.  As for -append I pass all the args (kernel/qemu and wrapper args) separately to the model-wrapper-socket script in internal repo and do the append there (a different PR https://github.com/enfabrica/internal/pull/40251).  

The change is at 
https://github.com/enfabrica/internal/pull/40251/commits/3e8fa4754bb040fae9bb968e9e04371ef6eb6151#diff-2d20aee07523ca136cf9db6baeb3534f46d4c33e6bb9add34e88a9f63821f0eaR87


Any way thinking again,  I went back the old way of doing things since if I do the above I realized there are other model-wrapper scripts which need to be changed too.  So I just pass the arguments similar to how we use to do so before.  

Also I just use the machine name now to know if the qemu to use is aarch64 or x86 one, instead of passing it as a wrapper option.  That way do not have to muck with wrapper opts.   If you rather have the arch passed as a wapper opts, I can change that.  Please let me know. 

Also to make it easy to read I am force pushing the diff.  Please take a look again.  Sorry again for the mess-up and thanks for finding the issue.

bazel/linux/qemu.bzl
```diff
@@ -127,3 +160,49 @@ See the RuntimeBundleInfo provider for details.
...
+    # Note to change the kernel, initramfs, kernel cmdline etc. a new FIT image with
+    # necessary changes has to be created and its KernelImageInfo passed as kernel_image
+    # to this rule
+    attrs = dict(
+        create_runner_attrs(
+            template_init_default = Label("//bazel/linux:templates/init-qemu.template.sh"),
+        ),
+        **{
+            "archs": attr.string_list(
+                default = ["aarch64"],
```
skarat-enf, line 186: if we are passing the archs, do we need a separate kerenl_qemu_run rule ? Could we not just use the existing one ?

PR #1062, opened 2024-06-03
Add back rules_docker because oci_pull auth is broken

ahungrynacho:
This change adds back the `rules_docker` external repo to fix transient `oci_pull` auth issues in presubmits. After this fix is merged, presubmits will be more stable.

Tested:
- `BAZEL_PROFILE=buildbarn bazel test --override_repository=enkit=$HOME/enkit //infra/...`

JIRA: INFRA-10664


bazel/utils/container/container.bzl
```diff
@@ -120,6 +120,10 @@ def container_image(*args, **kwargs):
     output = "{}_labels.txt".format(name)
     tags = kwargs.get("tags", [])
 
+    # Remove once oci_pull doesn't have auth errors anymore
+    if kwargs.get("base", "").startswith("@"):
```
jrp-enf, line 124: I don't understand this bit.

ahungrynacho: Ok, updated the comment with a better explanation.

gpaussaenf (approved):
I reviewed the bestie/server changes. Not all that familiar with the mechanics of this change, so I'm relying on the fact that the `bazel test` command worked.

PR #1074, opened 2024-06-21
kbuild: add developer convenience defaults to build-kernel-tree.sh

cbrune:
This PR adds two defaults with flags to the `build-kernel-tree.sh`
script, which are convenient for kernel development.

1. `-c` -- perform a clean build, removing the output directory.  This
   is off by default, meaning reuse any existing output directory for
   incremental builds.  This is handy for development.
   
2. `-q` -- perform a "quiet" build.  The default is a medium noisy
   build, useful during development.
   
The kbup-astore script, which builds and publishes artifacts to
astore, using these new flags to perform a clean builds that are less
noisy.


jungleraptor (approved):
lgtm!

PR #1081, opened 2024-08-06
linux/defs.bzl: allow deps to be kernel-specific

silas-enf:
Allow the interpolation of the kernel name into dependencies in the
`kernel_module` macro, so that users can depend on only the
kernel-specific dependency module for each built module rather than the
whole bundle.

With this change, `kernel_module` targets can do the following:

```
kernel_module(
    name = "foo",
    kernels = [
        "ubuntu-impish-generic",
        "ubuntu-impish-emu-aarch64",
    ],
    deps = [
        "//driver/bar:{kernel_name}-bar",
        "//driver/baz:{kernel_name}-baz",
        ":quux", # not all deps need to specify {kernel_name}
        ...
    ],
    ...
)
```

And the kernel name will be interpolated into `deps` at `{kernel_name}`
before deps is passed to the target being generated.

Background:

Currently when a kernel_module macro invocation uses multiple values in
`kernels`, there is the problem that other `kernel_module` invocations
which need to depend on one of the corresponding modules from the first
invocation must depend on the whole bundle to ensure that dependencies
for all of its modules are available.

This means that even if downstream targets properly depend on a specific
module from the second invocation, they will still indirectly depend on
all modules from the first invocation, i.e., you have a dependency
chain like the following:

//tests:foo_test  depends on
  //mods/mod1:ubuntu-impish-generic-mod1 depends on
      //mods/mod2:mod2

where `:ubuntu-impish-generic-mod1` properly selects a single module
selected for only the kernel that's actually used by the test, but
`:mod2` selects modules for *all* kernels, only one of which is actually
used by `:foo_test`.

This is currently needed because `kernel_module` has no way to express
"use this dependency for this kernel but that dependency for that
kernel"; this patch addresses that.

### DIFF
--- a/bazel/linux/defs.bzl
+++ b/bazel/linux/defs.bzl
@@ -386,9 +386,12 @@ def _kernel_module_targets(*args, **kwargs):
         kernels.append(kwargs.pop("kernel"))
 
     if len(kernels) == 1:
-        (kernel_tree_label, unused, arch) = _normalize_kernel(kernels.pop())
+        kernel = kernels.pop()
+        (kernel_tree_label, unused, arch) = _normalize_kernel(kernel)
         kwargs["kernel"] = kernel_tree_label
-        _gen_module_rule(arch, *args, **kwargs)
+        formatted_deps = [dep.format(kernel_name = kernel) for dep in kwargs["deps"]]
+        my_kwargs = dict(kwargs, deps=formatted_deps)
+        _gen_module_rule(arch, *args, **my_kwargs)
         return
 
     targets = []
@@ -400,7 +403,9 @@ def _kernel_module_targets(*args, **kwargs):
 
         kwargs["name"] = name
         kwargs["kernel"] = kernel_tree_label
-        _gen_module_rule(arch, *args, **kwargs)
+        formatted_deps = [dep.format(kernel_name = kernel) for dep in kwargs["deps"]]
+        my_kwargs = dict(kwargs, deps=formatted_deps)
+        _gen_module_rule(arch, *args, **my_kwargs)
 
     # This creates a target with the name chosen by the user that
     # builds all the modules for all the requested kernels at once.
@@ -431,6 +436,9 @@ def kernel_module(*args, **kwargs):
             and kernel configuration to build the module against.
       kernels: list of kernel trees (same as above). kernel_module will instantiate multiple
             kernel_module_rule, one per kernel spec, and ensure they all build in parallel.
+      deps: the usual meaning, but with a twist: {kernel_name} can be added to any label, and
+            the kernel name will be interpolated into it there. This allows dependencies on other
+            `kernel_module` invocations without depending on the whole bundle.
     """
 
     if "makefile" not in kwargs:


silas-enf:
FWIW, I'd be quite happy to be told to take a better path, but I expect it'll take significantly longer. This was the minimal change to structure needed to solve the dependency problems we've been facing recently.

cbrune:
Trying to understand -- for example, take `enf-rxe`.  With this change, some of the `deps = [ ...]` would migrate to the new `kernel_specific_deps = [...]`, with `{}` format strings in them?  Makes sense.

I'm not sure the format string is needed....  Since this is all inside enkit, could the `kernel_module` rule add the format string automatically to the `kernel_specific_deps`, hiding those details from the end user?  On one side the enkit rule is creating the targets with the kernel in the name, so it seems that the enkit rule could also create deps with the kernel name in the correct location.

bazel/linux/defs.bzl
```diff
@@ -400,15 +400,17 @@ def _kernel_module_targets(*args, **kwargs):
 
         kwargs["name"] = name
         kwargs["kernel"] = kernel_tree_label
-        _gen_module_rule(arch, *args, **kwargs)
+        deps = kwargs["deps"] + [kdep.format(kernel) for kdep in kernel_specific_deps]
```
cbrune, line 403: instead of a format string here, could the rule surgically insert the kernel name in the correct location?  hiding the format string from the end user?

silas-enf: yes, by doing something like `re.replace(":", ":{kernel}")` on the label names... a bit too "magic" for my taste, especially given the need to worry about canonicalization of labels (e.g. `//foo/bar:bar` will be minified to `//foo/bar` by tooling, which would mess with that process).

Your approach would definitely be easier to maintain / require the user to understand fewer details of how `kernel_module` names things, but that knowledge is already kind of required by the current design in order to have downstream dependencies not depend on the whole bundle (case in point: my other PR on this subject).

ccontavalli: Proposal:
1. Use named format, eg, `kdep.format(kernel_name=kernel)` in your for loop.
2. Don't introduce a new parameter, just use the normal `deps` parameter?

bazel/linux/defs.bzl
```diff
@@ -400,15 +400,17 @@ def _kernel_module_targets(*args, **kwargs):
...
+        _gen_module_rule(arch, *args, **my_kwargs)
 
     # This creates a target with the name chosen by the user that
     # builds all the modules for all the requested kernels at once.
     # Without this, the user can only build :all, or the specific
     # module for a specific kernel.
     return kernel_modules_bundle(name = original, modules = targets, visibility = kwargs.get("visibility"))
 
-def kernel_module(*args, **kwargs):
+def kernel_module(kernel_specific_deps=[], *args, **kwargs):
```
ccontavalli, line 413: Please update the docstring here.

silas-enf: Update for your other comment got rid of `kernel_specific_deps`, but I added a section for `deps` to describe the special interpolation behavior.

bazel/linux/defs.bzl
silas-enf: Done in force-push.

silas-enf:
#1085 is a better solution to the same problem

PR #1086, opened 2024-08-16
fix: enkit outputs subcommands

jungleraptor:
The `enkit outputs` subcommands have been broken for a little while - likely around the time we
moved to running a bb_clientd daemon in the devcontainers.

Previously the code set up and launched a `bb_clientd` daemon in order to enable users to mount files from
CAS using an invocation ID to a scratch directory.

This PR changes the code to use the existing mount provided by the `bb_clientd` daemon that is already running in
the devcontainer. This simplifies the code considerably.

I tested:
- `enkit outputs mount -i`
- `enkit outputs unmount -i`
- `enkit outputs shutdown`


enkit/outputs/commands.go
```diff
@@ -375,28 +357,22 @@ func NewShutdown(root *Root) *Shutdown {
...
-		}
-	}
-	if err := os.RemoveAll(c.root.OutputsRoot); err != nil {
-		errs = append(errs, err)
-		c.root.Log.Errorf("error removing output root %s %v", c.root.OutputsRoot, err)
-		return multierror.New(errs)
-	}
-	fmt.Printf("Successfully deleted output root at %s \n", c.root.OutputsRoot)
-	return nil
+        dir := DefaultOutputsRoot
```
jungleraptor, line 360: I used ChatGPT for this function so might need some extra scrutiny

enkit/outputs/commands.go
```diff
@@ -64,13 +67,6 @@ func NewRoot(base *client.BaseFlags) (*Root, error) {
 		BaseFlags: base,
 	}
 
-	homeDir, err := os.UserHomeDir()
-	if err != nil {
-		return nil, fmt.Errorf("failed to detect $HOME: %w", err)
-	}
-	defaultOutputsRoot := filepath.Join(homeDir, "outputs")
-
-	rc.PersistentFlags().StringVar(&rc.OutputsRoot, "outputs-root", defaultOutputsRoot, "Root dir of mounted outputs")
```
jungleraptor, line 73: `enkit` is no longer responsible for creating the mount so there is no need to make it user configurable

enkit/outputs/commands.go
jrp-enf: I don't quite get it. The below looks like it's going to ignore the files in the top directory.

enkit/outputs/commands.go
```diff
@@ -64,13 +67,6 @@ func NewRoot(base *client.BaseFlags) (*Root, error) {
...
 	rc.PersistentFlags().StringVar(&rc.BuildBuddyApiKey, "api-key", "", "build buddy api key used to bypass oauth2")
 	rc.PersistentFlags().StringVar(&rc.BuildBuddyUrl, "buildbuddy-url", "", "build buddy url instance")
 	rc.PersistentFlags().StringVar(&rc.BuildbarnHost, "buildbarn-host", "", "host:port of BuildBarn instance")
```
jrp-enf, line 72: Are buildbarntunneltarget, tunnellistenport, and gatewayproxy below still used anywhere? If not, we should remove the flags.

enkit/outputs/commands.go
```diff
@@ -167,29 +159,19 @@ func (c *Mount) Run(cmd *cobra.Command, args []string) error {
...
-		host,
 		kbuildbarn.WithNamedSetOfFiles(),
 		kbuildbarn.WithTestResults(),
 	)
 	if err != nil {
 		return fmt.Errorf("hard links could not be generated: %w", err)
 	}
-	scratchInvocationPath := filepath.Join(bbOpts.ScratchDir(), c.InvocationID)
+
+	scratchInvocationPath := filepath.Join(DefaultOutputsRoot, c.InvocationID)
```
jrp-enf, line 174: outputsInvocationPath and scratchInvocationPath are the same path. Isn't that going to make the symlink below fail (or produce something weird?)

jrp-enf (changes requested):

enkit/outputs/commands.go
jungleraptor: Good catch!

In the original code we created hard links between the `cas` mount and a seperate `scratch` directory (the scratch directory is writable unlike the `cas` mount) (both these are fuse filesystems managed by buildbarn)

Then it created another symlink between the actual scratch, and the final output location we expose to the user:
- scratchInvocationPath: `/home/isaac/outputs/.bb_clientd/mount/scratch/<invocation-id>`
- outputInvocationPath: `/home/isaac/outputs/<invocation-id>`

When I refactored I didn't catch that I had made this the same path - and none of my testing surfaced any errors or weirdness.

We can just remove this symlink and `outputInvocationPath` altogether. It isn't necessary now that `enkit` is not responsible for setting up the mounts and needs some way to hide the `bb_clientd`  mount implementation details and present the user with a non-hidden path.

Now we already have an existing `/enf_mounts/buildbarn/scratch` mount we can point the user to.

enkit/outputs/commands.go
jungleraptor: You're right that stuff isn't used anymore - I'll remove it.

enkit/outputs/commands.go
jungleraptor: The top directory is `enf_mounts/buildbarn/scratch`: a fuse filesystem managed by `bb_clientd` - used for syncing files from CAS to a writeable location that is garbage collected between `bb_clientd` restarts.

The `enkit outputs mount` command will create a subdirectory here under `<invocation-id>` with all of the relevant files from that particular bazel invocation.

The purpose of the `enkit outputs shutdown` subcommand is to delete all of the invocation dirs that were created with `enkit outputs mount`.

jrp-enf (approved):

PR #1092, opened 2024-09-04
try to auth to bb_clientd on login

jungleraptor:
Makes it so that `enkit login` also tries to authenticate our bbclientd CAS mounts.

### DIFF
--- a/lib/client/commands/login.go
+++ b/lib/client/commands/login.go
@@ -30,9 +30,10 @@ type Login struct {
 	agent     *kcerts.SSHAgentFlags
 	populator kflags.Populator
 
-        Debug       bool
-	NoDefault   bool
-	MinWaitTime time.Duration
+        BbclientdAddress string
+        Debug            bool
+	NoDefault        bool
+	MinWaitTime      time.Duration
 }
 
 // NewLogin creates a new Login command.
@@ -58,6 +59,8 @@ func NewLogin(base *client.BaseFlags, rng *rand.Rand, populator kflags.Populator
 	}
 	login.Command.RunE = login.Run
 
+        login.Flags().StringVar(&login.BbclientdAddress, "bbclientd-address", "localhost:8981", "bbcliend address")
+        login.Flags().MarkHidden("bbclientd-address")
         login.Flags().BoolVarP(&login.Debug, "debug", "d", false, "Print extra debugging information. Mostly useful for development")
 	login.Flags().BoolVarP(&login.NoDefault, "no-default", "n", false, "Do not mark this identity as the default identity to use")
 	login.Flags().DurationVar(&login.MinWaitTime, "min-wait-time", 10*time.Second, "Wait at least this long in between failed attempts to retrieve a token")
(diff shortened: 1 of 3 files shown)


lib/client/commands/login.go
```diff
@@ -116,5 +140,24 @@ func (l *Login) Run(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+        // TODO(isaac): wrap below in a function and call via defer
```
mhallmark, line 143: I never really looked at this myself, but it looks like defered funcs do not get called https://stackoverflow.com/questions/63257505/does-a-deferred-restore-run-if-you-ctrl-c-out-of-an-app

I'm curious the reason to run that defered though - is the idea to just run it async (seeing that we don't really care about any errors that may pop up)? If so, I'd recommend using a goroutine instead.

lib/client/commands/login.go
```diff
@@ -116,5 +140,24 @@ func (l *Login) Run(cmd *cobra.Command, args []string) error {
...
+        // will that still execute if the user ctrl-c's ?
+        grpclog.SetLoggerV2(grpclog.NewLoggerV2(os.Stdout, os.Stderr, os.Stderr))
+        grpc.EnableTracing = true
+        conn, err = grpc.Dial("localhost:8981", grpc.WithInsecure(), grpc.WithUnaryInterceptor(tokenAuthInterceptor(enCreds.Token)))
```
mhallmark, line 147: Hmm, didn't think about this, but this port will always need to be assumed `8981` right? Since we aren't going to add flags for this?

lib/client/commands/login.go
```diff
@@ -116,5 +140,30 @@ func (l *Login) Run(cmd *cobra.Command, args []string) error {
...
```
jrp-enf, line 147: You probably want to add grpc.WithTimeout(2 * time.Second).

lib/client/commands/login.go
jungleraptor: It looks like the rest of the code just assumes the user isn't going to ctrl-c before the other tasks complete (write the file to `~/.config`, etc..), so probably overly cautious here.

lib/client/commands/login.go
jungleraptor: This occured to me as well.

We're creating an implicit coupling between enkit login and bbclientd...

I thought about making it so that this value comes from a build time variable - and then we update the bb_clientd config to also use that variable (so they stay in sync).

Probably worthwhile since I can't say for certain when we'll remove this in favor of setting up credential helper support.

lib/client/commands/login.go
jungleraptor: done

lib/client/commands/login.go
```diff
@@ -116,5 +191,12 @@ func (l *Login) Run(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+        // Reuse the token to authenticate our bbclientd CAS mounts
+        err = AuthenticateBbclientd(enCreds.Token)
+
+        if err != nil {
```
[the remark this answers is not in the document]
jrp-enf: bb_clientd doesn't exist in penguin, so you'd always be printing a warning there, which is a frequent use-case. I think this is a convenience/best-effort function.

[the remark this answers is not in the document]
jungleraptor: That's a great point!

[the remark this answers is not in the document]
jungleraptor: I snooped around my system and didn't see any enkit specific logs so 🤷🏼

jrp-enf (approved):

lib/client/commands/login.go
jungleraptor: It turns out we have mechanisms to push overrides to the default flag values (without rebuilding/deploying enkit)

So if we make this a cobra flag we can change it if we change the address the daemon runs at.

PR #1094, opened 2024-09-05
kernel_module - add paths to extra flags

marcelo-enf:
Currently, the Makefiles for kernel modules infer the bin directory from the include flags, or set it to be `bazel-out/k8-fastbuild/bin`. Both methods are brittle and the second one is incorrect if more than one platform is targeted.

This change will allow kernel modules to point to the right directories during compilation without having to resort to such brittle methods.

### DIFF
--- a/bazel/linux/defs.bzl
+++ b/bazel/linux/defs.bzl
@@ -94,6 +94,9 @@ def _kernel_modules(ctx):
     if extra_symbols:
         extra.append("KBUILD_EXTRA_SYMBOLS=\"%s\"" % (extra_symbols))
 
+    extra.append('BAZEL_BIN_DIR="%s"' % ctx.bin_dir.path)
+    extra.append('BAZEL_GEN_DIR="%s"' % ctx.genfiles_dir.path)
+
     if ctx.attr.silent:
         silent = "-s"
     else:


bazel/linux/defs.bzl
```diff
@@ -94,6 +94,9 @@ def _kernel_modules(ctx):
     if extra_symbols:
         extra.append("KBUILD_EXTRA_SYMBOLS=\"%s\"" % (extra_symbols))
 
+    extra.append('BAZEL_BIN_DIR="%s"' % ctx.bin_dir.path)
+    extra.append('BAZEL_GEN_DIR="%s"' % ctx.genfiles_dir.path)
```
ccontavalli, line 98: how is BAZEL_GEN_DIR used in the Makefile? Maybe it should just append -I cflags correctly directly, like on line 132? 

Basically: if there's a way to get the Makefile / build do the right thing without adding knowledge of bazel magic in the Makefile, that would be strongly preferred. If not, this patch looks great.

ccontavalli (approved):

bazel/linux/defs.bzl
marcelo-enf: The problem is that we need to point to the correct place of files to compile are. It is not a matter of include directories.

PR #1097, opened 2024-09-06
allocation manager first look

kjw-enf:
- At this point, the server compiles, and the client seems to reserve a dummy entry and release without crashing. Much more left to do, however. but commit'ing to checkpoint progress.


roivanov (approved):
lgtm

PR #1102, opened 2024-09-11
gee: update support for bcompare

jonathan-enf:
A user had an issue where they ran `gee config enable_bcompare` but didn't
notice the warning message and instructions indicating that `bcompare` still
needed to be installed in their container. This PR fixes that issue by failing
to enable bcompare if bcompare is not installed (an error instead of a
warning).

While we're here, let's update to bcompare5 anyway.  Dark mode!

Tested: manually tested by creating a merge conflict and invoking the gui
mergetool to resolve it.


jonathan-enf:

hanu-enf (approved):

PR #1104, opened 2024-09-13
Added external dependency to boringssl repository and added patch to fix compilations errors observed on ARM64 architecture.

naveenaketi-enf:
Tested the boringssl repo changes for ARM64 and X86 based compilation and no issues were observed.

Verified that the enkit repo changes were picked up and the boringssl related patch got applied successfully.

Testing commands:

- clear && BAZEL_PROFILE=local bazel build
--override_repository=enkit=/home/naketi/gitrepos/enkit
//model/functional/model_test:model-grpc-e1
- clear && BAZEL_PROFILE=local bazel build
--override_repository=enkit=/home/naketi/gitrepos/enkit
//model/functional/model_test:model-grpc

### DIFF
--- a/bazel/init/stage_1.bzl
+++ b/bazel/init/stage_1.bzl
@@ -183,12 +183,12 @@ def stage_1():
         patches = [
             "@enkit//bazel/dependencies/boringssl:0001-ignore-pedantic-warnings-and-move-hrss-polynomial-declarations-under-x64-flag.patch",
         ],
-	sha256 = "534fa658bd845fd974b50b10f444d392dfd0d93768c4a51b61263fd37d851c40",
-	strip_prefix = "boringssl-b9232f9e27e5668bc0414879dcdedb2a59ea75f2",
-	urls = [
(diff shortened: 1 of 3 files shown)


bazel/init/stage_1.bzl
```diff
@@ -176,6 +176,21 @@ def stage_1():
         ],
     )
 
+    maybe(
```
ccontavalli, line 179: can you also put a comment just before grpc_deps() is invoked stating something like:
```
# IMPORTANT: grpc_deps() pulls in boringssl as a WORKSPACE dependency. In order to apply patches
# to boringssl, we define boringssl BEFORE grpc_deps is invoked - that way, our version will be picked
# over the default one.
# You must manually update boringssl when grpc is updated. If ARM support is added upstream, we may
# be able to remove the patches and the work here.
```

ccontavalli (approved):

bazel/init/stage_1.bzl
naveenaketi-enf: Done - https://github.com/enfabrica/enkit/pull/1104/files#diff-59c9ccc93dd10c39d3c243e5bf9d2a6cf5b596171078a3a12b052cc257b106b1R87

PR #1111, opened 2024-09-23
Reconfig server to listen on port :6435

kjw-enf:
- Customize port from lib/server/run.go default to :6435


roivanov (approved):
Wikipedia [says](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers) it's ok

moksh-enf (approved):

PR #1116, opened 2024-09-27
gee codeowners: fix bug introduced by multi-master feature

jonathan-enf:
While testing `gee` I noticed that `gee codeowners` was producing an incorrect
list of reviewers.  The root cause turned out to be PR #1088, which allows
for arbitrary upstream branches to be used for any given chain of branches.

As a result, `gee codeowners` was looking at the set of files that had changed
since "master" when the branch was actually a clone of "upstream/master" or
"upstream/master_a0".

This PR fixes the issue by refactoring out the code used by `gee pr_make` to
determine the correct upstream branch for a chain of PRs, and using that
upstream branch to determine the set of changed files to apply against the
CODEOWNERS rules.

I also added an `--upstream` option to `gee get_parent` to make it easier
to report the "ultimate upstream parent" information.

Tested: Tested independently in a enfabrica/internal PR.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -3506,13 +3506,21 @@ EOT
 
 function gee__get_parent() {
   _startup_checks "get_parent"
-
   _check_cwd
-  local BRANCH
-  BRANCH="${1:-$(_get_current_branch)}"
+
+  local BRANCH PARENT
   _read_parents_file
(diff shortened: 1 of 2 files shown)


jonathan-enf:

mhallmark (approved):

PR #1117, opened 2024-09-27
add metrics

roivanov:


allocation_manager/service/unit.go
```diff
@@ -46,18 +49,56 @@ metricLicenseReleaseReason = promauto.NewCounterVec(prometheus.CounterOpts{
...
+func init() {
+	fmt.Println("In init")
+	prometheus.MustRegister(unitCounter)
+}
+
+func (u unit) DoOperation(operationName string) {
+	// name := u.name
+	name := u.Topology.Name
+	unitCounter.With(prometheus.Labels{
+		"kind": fmt.Sprintf("%T", u),
```
kjw-enf, line 71: The risk with using %T (i.e. "service.unit") is in case of refactoring, the metric labels will change which is not desirable. Also, it arguably duplicated in the counter name "unit_..."

allocation_manager/service/unit_test.go
```diff
@@ -142,4 +142,23 @@ func TestForget(t *testing.T) {
 	assert.Equal(t, 0, u.Forget("idX"), "u.Forget second try")
 }
 
+func TestNew(t *testing.T) {
```
kjw-enf, line 145: Looks like this test is a work-in-progress?

allocation_manager/service/unit.go
```diff
@@ -46,18 +49,56 @@ metricLicenseReleaseReason = promauto.NewCounterVec(prometheus.CounterOpts{
...
+func (u *unit) Release(inv *invocation) bool {
+	u.DoOperation("release")
+	return true
+}
+
 // Allocate attempts to associate the supplied invocation with this Unit.
 // Returns whether this Unit was successfully allocated.
 func (u *unit) Allocate(inv *invocation) bool {
 	defer u.updateMetrics()
+	defer u.DoOperation("allocate")
```
kjw-enf, line 101: This will increment in case of the "return false" case. i.e. if the user issues a second Allocate rpc. It should be below the if on the next line.

allocation_manager/service/unit.go
```diff
@@ -46,18 +49,56 @@ metricLicenseReleaseReason = promauto.NewCounterVec(prometheus.CounterOpts{
...
 
+func NewUnit(topo apb.Topology) *unit {
+	u := new(unit)
+	u.Topology = topo
+	fmt.Println("New unit created", topo)
+	return u
+}
+
+// Returns whether this Unit was successfully released.
+func (u *unit) Release(inv *invocation) bool {
```
kjw-enf, line 92: I guess it wasn't obvious; the Release rpc calls Forget() below.

kjw-enf: Also, ExpireAllocations() is another way that Units are released, but it's not caused by client action (rather by inaction).

allocation_manager/service/unit.go
```diff
@@ -46,18 +49,56 @@ metricLicenseReleaseReason = promauto.NewCounterVec(prometheus.CounterOpts{
 	},
 )
 */
+
+unitCounter = prometheus.NewCounterVec(
+	prometheus.CounterOpts{
+		Name: "unit_operations_total",
```
kjw-enf, line 55: we should probably prefix all our metric names with allocation_manager_ like how service.go does, i.e. allocation_manager_request_duration_seconds

allocation_manager/server/example.md
```diff
@@ -0,0 +1,53 @@
...
+/tmp/server --service_config=allocation_manager_config.textproto
+
+# where do logs go?
+
+# debug protos
+pushd bazel-bin/allocation_manager/client/client.runfiles
+env PYTHONPATH=com_google_protobuf/python:enfabrica:$(echo $(find . -name site-packages | fmt -1 ) | tr ' ' : ) python3
+from allocation_manager.proto import allocation_manager_pb2 as apb
+
+bazel build //allocation_manager/client/allocation_manager_client:allocation_manager_client && \
```
kjw-enf, line 52: client docs belong in client/readme.md

allocation_manager/server/example.md
```diff
@@ -0,0 +1,53 @@
...
+
+Optionally, have example allocation code in `main`:
+
+```go
+	topo := apb.Topology{
+		Name: "Unit Name 2", Config: "Unit Config",
+	}
+	a := service.NewUnit(topo)
+
+	a.DoOperation("allocate")
```
kjw-enf, line 21: The RPC handlers in service.go use the same labels, but capitalized (i.e. Allocate). We should harmonize the labels; using both would cause confusion.

kjw-enf (commented):
Aside: Having found https://github.com/grpc-ecosystem/go-grpc-middleware/blob/main/providers/prometheus/server_metrics.go#L27 I think we should consider replacing service.go metric "allocation_manager_response_count" with that.

allocation_manager/service/unit_test.go
roivanov: Yes, I have to figure out how to unit tests Prometheus metrics

kjw-enf: I think we can trust that Prometheus publishes metrics. just checking that the counters are incremented as expected should be good enough. i.e. testing our business logic (that the counter incrementing is in the right place) should be good enough.

roivanov: Do you want me to remove this test? Does it seem useless?

allocation_manager/server/example.md
roivanov: I was looking if there is a way of getting the current function name as a string, but it does not seem very easy, although still doable. Afair, there were need for `reflection`

allocation_manager/service/unit_test.go
roivanov: removed the test:
```
allocation_manager/service/unit_test.go:175:31: metric.Get undefined (type prometheus.Counter has no field or method Get)
```

kjw-enf:
For future reference, looks like they recommend use of https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/testutil

PR #1119, opened 2024-10-01
Add yaml handling to golang server and client.

kjw-enf:
- Add yaml and bundle handling to go client.
- Add yaml  handling to go server.


roivanov (approved):
lgtm

PR #1120, opened 2024-10-01
Fix help to indicate argument placement for enkit astore download

kjw-enf:
- Fix help to indicate argument placement for enkit astore download


kjw-enf:
```
$ bazel-bin/enkit/enkit_/enkit  astore download -h
Downloads an artifact

Usage:
  enkit astore download <path|uid>... [flags]
```

the "<path|uid>..." string is what I added

ccontavalli (approved):

PR #1124, opened 2024-10-10
fix(allocation_manager): Fix server build target name so that the binary filename isnt just "server"

kjw-enf:
- fix(allocation_manager): Fix server build target name so that the binary filename isn't just "server"


roivanov (approved):
lgtm,
It might be worth having the same/shared binary for the client and server. When it is a server, start with option `-d` or use the command `allocation_manager server <options and arguments>`. Same for the client `allocation_manager client <options and arguments>`.  And `client` could be a default command when one is not specified.

PR #1125, opened 2024-10-15
gee vimdiff: now uses "git difftool"

jonathan-enf:
Previously, "gee vimdiff" tried to be clever and just run vimdiff to diff files
within two work trees.  This fails if the parent is a remote branch, which is
now the default behavior for gee 0.2.51.

To fix this, gee no longer tries to be clever and instead relies on the "git
difftool" command internally.  Additionally, "gee vimdiff" is now renamed "gee
difftool", though the old command is still supported for backwards
compatibility.  Additionally, instead of forcing use of vimdiff, the currently
configured difftool is used.

This PR also includes some `gee config` cleanup: the user is allowed to select
between vim and neovim as a difftool.  `gee config default` will now select
`gee config enable_nvim` if nvim is installed, otherwise `gee config
enable_vim` is selected.

Tested: Manually ran `gee difftool -- scripts/gee`

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -2416,17 +2416,22 @@ Usage: gee config <option>
 Valid configuration options are:
 
 * "default": Reset to default settings.
-* "enable_vim": Set "vimdiff" as your merge tool.
-* "enable_emacs": Set "emacs" as your merge tool.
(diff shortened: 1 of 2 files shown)


jonathan-enf:

mhallmark (approved):

PR #1131, opened 2024-10-28
[SF-1038] Add support for zlib dependency

justus-enf:
This PR adds zlib as a dependency. The default behavior is to use the zlib installed on the system. However, we may need to build zlib when cross compiling.

### DIFF
--- a/bazel/dependencies/libz.BUILD.bazel
+++ b/bazel/dependencies/libz.BUILD.bazel
@@ -1,0 +1,41 @@
+load("@rules_foreign_cc//foreign_cc:defs.bzl", "cmake")
+
+filegroup(
+    name = "libz_all",
+    srcs = glob(["**"]),
+    visibility = ["//visibility:public"],
+)
+
+filegroup(
+    name = "libz_headers",
+    srcs = glob(["**/*.h"]),
+    visibility = ["//visibility:public"],
+)
+
+
+# zlib
+# The name of this rule is libz to avoid conflicts with the system `@zlib//:zlib`.
+# We cannot use @zlib//:zlib because the libraries that are generated are named
+# libzlib.(a|so), which is incompatible with other foreign_cc builds.
+cmake(
+    name = "libz", 
+    env = select({
+        "@//bazel/platforms:e1_gnu_build" : {
+            "CHOST": "arm",
+            "LDFLAGS": "-fuse-ld=gold",
+        },
+        "//conditions:default" : {}
+    }),
+    cache_entries = {
+        "CMAKE_C_FLAGS": "${CMAKE_C_FLAGS:-} -fPIC",
+        "CMAKE_CXX_FLAGS": "${CMAKE_CXX_FLAGS:-} -fPIC",
+    },
+    lib_source = ":libz_all",
+    generate_args = [],
+    targets = ["", "install"],
+    build_args = ["--parallel $$(nproc)"],
+    out_static_libs = ["libz.a"],
+    out_shared_libs = ["libz.so"],
+    includes = ["."],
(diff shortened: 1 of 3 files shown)


bazel/dependencies/libz.BUILD.bazel
```diff
@@ -0,0 +1,41 @@
...
+# zlib
+# The name of this rule is libz to avoid conflicts with the system `@zlib//:zlib`.
+# We cannot use @zlib//:zlib because the libraries that are generated are named
+# libzlib.(a|so), which is incompatible with other foreign_cc builds.
+cmake(
+    name = "libz", 
+    env = select({
+        "@//bazel/platforms:e1_gnu_build" : {
+            "CHOST": "arm",
+            "LDFLAGS": "-fuse-ld=gold",
```
jungleraptor, line 25: Hopefully we can remove this once my `rules_foreign_cc` patch goes in?
https://github.com/enfabrica/enkit/pull/1129

Also curious about `CHOST` arg.

bazel/dependencies/libz.BUILD.bazel
```diff
@@ -0,0 +1,41 @@
...
+        "//conditions:default" : {}
+    }),
+    cache_entries = {
+        "CMAKE_C_FLAGS": "${CMAKE_C_FLAGS:-} -fPIC",
+        "CMAKE_CXX_FLAGS": "${CMAKE_CXX_FLAGS:-} -fPIC",
+    },
+    lib_source = ":libz_all",
+    generate_args = [],
+    targets = ["", "install"],
+    build_args = ["--parallel $$(nproc)"],
```
jungleraptor, line 36: Using `$$(nproc)` in a build is non-hermetic - we should avoid this.

bazel/dependencies/libz.BUILD.bazel
```diff
@@ -0,0 +1,41 @@
...
+    visibility = ["//visibility:public"],
+)
+
+
+# zlib
+# The name of this rule is libz to avoid conflicts with the system `@zlib//:zlib`.
+# We cannot use @zlib//:zlib because the libraries that are generated are named
+# libzlib.(a|so), which is incompatible with other foreign_cc builds.
+cmake(
+    name = "libz",
```
jungleraptor, line 21: I'd prefer to go with a bazel native build of zlib (and libpng for that matter) - then using the foreign_cc rules.

bazel/dependencies/libz.BUILD.bazel
justus-enf: https://github.com/madler/zlib/blob/d476828316d05d54c6fd6a068b121b30c147b5cd/configure#L34

`CHOST` might not be needed since I changed over to cmake build. I don't see it used there.

bazel/dependencies/libz.BUILD.bazel
justus-enf: I understand the pros of this preference. However, I think that is quite a lot of work without a driving need. My feeling on this is:

1. We are already using foreign_cc rules for other dependency
2. It is a lot of work to completely recreate all the configure logic for every third party dependency

I am happy to go in that direction if there is a consensus that we need to change our approach for dependencies moving forward.

bazel/dependencies/libz.BUILD.bazel
justus-enf: I will test with your change.

bazel/dependencies/libz.BUILD.bazel
justus-enf, line 36:
```suggestion
```

jungleraptor (approved):

PR #1133, opened 2024-10-29
fix(bazel): make kernel module builds use gcc-13

jungleraptor:
`kernel_module` targets were still silently building against our previous aarch64-linux cross compiler based on `bazel_embedded` instead of our new hand rolled one.

To fix this I've moved the `gcc-13` constraint into this repo so and made it a constraint of the `kernel-aarch64` platform.

This aligns the constraints setting of all the aarch64-linux platforms.


justus-enf (approved):

ccontavalli (approved):

cbrune (commented):
nice

PR #1136, opened 2024-11-14
Update googletest to v1.15.2

RobSpringer:
Despite the branch name, this does what it says in the title.

Along the way, I found that it requires an explicit specification of C++ version >= C++14, so I added that to the .bazelrc.

All tests under [internal] //ib, //driver, and //sdk perform as before this change if I incorporate it in the main repository (with one necessary change to one line of code).

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -3,7 +3,8 @@
 # see https://docs.bazel.build/versions/master/guide.html#bazelrc
 #
 # The full list of Bazel options: https://docs.bazel.build/versions/master/command-line-reference.html
-build --cxxopt="-std=c++17"  --host_cxxopt="-std=c++17" --client_env=BAZEL_CXXOPTS=-std=c++17
+common --repo_env=CC=gcc-10 --repo_env=CPP=g++-10 --repo_env=CXX=g++-10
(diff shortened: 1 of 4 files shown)


.bazelrc
```diff
@@ -3,6 +3,7 @@
 # see https://docs.bazel.build/versions/master/guide.html#bazelrc
 #
 # The full list of Bazel options: https://docs.bazel.build/versions/master/command-line-reference.html
+build --cxxopt="-std=c++17"  --host_cxxopt="-std=c++17" --client_env=BAZEL_CXXOPTS=-std=c++17
```
jungleraptor, line 6: We probably don't have much C++ code in this repo but lets just set c++20 here so we're consistent?

jungleraptor: Also where did you get `--client_env` from? I'm not familiar with that flag.

I think it's redundant anyway since you specified `--cxxopt` - but I *think* you want `repo_env` here because this environment value is accessed as part of the evaluation of the repository rule that sets up bazel's default c++ toolchain.

RobSpringer: SG! I didn't want to be too greedy on the first stab. If I don't move to g++-10, it complains about c++2a vs. c++20. I added the g++ line, your call if we should go in a different direction.

ccontavalli (approved):

jungleraptor (approved):

PR #1138, opened 2024-11-18
go: Tidy go.mod, make BUILD files gazelle-clean

minor-fixes:
This change:

* performs a `go mod tidy` and arranges dependencies into two `require` sections: one containing only direct deps, and the other containing only indirect deps. Dependency versions are inadvertently updated in the process.
* makes the repo "gazelle-clean" (`bazel run //:gazelle` succeeds with no errors, and generates no diffs). In order to do this, missing `empty.go` markers were added to proto dirs missing them, and a Gazelle directive was added to exclude all `empty.go` files from generated rules.
* adds `keep` directives to some `go_embed_data` targets; Gazelle seems to want to remove `importpath` attributes when it removes `empty.go` from `srcs` and then finds `srcs` empty. Though this is a bit of a hack, `go_embed_data` targets will need to go in the (near) future, as this rule is dropped from future versions of rules_go.
* removes the `go_rbe` target that embeds the remote-apis codegen, in favor of a Gazelle directive that resolves to the original target; this may avoid duplicate linking issues in the future.
* removes the `enfuse` dir, as it contains no code

Tested: Some targets build locally (testing is limited due to ACL issues pulling container images)


minor-fixes:
Some of the [container images pulled](https://github.com/enfabrica/enkit/blob/6bacccf18723655d8e896aed65638777b5c747ce/bazel/init/stage_4.bzl#L28) are private, which makes it hard to build locally. Can these be made public, or changed to refer to publicly-available images?

minor-fixes:
cc @mhallmark (I cant add reviewers)

mhallmark (approved):
Built and ran all tests from my end. There's currently work being done to remove the requirement for those base images.

minor-fixes:
Do you want to merge when you're ready? I don't have a button for that either

image

mhallmark:
@minor-fixes  Done - Thanks for the work!

PR #1144, opened 2024-12-06
gee gcd: improve gcd to fail if target branch isn't specified

jonathan-enf:
Previously:

    gee gcd -B upstream/master

would fail with an obscure error "git worktree" generated error message.

Now, it fails with a clear error indicating that the target branch name
argument was unspecified.

Tested: manually tested.


jonathan-enf:

mhallmark (approved):

PR #1152, opened 2025-01-11
auth/client and server: adjust timeout handling.

ccontavalli:
Currently, the auth server and client relies on long lived connections
(up to 30 minutes per connection).

With the default appengine connection limit of max_concurrent_requests of 10,
this means if 10 users do enkit login and walk away from the connection, the
server is hosed for ~30 minutes.

Although this limit was increased to the max a while back in our deployment,
let's start rolling out changes so we can start using short lived connections
instead (deployment requires users moving to a new client, which will take time).

Additionally, we move the client side timeout and retry handling to be
configurable with our default flag infrastructure, which will allow us
to tune those values via flag push on astore server without asking users for an
upgrade in the future.

See individual commits for the specific changes.

- lib/clients/command: run go fmt
- lib/clients/command/login.go: read retry parameters from flags.
- auth/server (and client): change default retry and sleep parameters.

### DIFF
--- a/lib/client/commands/login.go
+++ b/lib/client/commands/login.go
@@ -26,14 +26,15 @@ type Login struct {
 	*cobra.Command
 	rng *rand.Rand
 
-	base      *client.BaseFlags
-	agent     *kcerts.SSHAgentFlags
+	base  *client.BaseFlags
+	agent *kcerts.SSHAgentFlags
+	retry *retry.Flags
+
 	populator kflags.Populator
 
 	BbclientdAddress string
 	Debug            bool
 	NoDefault        bool
-	MinWaitTime      time.Duration
 }
 
 // NewLogin creates a new Login command.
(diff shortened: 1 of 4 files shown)


lib/client/commands/login.go
```diff
@@ -54,39 +55,61 @@ func NewLogin(base *client.BaseFlags, rng *rand.Rand, populator kflags.Populator
...
+	//
+	// A value of 1800 retries, will allow about 30 mins to complete auth assuming
+	// the server keeps the connection open for about 1 second.
+	login.retry.AtMost = 1800
+	// A value of 10 seconds (pre-2025 default) means that the minimum connection
+	// open time on the server is also ~10 seconds, otherwise the user would see
+	// the client wait for ~10 seconds between attempts to retrieve the token.
+	// Set this (the minimum wait time) to a much more aggressive value.
+	login.retry.Wait = 1 * time.Second
+
```
jrp-enf, line 82: This feels too aggressive. You're typically waiting for a user to click a link, which they might not do immediately.

jrp-enf (approved):

lib/client/commands/login.go
ccontavalli: I'll move forward with merging, if you disagree I'm happy to open a PR adjusting this further.

PR #1194, opened 2025-04-07
gee update: fix behavior when updating PR branches

jonathan-enf:
`gee up` was failing for some branches created with `gee pr_checkout`, with error
messages along the lines of "upstream commit was not found."

`gee` was previously performing the following steps:

```
git fetch upstream
git ls-remote upstream refs/pull/12345/head
```

The first command was intended to ensure that the local image of the upstream
repo was up to date, and the 2nd command was designed to look up the commit
associated with the upstream pull request.

The problem was: `git fetch upstream` was only fetching commits associated
with branches (ie. master), but wasn't fetching commits associated with
references.  So: ls-remote was returning the SHA of a commit that `git fetch`
had not fetched.

The simplest fix is to just tell git to fetch what we want it to fetch.
The above two commands are now replaced with:

```
git fetch upstream refs/pull/12345/head  # updates FETCH_HEAD
git show -s --pretty=oneline FETCH_HEAD  # queries FETCH_HEAD
```

Tested: manually tested in my local pr branches.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -3323,8 +3323,8 @@ function _get_parent_head_commit() {
   local -a WORDS=()
   if [[ "${PARENT}" == */* ]]; then
     # make sure remote is up to date:
-    _git fetch "${PARENT%%/*}"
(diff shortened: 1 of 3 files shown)


jonathan-enf:

mhallmark (approved):

PR #1198, opened 2025-04-10
gee lsbr: fix rev-list counting of branches derived from upstream/master

jonathan-enf:
Udi found a bug in gee, such that gee was reporting errors while running `gee lsbr`
that looked like this:

```
$ gee lsbranches
CMD: /usr/bin/git fetch
all_traffic         : 1 ahead, 248 behind master
cleanup             : 1 ahead, 1657 behind master
/home/udi/bin/gee: line 1460: ((: == 0 : syntax error: operand expected (error token is "== 0 ")
counters            :  ahead,  behind upstream/master
```

The issue turned out to be: the way gee was parsing "upstream/master" when constructing
the fetch request for a parent branch.  I fixed the parsing, the error went away, and
I made some other improvements to the way lsbr reports information that will make it
easier to catch this issue if something similar arises in the future.

Tested:

I confirmed that I could reproduce the problem locally, and then confirmed
that my fix resolved the issue.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -2,7 +2,7 @@
 # shellcheck disable=SC2034
 # (due to https://github.com/koalaman/shellcheck/issues/817)
 
-readonly VERSION="0.2.55"
+readonly VERSION="0.2.56"
 
 if read -r -d '' USAGE <<'EOT'
 . __ _  ___  ___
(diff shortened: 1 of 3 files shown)


jonathan-enf:

mhallmark (approved):

PR #1202, opened 2025-04-23
gee pr_checks: handle old and new build URL parsing

jonathan-enf:
The URL that gee parses to extract the build number recently changed.

This is a point fix that lets gee handle both the old and new URL
syntax, so that both old and new PRs will work.

Also, Rolf pointed out that "_error" isn't defined.  I could've sworn
it used to be there.  Adding it (back?).

also deleting a spurious version of gee that seems to have been dropped
into the wrong directory.

Tested: manually tested.


jonathan-enf:

mhallmark (approved):

PR #1204, opened 2025-04-24
gee: fix runaway rebase operations

jonathan-enf:
A month ago, I introduced a bug into gee that caused an infinite loop while
doing a `gee update` operation.

The problem turned out to be a change to the `_is_rebase_in_progress` function
that I learned about on stackoverflow.  Turns out the author was wrong, and
that the function would sometimes return false positives.

I reverted to my earlier implementation of this function to fix the issue.

Tested:

I had a branch that reproduced the issue.  I used `git reset --hard branch.REBASE_BACKUP` to
go back to the initial state, and re-tried `gee up` with the patched gee.  The problem was
resolved.

### DIFF
--- a/scripts/gee
+++ b/scripts/gee
@@ -1612,7 +1612,7 @@ function _interactive_conflict_resolution() {
     SKIP=0
     RESTART=0
     if [[ "${#STATUS[@]}" -eq 0 ]]; then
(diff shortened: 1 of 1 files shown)


jonathan-enf:

mhallmark (approved):

PR #1209, opened 2025-05-02
bestie/utils: add C++ library to plumb google bench with bestie.

ccontavalli:
In short:
This commit provides a utility class and a few functions to allow
google bench library to save all metrics in a format that is
understood by bestie.

### DIFF
--- a/bestie/utils/googlebench/bestie_reporter.cc
+++ b/bestie/utils/googlebench/bestie_reporter.cc
@@ -1,0 +1,177 @@
+#include "bestie/utils/googlebench/bestie_reporter.hh"
+
+#include <google/protobuf/text_format.h>
+
+#include <chrono>
+#include <fstream>
+
+#include "google/protobuf/io/zero_copy_stream.h"
+#include "google/protobuf/io/zero_copy_stream_impl.h"
+
+namespace bestie {
+
+std::string MetricsPath(const std::string_view filename, std::string_view path) {
+  return absl::StrCat(path, "/", filename, ".metrics.pb");
+}
+
+std::string MetricsPath(const std::string_view filename) {
+  // getenv is not thread safe, static variable initialization is.
+  // Also, try to discourage changing the environment variables at run time.
+  static const std::string path([]() {
+    if (const char* path = std::getenv("TEST_UNDECLARED_OUTPUTS_DIR"); path) return path;
+    if (const char* path = std::getenv("TMPDIR"); path) return path;
+
+    return "/tmp";
+  }());
+
+  return MetricsPath(filename, path);
+}
+
+std::string MetricsPath(int attempt, const std::string_view filename) {
+  if (attempt <= 0) return MetricsPath(filename);
+  return MetricsPath(absl::StrFormat("%s%03d", filename, attempt));
+}
+
+static void AddTag(bestie::proto::TestMetric* metric, const std::string& key,
+                   const std::string& value) {
+  auto* tag = metric->add_tags();
+  tag->set_key(key);
+  tag->set_value(value);
+}
+
+static bestie::proto::TestMetric* AddMetric(bestie::proto::TestMetrics* metrics,
+                                            const std::string& name, double value,
+                                            const auto&... prototype) {
+  auto* metric = metrics->add_metrics();
+
+  metric->set_metricname(name);
+  metric->set_value(value);
+
+  (metric->MergeFrom(prototype), ...);
+
+  return metric;
+}
+
+bool Reporter::ReportContext(const Context& context) {
+  int64_t epoch = std::chrono::duration_cast<std::chrono::nanoseconds>(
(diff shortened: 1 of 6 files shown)


bestie/utils/googlebench/bestie_reporter_test.cc
```diff
@@ -0,0 +1,735 @@
...
+    key: "context__sys_info__name"
+    value: "<hostname>"
+  }
+  tags {
+    key: "context__executable_name"
+    value: "<path-of-binary>"
+  }
+}
+metrics {
+  metricname: "run__maxheapbytes_used"
```
jacob-adelmann, line 157: Out of curiosity, why is this block replicated twice in kExpectedMessageText:

```
metrics {
  metricname: "run__maxheapbytes_used"
  tags {
    key: "run__benchmark_name"
    value: "Message"
  }
  tags {
    key: "run__skipped"
    value: "message"
  }
  tags {
    key: "run__skip_message"
    value: "For the powerful, crimes are those that others commit."
  }
  tags {
    key: "context__sys_info__name"
    value: "<hostname>"
  }
  tags {
    key: "context__executable_name"
    value: "<path-of-binary>"
  }
}
```

bestie/utils/googlebench/bestie_reporter.cc
```diff
@@ -0,0 +1,180 @@
...
+
+bool Reporter::ReportContext(const Context& context) {
+  int64_t epoch = std::chrono::duration_cast<std::chrono::nanoseconds>(
+                      std::chrono::system_clock::now().time_since_epoch())
+                      .count();
+
+  context_.Clear();
+  context_.set_timestamp(epoch);
+
+  AddTag(&context_, "context__sys_info__name", context.sys_info.name);
```
jacob-adelmann, line 63: Is context.sys_info.name the hostname the benchmark is run on?

jacob-adelmann: I think yes based on details later in the test

bestie/utils/googlebench/bestie_reporter.cc
```diff
@@ -0,0 +1,180 @@
...
+  AddTag(&context_, "context__executable_name", context.executable_name);
```
jacob-adelmann, line 64: Is context__executable_name just the executable name with the path stripped, or does it contain absolute or relative path?

bestie/utils/googlebench/bestie_reporter.cc
```diff
@@ -0,0 +1,180 @@
...
+    filename = MetricsPath();
+    std::ofstream filestream(filename,
+                             std::ios::out | std::ios::binary | std::ios::trunc | bestie_noreplace);
+    if (filestream.is_open()) return message.SerializeToOstream(&filestream);
+
+    if (errno != EEXIST) {
+      estream << "ERROR: saving benchmark results in " << filename
+              << " failed with errno: " << errno << "\n";
+      return false;
+    }
```
justus-enf, line 144: I think it is better to just check for file existance at the beginning rather than depending on the underlying implementation setting `errno`.

ghost (commented):
There are some minor things that need attention. Otherwise looks good.

jungleraptor (approved):
lgtm

bestie/utils/googlebench/bestie_reporter_test.cc
ccontavalli: good catch, cleaned.

ccontavalli:
/gcbrun although the failure does not seem to be related to this PR at all.

ccontavalli:
> /gcbrun although the failure does not seem to be related to this PR at all.

I stand corrected, I believe it's a missing dependency in the enkit repo. Basically: my PR is using googlebench for the first time in the enkit repo. googlebench depends on libpfm, but libpfm is not in the enkit repo -> failure. Will add a few commits to this PR once I figure it out.

https://github.com/google/benchmark/pull/1408/files (by @rajachan - thank you!!) added support for it upstream, there's probably a macro import that we're missing.

rajachan:
See https://github.com/google/benchmark/pull/1520, benchmark_deps() might be what you are looking for (I think).

ccontavalli:
(of course.. googlebench 1.9.0 vs 1.9.2, the content of the Run object changed, heap info is no longer an optional pointer, so now commits are 1. updating the library, 2. actually adding the glue between googlebench and bestie).

bestie/utils/googlebench/bestie_reporter.cc
justus-enf: Good point with the race. It was not clear to me that their was an exepctation of a race here.

ghost (approved):

gpaussaenf (approved):
Approving so as to not hold up this PR. Others more familiar with the topic have already commented.

PR #1210, opened 2025-05-06
gee: disable token scope checks

jonathan-enf:
This code required a modern version of gh, but is not supported by the 2023
version of gh that is part of the dev container.

Tested:


jonathan-enf:

mhallmark (approved):
Ah poo, this was a cool feature. If you get an infra ticket up requesting the GH upgrade we can do that. We're building a new version very soon

PR #1211, opened 2025-05-08
ENGPROD-1153: Bugfixes to affected targets detection

gleyba:
What's done in this PR:
- workspace events mechanism was broken, could be after Bazel 7 migration, now all functional in `workspace_events.go`
-  added Bazel's `--repository_cache` usage, for two queries from nearby git points it can speed up process a lot
- After subsequent `bazel query --experimental_workspace_rules_log_file=<> dep(//...)` invocations, workspace log gets regenerated with only diff workspace events. If no changes in workspaces between runs, workspace log could be empty. So, added `start_workspace_log` and `end_workspace_log` arguments. These helps a lot during debug to reuse pre-generated workspace logs files.
- some third party repositories do not tracked with our workspace logs traversing mechanic. For these repositories fallback to parsing all rule targets. But as optimization, calculate only source files size and not hashes.
- and other minor bug fixes


.bazelrc
```diff
@@ -5,6 +5,7 @@
 # The full list of Bazel options: https://docs.bazel.build/versions/master/command-line-reference.html
 common --repo_env=CC=gcc-10 --repo_env=CPP=g++-10 --repo_env=CXX=g++-10
 build --cxxopt="-std=c++20"  --host_cxxopt="-std=c++20" --client_env=BAZEL_CXXOPTS=-std=c++20
+build --host_cxxopt="-Wno-sign-compare" --host_cxxopt="-Wno-deprecated"
```
jrp-enf, line 8: I'm not comfortable with this, maybe we should talk about it.

.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
+{
```
jrp-enf, line 1: I'm not 100% comfortable with this, this isn't sufficiently generic, uses port 50034 which I have no idea if it could conflict with something else, etc.

run_dlv.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
jrp-enf, line 1: I don't think this should be at the top level.

jrp-enf (commented):
I left a few comments, but I want to go over it in more detail Monday.

.bazelrc
gleyba: Hmm, I got build errors without, but can't reproduce already, reverted.

.vscode/launch.json
gleyba: The related command line is in `run_dlv.sh`. I just spent some time figuring out how to run enkit with debugger attached and don't want to get this configuration lost.

run_dlv.sh
gleyba: ok moved to `scripts` directory

jrp-enf (commented):
A few more comments.

lib/bazel/options.go
```diff
@@ -157,12 +164,33 @@ func WithTempWorkspaceRulesLog() (QueryOption, error) {
...
+	}
 	return func(o *queryOptions) {
 		o.workspaceLog = f
+		o.reuseWorkspaceLog = true
 	}, nil
 }
 
+func WithRepositoryCache() QueryOption {
+	return func(o *queryOptions) {
+		o.repositoryCache = "~/bazel/repo_cache"
```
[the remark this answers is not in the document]
gleyba: It is the separate on-disk cache layer. For example you have:
```
    http_archive(
        name = "com_google_protobuf",
        integrity = "sha256-EKDVjzmhqQnpXgDougtbHcZNApl/dBFRlTorNln254w=",
        strip_prefix = "protobuf-29.0",
        urls = ["https://github.com/protocolbuffers/protobuf/archive/refs/tags/v29.0.tar.gz"],
    )
```

Repository cache will contain  only downloaded archive `v29.0.tar.gz`.
And if we have 2 Bazel queries running in parallel with different output_bases, it is not a bad idea for them to share repository download cache.

lib/bazel/workspace_events.go
```diff
@@ -0,0 +1,111 @@
...
+		case *bpb.WorkspaceEvent_DownloadAndExtractEvent:
+			if e.DownloadAndExtractEvent.GetSha256() != "" {
+				checksums = append(checksums, e.DownloadAndExtractEvent.GetSha256())
+			}
+			if e.DownloadAndExtractEvent.GetIntegrity() != "" {
+				checksums = append(checksums, e.DownloadAndExtractEvent.GetIntegrity())
+			}
+		case *bpb.WorkspaceEvent_ExecuteEvent:
+			arguments := e.ExecuteEvent.GetArguments() 
+			if len(arguments) == 2 && arguments[0] == "echo" {
```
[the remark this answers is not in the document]
gleyba: Ok added

lib/bazel/workspace_events.go
```diff
@@ -0,0 +1,111 @@
...
+				// execute_event {
+				//   arguments: "enkit"
+				//   arguments: "astore"
+				//   arguments: "download"
+				//   arguments: "--force-uid"
+				//   arguments: "6vescy2uyzqao3y8nvqrbj4a57ja6bcx"
+				// ...
+				// }
+				// So extract uid as checksum here
+				if arguments[1] == "astore" && arguments[2] == "download" && arguments[3] == "--force-uid" {
```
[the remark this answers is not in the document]
gleyba: I think we need to migrate these store downloads to Bazel-native [repository_ctx#download_and_extract](https://bazel.build/rules/lib/builtins/repository_ctx#download_and_extract).
Let's maybe have this hack for now, but plan refactor in future.

jrp-enf (approved):
Make a new ENGPROD ticket to track that we want to update the workspace_events.go:60 section to be more resilient.

jrp-enf (approved):

mhallmark (approved):

minor-fixes (commented):
The affected targets detection code is unfortunately pretty load-bearing right now, so it's important to understand the risks:

* bugs causing crashes/OoMs will block folks' PR submission. Over the course of a day, people will force in PRs that may cause other breakage.
* bugs causing incomplete results could allow breakage to be merged on master

Ideally, folks would handle breakage well - root-causing the breakage, figuring out what targets they need to test, doing their own run, posting results on their PR, etc. In practice, people will consider themselves hard-blocked, or merge other breakage, which will increase confusion.

---

This PR may have bug fixes, but they look complected with optimizations, which create more risk. I'd recommend splitting out the bug fixes that we must have - those code edits are likely not very contentious, and don't need to be blocked on optimizations.

---

w.r.t. the optimizations - I'm nervous about reasoning about incomplete workspace event logs without the testing improving greatly - we're reversing course on a number of invariants (workspace log is always complete, we don't chase files in external repos) that could introduce correctness issues.

Another path towards getting rid of redundant workspace downloads is to:
* move to bzlmod
* parse the MODULE.bazel.lock for hashes

Then we wouldn't need the workspace event log to snoop on hashes for fetched repositories, and we'd be free to use a pre-populated repository cache in presubmits as well, which would greatly speed things up.

We do need to move to bzlmod eventually to pick up future bazel updates; we might save churn/headaches by going for that goal directly, especially if further optimizations accidentally cause correctness issues.

.vscode/launch.json
gleyba: We have `.vscode/settings.json` checked in, so thought it is ok to have more config.
But deleted already.

lib/bazel/options.go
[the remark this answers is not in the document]
gleyba: @minor-fixes I double-checked, `download_and_extract_event` appear in workspace event log with shared repository cache.
I also tested `affected-targets`  for same git hashes, and workspace event logs are identical.
And shared repository cache for non-parallel queries could be even more effective.

lib/bazel/target.go
```diff
@@ -148,7 +189,8 @@ func shallowHash(w *Workspace, t *bpb.Target) (uint32, error) {
 		// No need to do anything more here.
 
 	case bpb.Target_PACKAGE_GROUP:
-		return 0, fmt.Errorf("PACKAGE_GROUP hashing not implemented")
+		// Just skip PACKAGE_GROUP hashing
+		// return 0, fmt.Errorf("PACKAGE_GROUP hashing not implemented")
```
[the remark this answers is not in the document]
gleyba: I triggered this during testing.
Ok, I removed commented error and added Morte description, 
https://bazel.build/reference/be/functions#package_group rule controls visibility, nothing to hash here.

lib/bazel/workspace_events.go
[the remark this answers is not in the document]
gleyba:
```
cat ~/eob_workspace_events.txt | grep -e "--force-uid" | wc -l
265

cat ~/eob_workspace_events.txt | grep -e "echo" | wc -l 
266
```
Well, likely yes, there are mainly duplicates.
But let's may be have more checks than less.
Also, bzlmod will not help much here, we lose downloaded artifacts caching if not use `repository_ctx` to download. 
Bzlmod only acts as additional wrapper for version/repositories visibility control, but underlying mechanisms are the same.
I added TO DO linked to issue - https://enfabrica.atlassian.net/browse/ENGPROD-1075

.vscode/launch.json
mhallmark: If debugging shouldn't be the way of dealing with this code, what other approach are u suggesting?

gleyba: Well, the codebase is quite complex already. Debugging (and having call stack available) helped me lot to dig into.
But let's not checkout `launch.json`, ok, I will just store in notes.

.bazelrc
```diff
@@ -4,7 +4,7 @@
 #
 # The full list of Bazel options: https://docs.bazel.build/versions/master/command-line-reference.html
 common --repo_env=CC=gcc-10 --repo_env=CPP=g++-10 --repo_env=CXX=g++-10
-build --cxxopt="-std=c++20"  --host_cxxopt="-std=c++20" --client_env=BAZEL_CXXOPTS=-std=c++20
+build --cxxopt="-std=c++20"  --host_cxxopt="-std=c++20" --client_env=BAZEL_CXXOPTS=-std=c++20
```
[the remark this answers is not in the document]
gleyba: done

lib/bazel/target.go
```diff
@@ -293,6 +304,7 @@ func (t *Target) getHash(w *Workspace) (uint32, error) {
 		hash, err := dep.getHash(w)
 		if err != nil {
 			// TODO(scott): Log this condition
+			fmt.Errorf("Target.getHash() for: %s, error:", dep.name, err)
```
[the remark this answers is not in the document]
gleyba: done

gleyba:
@jrp-enf 
> Make a new ENGPROD ticket to track that we want to update the workspace_events.go:60 section to be more resilient.
It is actually [ENGPROD-1075](https://enfabrica.atlassian.net/browse/ENGPROD-1075)

@minor-fixes 
> Another path towards getting rid of redundant workspace downloads is to:
> * move to bzlmod
> * parse the MODULE.bazel.lock for hashes
This is the best way possible, but bzlmod migration is long-term plan.

> This PR may have bug fixes, but they look complected with optimizations, which create more risk. I'd recommend splitting out the bug fixes that we must have - those code edits are likely not very contentious, and don't need to be blocked on optimizations.

These are mainly bug fixes I made heavily debugging, currently workspace log parsing is broken completely, checksums are always empty.
Probably it happened after Bazel 7 migration. I already spent lots of time testing this, could be hard/too much time to split it.
Let's discuss in chats possible strategies to test.

minor-fixes:
> I already spent lots of time testing this, 

I can appreciate that this code is hairy/buggy/error-prone (my sincere apologies!) but the reality is that time spent on a change doesn't constitute mergeable code - what we care about at this stage is whether the risk the change poses is acceptable or not.

It wouldn't surprise me that the code is broken, but it doesn't appear to be blocking folks (unless I'm misunderstanding the impact). Blocking folks is easy to do if this change causes the presubmits to erroneously fail. With >100 people relying on presubmits, it doesn't take much of an outage to cause more overall time loss than spending a bit more time here with more careful editing.

Reading between the lines here (please correct me if I'm wrong!) @mhallmark has approved, but maybe has not reviewed too closely; @jrp-enf has approved but with concerns, and I also echo these concerns. Based on all of this, I think this change attempts too many modifications to be reviewable in its current state. I think the best path forward is to split out singular bug fixes now that you've worked with the code and better understand it; this should be a lot easier for reviewers to digest, and we can all move forward with confidence.

As a side benefit, once bug fixes are merged, we can then invest in refactors and optimizations at the appropriate priority, without risking needing to roll back the bug fixes.

> could be hard/too much time to split it.

If this is too hard, lets get together with @mhallmark to identify the appropriate person to make these changes. I'd estimate it would take someone ~1d to look at this change, isolate the portion that fixes the bug, and test it - even at double that duration (2d) this still seems like a good amount of time given the risk. Maybe my estimations here are completely off?

jrp-enf:
> Reading between the lines here (please correct me if I'm wrong!) @mhallmark has approved, but maybe has not reviewed too closely; @jrp-enf has approved but with concerns, and I also echo these concerns. Based on all of this, I think this change attempts too many modifications to be reviewable in its current state. I think the best path forward is to split out singular bug fixes now that you've worked with the code and better understand it; this should be a lot easier for reviewers to digest, and we can all move forward with confidence.

I've gone on to approve it, I think it's okay as it is.

PR #1217, opened 2025-05-13
feat(astore_upload): add option to print uploaded file and uid to stdout

roivanov:
In the release pipeline we need a way to capture file name and its uid so we can notifiy users and, optionally, save file:uid mapping somewhere.

JIRA: REL-77


minor-fixes (commented):
Recommend something like https://github.com/enfabrica/enkit/pull/1219, so we can keep this option on the commandline and avoid having to add an attribute to every single `astore_upload` target. Then our invocations in internal will look like:

```
bazel run //some/astore/upload:target --@enkit//f/astore:output_format=json <other bazel flags>
```

...with the goal being jq-parseable output on stdout

There are 3 commits on that draft, so you can see the diffs (I'm not going to merge it, but you should be able to copy/tweak naming/whatever as you see fit):

1. update the astore client (https://github.com/enfabrica/enkit/pull/1218, which I need to fixup and merge)
2. your commit from this PR
3. bazel custom flag

bazel/astore/astore_upload_file.sh
```diff
@@ -51,6 +52,9 @@ for TARGET in "${TARGETS[@]}"; do
     exit 2
   fi
   echo >&2 "${TARGET} uploaded as ${FILE}: assigned UID ${FILE_UID}"
+  if [[ ${PRINT_UID:-false} == "true" ]]; then
+    echo "${TARGET} uploaded as ${FILE}: assigned UID ${FILE_UID}"
```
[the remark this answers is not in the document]
roivanov: Yes, this is parsed (grep-ed by the "uploaded as").

For now, I'd like to leave it as plain text, because it is posted on a channel as is.
In the future, when we convert code to 99% python we may think if json is needed.

bazel/astore/astore_upload_file.sh
```diff
@@ -43,14 +44,26 @@ function update_build_file() {
...
+  case "${OUTPUT_FORMAT}" in
+  table) echo "${FILE} ${TARGET}"; cat ; echo ;;
+  json)  if which jq >/dev/null; then
+  # Q: Do we want to fail on missing jq or fallback to json without data added?
+            jq -c '. + {"FILE": "'"${FILE}"'", "TARGET": "'"${TARGET}"'"}'
+         else
+            cat
+         fi ;;
+  *)     cat ;;
+  esac
```
minor-fixes, line 60: why not trust the output format of the tool?

minor-fixes: ...answering my own question - looks like file + target are getting added in here... I think this should be added in at the higher level, to keep this script as short as possible.

reasons:
* defaults in the example PR don't change the existing behavior (not sure if this does)
* this is quickly becoming a tool-wrapping-a-tool with an undefined/unclear API boundary - until we understand the direction of this script, we should keep edits minimal, and not add unneeded additional surface that others might start depending on

roivanov: I agree, I don't like those extra steps either, lmk check astore tool then

minor-fixes: looks like a bit of a pain right now (read: can't send you a PR in <15min :) ). The remote astore path - we'll need that soon, but maybe skipping it for now and returning the uid gets us unblocked? Users should be able to download without the path (`enkit astore download -u <uid> -o <output_file>`) and if needed we can find the BUILD file to list the astore path. This buys us time to figure out how to best get the astore path into this payload

`TARGET` here is actually not a target - looking at the test example, it's a bazel-bin-relative local path, which I don't think is relevant to expose/log (and I don't think we plan to log the target)

roivanov: is `FILE` the right one file path/name?

I can scratch `json` rabbit hole and just do `table` with a bit of `awk`. Please find the next commit.

```
$ bazel run //flextape/server:astore_push --//f/astore:output_format=table >stdout2.out 2>stderr2.out

$ cat stdout2.out 
File=infra/flextape/flextape_server Target=flextape/server/flextape_/flextape UID=4ca03f0e6a70a2e3fc925955ab301e5d=
```

bazel/astore/defs.bzl
```diff
@@ -12,6 +12,16 @@ astore_tag = rule(
     build_setting = config.string_list(flag = True, repeatable = True),
 )
 
+def _astore_output_format(ctx):
+    if ctx.build_setting_value not in ["table", "json"]:
```
kjw-enf, line 16: the shell script above technically allows more values

help says the allowed values are:
table toml json yaml gob

Did you want to expand the list or remove the testing or what? seems odd to have this discrepancy

kjw-enf (approved):

roivanov:
PR scratched
see [REL-210](https://enfabrica.atlassian.net/browse/REL-210)

PR #1223, opened 2025-05-15
ENGPROD-1167: add output to astore_download_and_extract, 

gleyba:
It is needed to initialize cuda packages from astore.


jrp-enf (approved):
Can you get Brian's opinion on this? I don't really know much.

PR #1226, opened 2025-05-20
ENGPROD-1075: Implement `astore get-url` command in enkit to retirieve  download URLs for artifacts by path|uid.

gleyba:
This is the first step to migrate a store rules to `repository_ctx`.

### DIFF
--- a/astore/client/astore/astore.go
+++ b/astore/client/astore/astore.go
@@ -55,19 +55,22 @@ type DownloadOptions struct {
 	*ccontext.Context
 }
 
-type FileToDownload struct {
+type FileToDownloadDescriptor struct {
 	// Name of the file on the remote system.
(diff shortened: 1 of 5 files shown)


minor-fixes:
Thanks! Is the credential helper approach going to work (https://enfabrica.atlassian.net/browse/INFRA-11385?focusedCommentId=110311)? I think that should eliminate the need for this (as well as allow us to get rid of the custom rules entirely!)

gleyba:
> Thanks! Is the credential helper approach going to work (https://enfabrica.atlassian.net/browse/INFRA-11385?focusedCommentId=110311)? I think that should eliminate the need for this (as well as allow us to get rid of the custom rules entirely!)

Yess, likely this work is not needed, closing for now

PR #1231, opened 2025-05-23
ENGPROD-1075: migrate astore rules to repository_ctx as we can use credential helper to authenticate

gleyba:


jrp-enf (approved):
Please provide a description.

Is there any way we can gain back the retry behavior?

PR #1242, opened 2025-06-11
[Allocation Manager] Encapsulate topo requests

paul-enf:
non-building (currently) work-in-progress. fixing issues but putting up for initial look at changes


paul-enf:
AM Server now builds. Will begin testing functionality and debugging any problems found. Goal is to have allocations by topology name working by Monday (6/16). i.e. request topology "a" which maps to an nc-gpu machine, and run commands on that machine.

allocation_manager/client/client.go
```diff
@@ -41,11 +40,15 @@ func New(client apb.AllocationManagerClient, names, configs []string, username,
 		allocationErr: make(chan error),
 	}
 	inv := a.invocation
-	for idx := range names {
-		t := apb.Topology{Name: names[idx], Config: configs[idx]}
-		inv.Topologies = append(inv.Topologies, &t)
-		break // TODO: remove to implement bundles
+
+	// if given a topology name, set that in the invocation
+	if topology_name != "" {
```
kjw-enf, line 45: I'm surprised the go linter didn't complain. go wants to see topologyName; apparently they're allergic to underscores.

allocation_manager/proto/allocation_manager.proto
```diff
@@ -56,14 +56,115 @@ service AllocationManager {
   // SetStatus(SetStatusRequest) returns (SetStatusResponse);
 }
 
+// Contains the entire inventory of known hosts, along with information gathered about those hosts
```
kjw-enf, line 59: merely FYI, I don't think we need to stick too closely to the protobuf style guide, but the most important part on the left-hand-side is "Proto Best Practices" which discusses how to use tag numbers (see the 'reserved' keyword)

allocation_manager/proto/allocation_manager.proto
```diff
@@ -210,13 +324,16 @@ message StatusResponse {
 }
 
 message Stats {
-  // Unit's topology
-  Topology topology = 1;
+  // OBSOLETED: replaced with UnifInfo, below
```
kjw-enf, line 327:
```suggestion
  // OBSOLETED: replaced with UnitInfo, below
```

kjw-enf (approved):

PR #1262, opened 2025-08-21
ENGPROD-1379: Remote asset downloader service, based on `bazel-remote` and compatible with BuildBarn

gleyba:
A very tiny stateless proxy service with main purpose to replace `Download Repository Cache` step in presubmits (currently taking up to 10 minutes of heavy IO every time and heavyweight repository cache preparation regular job).
This service implements [Remote Asset Protocol](https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/asset/v1/remote_asset.proto) to fetch external dependencies and place them into cache.
[bazel-remote](https://github.com/buchgr/bazel-remote) and [bb-remote-asset](https://github.com/buildbarn/bb-remote-asset) considered, but contains some inefficiencies.
For example, there is no way to make `bazel-remote` stateless, but I reused its codebase here.
More info in the [design doc](https://docs.google.com/document/d/1l4lJwOgW1Evdc_aalsIGGeRPpm3fhn0I3cpMRZreymk/edit?tab=t.0#heading=h.i8mhtcz6pa0u).


jrp-enf (approved):
There's no LRU to the cache or anything, right? So if we run out of space, it's just going to start failing? Will this fail builds?

gleyba:
> There's no LRU to the cache or anything, right? So if we run out of space, it's just going to start failing? Will this fail builds?

It use no disk space in most cases, only RAM, up to 2mb per concurrent http download. I may add some rate limiting of number of concurrent downloads based on this.
It fetches data from http and immediately proxy to existing Bazel remote cache.

PR #1263, opened 2025-08-25
enable deb artifacts for arm64 kernel build

roivanov:
Blocked by: https://github.com/enfabrica/linux/pull/102

Blocks: https://github.com/enfabrica/internal/pull/63047

Tested: Can build on gcloud, although arm64 build server is need to create an actual build. On amd64 some binaries were not produced.

JIRA: SF-1573


kbuild/v2/scripts/arm64-build-publish.sh
```diff
@@ -16,6 +16,50 @@ ASTORE_BASE="$7"
 ASTORE_META_DIR="$8"
 
 TARGET="${ARCH}-${FLAVOUR}"
+
+if [ "$FLAVOUR" = "generic" ] ; then
+
+     BUILD_DEB_DIR="$BUILD_ROOT/deb-build/${TARGET}"
```
roivanov, line 29: This is a copy of `amd64` script, technically we can combine them into one.

